PR #4, opened 2024-04-25
Add initial CI

saienduri:
This commit brings over all the CI from SHARK-Turbine and configures it properly for iree-turbine. I also add parallelization to the model testing, so that the cpu model tests run on ubuntu 64 core and gpu tests run on mi250. This way we only hog the mi250 for 4 mins compared to the ~30 mins in SHARK Turbine (we use mi250 for gpu model regression testing on iree PRs so wanted to reduce time being used to only necessity). I didn't see anywhere in iree-org that solely depends on ubuntu 64 core so I don't think we would be blocking anywhere by hogging for while. But, let me know if we want further parallelization due to github runner cost of using 64 core (I can try and see if we can get some of the sd tests running on 32 or 16 core even)


.github/workflows/test_models.yml
```diff
@@ -0,0 +1,90 @@
...
+  # A PR number if a pull request and otherwise the commit hash. This cancels
+  # queued and in-progress runs for the same PR (presubmit) or commit
+  # (postsubmit). The workflow name is prepended to avoid conflicts between
+  # different workflows.
+  group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  test-turbine-models-cpu:
+    runs-on: ubuntu-22.04-64core
```
stellaraccident, line 20: As noted in your comment, let's not do this. This costs $$ to GitHub and would be better to use one of our self managed instances.

.github/workflows/test_shark.yml
```diff
@@ -0,0 +1,52 @@
...
+  # different workflows.
+  group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  test-shark:
+    strategy:
+      matrix:
+        version: [3.11]
+        os: [ubuntu-22.04-64core]
```
stellaraccident, line 23: And here.

stellaraccident (approved):
Thanks!

PR #6, opened 2024-04-25
add decomposition for linspace.Tensor_Tensor

dan-garvey:
this op will otherwise fail in torch-mlir due to shape information not propagating

### DIFF
--- a/shark_turbine/dynamo/decompositions.py
+++ b/shark_turbine/dynamo/decompositions.py
@@ -115,6 +115,7 @@ def _get_default_decomposition_ops() -> DecompositionOpsList:
         aten.lift_fresh_copy.default,
         aten._unsafe_index.Tensor,
         aten.unbind.int,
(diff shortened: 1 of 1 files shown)


dan-garvey:
Should we add some e2e tests somewhere for all of these? The rationale being if someone wants to remove one from the default list for some reason, the op lowering still needs to succeed?

shark_turbine/dynamo/decompositions.py
```diff
@@ -115,6 +115,7 @@ def _get_default_decomposition_ops() -> DecompositionOpsList:
         aten.lift_fresh_copy.default,
         aten._unsafe_index.Tensor,
         aten.unbind.int,
+        aten.linspace.Tensor_Tensor,
```
ScottTodd, line 118: This list could be sorted alphabetically 🤔

ScottTodd (approved):

PR #10, opened 2024-05-02
Add template helpers to the op_reg facility.

stellaraccident:
* This support was ported and generalized from sharktank, where the technique originated.
* Support ops defined in terms of either a str-format or jinja2 based ASM template.
* Helpers for inlining and calling.
* Example ops used for testing.
* Adds dependency on Jinja2.


shark_turbine/runtime/op_reg/impl_helper.py
```diff
@@ -0,0 +1,182 @@
...
+        *,
+        suffix: str = ".mlir",
+    ):
+        try:
+            from jinja2 import Environment, PackageLoader, select_autoescape
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError(
+                "Cannot use JinjaTemplateLoader if jinja2 is not installed"
+            ) from e
+        self.env = Environment(loader=PackageLoader(package_name, package_path))
```
raikonenfnu, line 160: This is nice, so we can just specify the templates/env from the caller. :)

shark_turbine/ops/_str_format_test_ops.py
```diff
@@ -0,0 +1,73 @@
...
+        result_desc.specialize_all_dims()
+
+    def generate(self, ksel: KernelSelection, kb: KernelBuilder):
+        result_type = kb.arg_bindings[0].type  # type: ignore
+        rtt = RankedTensorType(result_type)
+        function_name = (
+            f"turbine_test_add_strformat_{rtt.rank}d_{str(rtt.element_type)}"
+        )
+        func_op = _templates.inline_template_function(
+            kb,
```
raikonenfnu, line 46: Just curious, any special reasoning behind adding support for string-based templates?

stellaraccident: Not a huge one: it let me test everything without an external dep. I had used str format for all of the early work. Jinja is a better experience but just felt better having a path that was dep free.

tests/runtime/device_test.py
```diff
@@ -151,20 +151,28 @@ def testFromTorchDevice(self):
         print(device.dump_device_info())
 
     def testJit(self):
-        from shark_turbine.ops import iree as iree_ops
+        from shark_turbine.ops import _str_format_test_ops as test_ops
```
raikonenfnu, line 154: NIT: Can do later or set as TODO. Add test for jinja-based formatting and check output MLIR string from export perhaps. :)

shark_turbine/ops/_str_format_test_ops.py
raikonenfnu: makes sense :)

raikonenfnu (approved):

tests/runtime/device_test.py
stellaraccident: There is one below on the CPU. I debated having more of a lower level test but in this case, I think a functional test gets the job done.

PR #17, opened 2024-05-15
Add a Launchable class for directly running IREE modules on PyTorch devices.

stellaraccident:
This uses much of the plumbing of the custom ops workflow but is geared for larger scale integrations. Whereas the custom op system is optimized for things more "kernel" sized, focusing on specialization and JIT compilation of variants, this workflow is geared towards integrating entire programs (either from VMFB or JIT compiled for the device in use on the fly) as a Torch callable.

Usage for bring-your-own-VMFB:

```
launch = Launchable.from_vm_module(lambda device: VmModule.mmap(device.vm_instance, "foo.vmfb"))
result = launch(tensor1, tensor2)
```

Usage for JIT compiling:

```
launch = Launchable.jit_compile(MLIR_ASM)
result = launch(tensor1, tensor2)
```

In the first case, it is the caller's responsibility to produce a VMFB that is valid for the given device. In the JIT case, appropriate compiler flags and targeting information will be set based on the type of the device the input tensors are located on (i.e. if ROCM/CUDA, this will also properly differentiate between heterogenous devices on the system and compile a binary for each distinct target).

Limitations:

* The underlying mechanism currently uses the default stream for synchronization. It is TBI to plumb through more explicit support.
* As a consequence of the above, we also are syncing the device after launch.
* We are waiting for upstream PyTorch patches to land to get UUIDs from torch devices. Without this, enumeration order has to match, which is not guaranteed.

Includes workarounds for:

* https://github.com/iree-org/iree/issues/17402
* https://github.com/iree-org/iree/issues/17403

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 # Build/test requirements.
 Jinja2==3.1.3
 numpy==1.26.3
+parameterized==0.9.0
 pytest==8.0.0
 pytest-xdist==3.5.0
 mypy==1.8.0
--- a/shark_turbine/aot/params.py
+++ b/shark_turbine/aot/params.py
@@ -256,6 +256,10 @@ class ParameterArchiveBuilder:
     def __init__(self):
         self._index = ParameterIndex()
 
+    @property
+    def index(self) -> ParameterIndex:
+        return self._index
+
     def save(self, file_path: Union[str, Path]):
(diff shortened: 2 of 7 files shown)


monorimet (approved):
Very cool, I can't wait to use this.

LGTM.

PR #18, opened 2024-05-15
Add initial framework for TKW

martin-luecke:
TODO: expand PR description, this is currently for enabling commenting 

Detailed description of this project in [shark_turbine/kernel/functional/README.md](https://github.com/iree-org/iree-turbine/compare/wave_kernel_language?expand=1#diff-8bc4de72a74936137a9fc8cb3d47f17bb27c997ce651b876143d731a17707041)


shark_turbine/kernel/_support/nodes.py
```diff
@@ -0,0 +1,151 @@
...
+
+    @classmethod
+    def from_fx_node(cls, node: fx.Node):
+        instance = cls(node.graph, node.op, *node.args)
+        instance.fx_node = node
+        return instance
+
+    def __str__(self) -> str:
+        name = get_node_name(self.__class__.__name__)
+        vars_list = [f"{key}={value}" for key, value in vars(self).items()][2:]
```
harsh-nod, line 41: where does the number 2 come from?

shark_turbine/kernel/_support/nodes.py
```diff
@@ -0,0 +1,151 @@
...
+@dataclass
+class CustomNode(ABC):
+    """
+    Base class for all custom fx nodes.
+    """
+
+    op: Any
+
+    @classmethod
+    def from_fx_node(cls, node: fx.Node):
```
harsh-nod, line 34: could you annotate the return type as well?

shark_turbine/kernel/_support/nodes.py
```diff
@@ -0,0 +1,151 @@
...
+        name = get_node_name(self.__class__.__name__)
+        vars_list = [f"{key}={value}" for key, value in vars(self).items()][2:]
+        vars_str = ", ".join(vars_list)
+        return f"{name} {vars_str}" + "\n"
+
+    def custom_string(self, value_map: dict[str, str]) -> str:
+        # If a subclass does not define custom printing we revert to the default
+        return str(self)
+
+    def add_to_graph(self, graph: fx.Graph):
```
harsh-nod, line 49: This function is a bit confusing. When adding a node to a graph, I would have expected a graph.create_node. Instead there is a graph.create_proxy here. What is the difference? Maybe consider calling this add_proxy?

shark_turbine/kernel/_support/nodes.py
```diff
@@ -0,0 +1,151 @@
...
+        arg_list = tuple([value for _, value in vars(self).items()][2:])
+        self.fx_node = graph.create_proxy(
+            "call_function",
+            target=self.op,
+            args=arg_list,
+            kwargs={},
+        )
+
+    @classmethod
+    def handle(cls, graph, *args, **kwargs):
```
harsh-nod, line 59: nit: return type annotation

shark_turbine/kernel/_support/nodes.py
```diff
@@ -0,0 +1,151 @@
...
+class UnknownNode(CustomNode):
+    """
+    Represents an fx.Node that has no corresponding CustomNode class.
+    """
+
+    args: Sequence[Any]
+    kwargs: dict[Any, Any]
+
+    @classmethod
+    def from_fx_node(cls, node: fx.Node):
```
harsh-nod, line 81: nit: return types here and below

stellaraccident:
I saw this branch in my latest fetch and was intrigued. Looks interesting

shark_turbine/kernel/_support/tracing.py
```diff
@@ -177,6 +178,29 @@ def __init__(self, region_graph: RegionGraph, *, grid_type: Type[Grid]):
...
+
+        setattr(self, f"handle_{name}", handler)
+
+    def node(self, node: fx.Node) -> CustomNode:
+        for name, nodeT in self.custom_ops.items():
+            # The fx_nodes have a suffix depending on the number of occurrences
+            # of similar nodes. We are only interested in the name prefix.
+
+            # TODO: Instead simply shape the suffix off, as this depends on the
+            #       iteration order for nodes sharing a prefic,
```
harsh-nod, line 197: typo: prefix

shark_turbine/kernel/functional/README.md
```diff
@@ -0,0 +1,99 @@
+# TKF: A Wave Kernel Language for GPUs
+
+TKF is a high-level programming language designed to simplify the development of GPU micro-kernels by abstracting over intricate details of GPU hardware. It allows developers to write efficient micro-kernels, focusing on core computations while inferring the required data transfers and indexing automatically. By leveraging a wave-based programming model, TKF enables expressing programs leveraging coalesced memory accesses effortlessly.
```
harsh-nod, line 3: Could you modify this to include the mma intrinsics as well , like - TKF enables expressing programs leveraging coalesced memory accesses and matrix multiplication intrinsics effortlessly.

shark_turbine/kernel/functional/README.md
```diff
@@ -0,0 +1,99 @@
...
+        # a_reg: tkf.Register[M, K, tkl.f16]
+        a_reg = tkf.read(a, elements_per_thread=LOAD_ELEMS_PER_THREAD)
+        # b_reg: tkf.Register[N, K, tkl.f16]
+        b_reg = tkf.read(b, elements_per_thread=LOAD_ELEMS_PER_THREAD)
+        # c_reg: tkf.Register[M, N, tkl.f32]
+        c_reg = tkf.mma(a_reg, b_reg, c_reg)
+        return c_reg
+
+    # repeat represents the results of the loop
+    tkf.write(repeat, c, elements_per_thread=STORE_ELEMS_PER_THREAD)
```
harsh-nod, line 80: A minor syntax issue: here it seems like we have just defined a repeat function and are passing that function to tkf.write. Instead we might want to call repeat and get a result and then pass that to tkf.write. Something like
```
result = repeat(c_reg)
tkf.write(result, c, ...)
```
What do you think?

shark_turbine/kernel/functional/README.md
```diff
@@ -0,0 +1,99 @@
...
+]
+
+# Wave-level micro-kernel.
+# Since warps are not directly addressable, there is no
+# explicit notion of a warp id (like a workgroup or thread id).
+@tkf.wave(constraints)
+def gemm(
+    a: tkl.Memory[M, K, ADDRESS_SPACE, tkl.f16],
+    b: tkl.Memory[N, K, ADDRESS_SPACE, tkl.f16],
+    c: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f32],
```
harsh-nod, line 62: One thing that also comes up is what are these M, N and K values? Are they the original tensor shapes or just the shapes the wave sees? Can you add a comment here explaining that?

shark_turbine/kernel/functional/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+In GPU programming, a wavefront is a group of threads (or work items) that execute the same instruction in lockstep. In particular, coalesced memory accesses by all threads in a wavefront are executed together. This is analogous to the concept of a "warp" in NVIDIA's CUDA programming model.
+Typically, on AMD GPUs, a wavefront contains 32 or 64 threads, all of which participate in executing the same instruction at the same time.
+
+In this representation, memory access patterns are more naturally optimized for coalescing, reducing the complexity and manual effort required to achieve optimized memory transactions. In consequence, programmers can focus more on the core computational logic rather than the intricacies of thread coordination and memory management.
+This approach contrasts with traditional models like OpenCL and CUDA, which often require more explicit management of threads and synchronization.
+
+## Gemm example
+
+```Python
```
harsh-nod, line 28: Is there a way to have this reference the file directly?Then you won't have to copy it here and maintain 2 versions.

harsh-nod (changes requested):
Thanks for pushing this. Just some minor comments but looks good overall.

stellaraccident:
One naming nit: aside from some functional inspiration, the big idea here is a wave level programming model. If we're allocating the third letter of the prefix to what something is, probably tkw vs tkf. I doubt this will be the only thing that takes more of a functional view of things. And in general, that is a how, not a what.

martin-luecke (commented):
Thank you for the comments, they should now all be addressed

martin-luecke:
I updated the IREE version [here](https://github.com/iree-org/iree-turbine/blob/fa8819b9a2ac446933421afeda26d58781db2dc1/iree-requirements.txt) but the CI still seems to pick up a version which does not expose the amdgpu dialect. Do you know where to adjust this? @harsh-nod @stellaraccident

stellaraccident:
Look to the failing CI run?

martin-luecke:
Sorry, I rushed a bit to get to some other work.

I only now understood that the CI pulls nod-ai/SHARK and uses those requirements.
The CI here contains an sed command to adjust the SHARK requirements from iree-turbine/main to the currently tested commit. 
The sed command is here: 

https://github.com/iree-org/iree-turbine/blob/45f1de86776c64737d8b4b8509048a0948e98177/.github/workflows/test_shark.yml#L49-L49

and seems to be supposed to match [here](https://github.com/nod-ai/SHARK/blob/78160b82e860f65748d523d9907c20ca5be44fa5/requirements.txt#L9)

However, it does not. I am not sure whether that is intended to test everything with the requirements of iree-turbine/main or an oversight. 
I modified it to do the replacement, but that currently messes up torch dependencies. Also it still seems to not pull the iree version indicated in `iree-requirements.txt`. I assumed this is the place to modify the requirement, similar to how it is done in [this patch](https://github.com/iree-org/iree-turbine/pull/8)
But I also see that in earlier CI runs unrelated to this PR a different iree version from the `iree-requirements.txt` is used for testing. e.g. [here](https://github.com/iree-org/iree-turbine/actions/runs/9088680117/job/24978647677#step:5:182) which pulls version 20240410.859 instead of version 20240514.893 indicated in the iree-requirements.

I will have to revisit this later.

shark_turbine/kernel/_support/nodes.py
```diff
@@ -0,0 +1,159 @@
...
+    lhs: fx.Node
+    rhs: fx.Node
+    acc: fx.Node
+
+
+@dataclass
+class ReadNode(CustomNode):
+    memory: fx.Proxy
+    elements_per_thread: Optional[Any] = None
+    type: Optional[Type[Register]] = None
```
harsh-nod, line 127: Does Optional here mean we can omit the type?

shark_turbine/kernel/wave/ops.py
```diff
@@ -0,0 +1,36 @@
...
+
+__all__ = [
+    "construct_register_from_metadata",
+    "read",
+    "write",
+    "mma",
+    "reduction",
+]
+
+
```
harsh-nod, line 13: Maybe not in this PR, but could these be automatically generated?

shark_turbine/kernel/wave/wave.py
```diff
@@ -0,0 +1,91 @@
...
+        self.grid_type = Grid[None, None]
+        self._name = name
+        self._f = eager_function
+        self._sig = inspect.signature(eager_function)
+
+    def _trace(self, dump: bool = False) -> CapturedTrace:
+        region_graph = KernelRegionGraph()
+        with CompiledContext(region_graph, grid_type=self.grid_type) as context:
+            custom_ops: dict[str, Type[CustomNode]] = {
+                "construct_register_from_metadata": ConstructRegisterFromMetadataNode,
```
harsh-nod, line 42: Do we need to specify all new nodes here or can they be inferred from one single source of truth?

shark_turbine/kernel/wave/wave.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # Register custom ops
+            for name, op in custom_ops.items():
+                context.register_custom_op(name, op)
+
+            with region_graph.subtracer() as subtracer:
+                root_name, _ = subtracer.trace(self._f)
+                trace = CapturedTrace(region_graph, root_name)
+
+            if dump:
```
harsh-nod, line 58: What do you think of replace dump with a "LLVM_DEBUG" style flag that can be activated based on an incoming argument (like debug-only=)

shark_turbine/kernel/_support/tracing.py
```diff
@@ -177,6 +178,29 @@ def __init__(self, region_graph: RegionGraph, *, grid_type: Type[Grid]):
...
+
+    def node(self, node: fx.Node) -> CustomNode:
+        for name, nodeT in self.custom_ops.items():
+            # The fx_nodes have a suffix depending on the number of occurrences
+            # of similar nodes. We are only interested in the name prefix.
+
+            # TODO: Instead simply shape the suffix off, as this depends on the
+            #       iteration order for nodes sharing a prefix,
+            #       e.g. read and read_shared
+            if node.name.startswith(name):
```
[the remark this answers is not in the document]
martin-luecke: I agree 👍 
I added a field `tkw_op` where the custom op that was used to create this node is recorded

martin-luecke (commented):
Thank you for the comments @harsh-nod 
I will explore tomorrow autogenerating more of the operation boilerplate and in particular see whether the registration of the custom ops to the context can be done differently.

shark_turbine/kernel/_support/tracing.py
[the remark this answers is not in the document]
harsh-nod: awesome thank you!

shark_turbine/kernel/wave/ops.py
harsh-nod: great, glad to know this is possible

shark_turbine/kernel/wave/wave.py
martin-luecke: I changed this to use an environment variable `DEBUG`

shark_turbine/kernel/_support/nodes.py
```diff
@@ -0,0 +1,159 @@
...
+@dataclass
+class CustomNode(ABC):
+    """
+    Base class for all custom fx nodes.
+    """
+
+    graph: fx.Graph
+    op: Any
+
+    @classmethod
```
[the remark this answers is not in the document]
martin-luecke: I added unit tests for the tracing of all our ops

shark_turbine/kernel/wave/wave.py
martin-luecke: I reworked how new operations are registered. The single source of truth is `shark_turbine/kernel/ops/wave_ops`, e.g. [here](
https://github.com/iree-org/iree-turbine/blob/ec5b437ed1922cb6d0aea1e121d6ddd86cc3f307/shark_turbine/kernel/ops/wave_ops.py#L182)

This generates the functions used to construct and trace a program. The current limitation is proper IDE support as these functions are not statically known. I intend to work on that next and explore using `.pyi` stubs for that.

martin-luecke:
Oops, seems my VSCode plugin glitched and I accidentally closed. Now reopened. 
I pushed the lit tests to their own top-level directory as suggested to avoid test tools fighting each other.

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -0,0 +1,238 @@
+from abc import ABC
+from dataclasses import dataclass, field
+from functools import wraps
+import sys
+from typing import (
+    TYPE_CHECKING,
```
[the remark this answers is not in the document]
martin-luecke: This is a constant that evaluates to `True` only when the type checker runs (e.g. mypy / Pylance). Just below I import something guarded by this constant to enable proper type checking. 
The unguarded import would result in a circular dependency during runtime, but as I need the import only for type checking this works nicely.

tests/lit.cfg.py
```diff
@@ -0,0 +1,61 @@
...
+        ("%PYTHON", sys.executable),
+    ]
+)
+
+# Find a suitable filecheck.
+filecheck_exe = None
+if filecheck_exe is None:
+    filecheck_exe = shutil.which("FileCheck")
+    if filecheck_exe:
+        print(f"Using LLVM FileCheck: {filecheck_exe}")
```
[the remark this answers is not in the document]
martin-luecke: Good suggestion, done! 
For wave.py I removed printing entirely as I don't think we always want to have the graph in the debug output. Initially the tests depended on that, but I found a better approach using a different decorator `wave_trace_only`

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -0,0 +1,238 @@
...
+import torch.fx as fx
+
+if TYPE_CHECKING:
+    from ..lang.functional_types import Memory, Register
+from .._support.indexing import IndexExpr
+from .._support.dtype import DataType
+from .base import OpDispatcher
+
+
+def get_node_name(string: str, skip_first: bool = True):
```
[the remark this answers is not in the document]
martin-luecke: The fx node name depends on previous nodes in the graph, e.g. it might be `read_6`. This function extracted a readable name from our node classes. I removed it as we already give ops a name in the new `define_op` [here](https://github.com/iree-org/iree-turbine/blob/da97c1c9de86daa557902f0b0560a2f49301aa85/shark_turbine/kernel/ops/wave_ops.py#L214) and can simply use that instead.

tests/kernel/wave/tracing.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+@run
+def test_trace_empty():
+    @tkw.wave_trace_only()
+    def test(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
+        pass
+
+    trace = test()
+    print(trace.get_root_graph())
```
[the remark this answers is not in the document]
martin-luecke: I added a helper that prints all subgraphs of a trace. So all tests now simply call 
```
trace = test()
print_trace(trace)
```
and then have a bunch of CHECK statements.

martin-luecke:
Thank you @harsh-nod 

Inspired by your comment I added the `add_to_graph` to our custom ops and took the time to simplify the interface for adding them to the graph after tracing.
In particular I removed the `fx_op` field from CustomOp which was required for initialization before. It is more a technical detail and remnant from the previous implementation. Instead, there is now a `_tracing_function` field on the class, which is set automatically by the `define_op` decorater.

With that the API for creating nodes after the tracing looks like [this](https://github.com/iree-org/iree-turbine/blob/da97c1c9de86daa557902f0b0560a2f49301aa85/lit_tests/kernel/wave/tracing.py#L70):
```
[...]
read = Read(a).add_to_graph(graph)
write = Write(read, a, 4).add_to_graph(graph)
```

shark_turbine/kernel/lang/functional_types.py
```diff
@@ -0,0 +1,137 @@
...
+            symbolic_shape = init_symbolic_shape
+            rank = len(symbolic_shape)
+            address_space = init_address_space
+            dtype = init_dtype
+            usage = init_usage
+
+        return cast(Type[MemoryTypeT], MemoryType)
+
+
+class Memory(metaclass=_MemoryStorage):
```
Groverkss, line 54: Is this type different from KernelBuffer? If not, can we reuse them?

harsh-nod:
> Thank you @harsh-nod
> 
> Inspired by your comment I added the `add_to_graph` to our custom ops and took the time to simplify the interface for adding them to the graph after tracing. In particular I removed the `fx_op` field from CustomOp which was required for initialization before. It is more a technical detail and remnant from the previous implementation. Instead, there is now a `_tracing_function` field on the class, which is set automatically by the `define_op` decorater.
> 
> With that the API for creating nodes after the tracing looks like [this](https://github.com/iree-org/iree-turbine/blob/da97c1c9de86daa557902f0b0560a2f49301aa85/lit_tests/kernel/wave/tracing.py#L70):
> 
> ```
> [...]
> read = Read(a).add_to_graph(graph)
> write = Write(read, a, 4).add_to_graph(graph)
> ```

Thanks Martin, that's awesome. Thank you for the push and this is a great first PR to build on top of.

harsh-nod (approved):
lgtm! thanks!

shark_turbine/kernel/lang/functional_types.py
```diff
@@ -0,0 +1,136 @@
+from typing import Optional, Type, TypeVar, cast, ClassVar
```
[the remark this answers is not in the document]
harsh-nod: Good catch! Yeah this should be wave_types since we have renamed from functional -> wave.

shark_turbine/kernel/lang/functional_types.py
harsh-nod: I believe the biggest difference is the addition of the AddressSpace in the type to be able to model different levels in the memory hierarchy (think AIE for example which has multiple levels) and also didn't want to break things in the other pipeline but will let Martin comment as well.

shark_turbine/kernel/lang/functional_types.py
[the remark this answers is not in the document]
martin-luecke: I updated this to `wave_types.py` and added unit tests.

shark_turbine/kernel/lang/functional_types.py
martin-luecke: Indeed, the plan is to infer where the memory is to be placed and also whether it is used for input / output / both.
Examples of the supported usage is [here](https://github.com/iree-org/iree-turbine/blob/eaa04e20389df28304eeb3efe12acc8c68738ca8/tests/kernel/wave/types_test.py#L15)

Groverkss: It's fine to land this, but effectively, KernelBuffer and this MemoryStorage are the same thing. KernelBufferUsage.Local would just get another attribute specifying what kind it is.

martin-luecke: You are right, I reworked the types to build on top of `KernelBufferMeta` instead.
For that I changed it as follows:
- made it public
- added address_space to it + moved enum to this file
- added the option to manually specify a type name besides inferring from KernelBufferUsage

We should only deviate from the established framework if we have good reasons, so thanks for highlighting this.

Let me know if this is alright for you.

PR #20, opened 2024-05-18
Add ops.iree.transfer_to_logical_device() custom op.

stellaraccident:
This exposes IREE's multi-device capabilities to PyTorch programs, allowing them to transfer parts of the computation to other devices. This will be used as part of the sharding work in sharktank to do auto placement of sharded tensors (i.e. across GPU devices, etc). But it is also available for heterogenous/ad-hoc offload to other devices (GPU slices, CPU slices, NPU, etc).

Since this is the first op we have that has different eager vs graph mode semantics, add a new CustomOp.eager_execute() hook which allows the eager execution semantics to be specified separately from graph compilation. Since these local device transfers are only meaningful within an IREE program, they are no-ops during eager execution.

(requires a WIP branch on IREE)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,7 +49,7 @@ jobs:
           # from non default locations first. Installing the PyTorch CPU
           # wheels saves multiple minutes and a lot of bandwidth on runner setup.
           pip install --no-compile -r pytorch-cpu-requirements.txt
-          pip install --no-cache-dir -r iree-requirements.txt
+          pip install --no-cache-dir -r iree-requirements-ci.txt
           pip install -r requirements.txt -e .
 
       - name: Run unit tests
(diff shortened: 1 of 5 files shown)


benvanik (approved):

sogartar:
I rebased the PR, I have more PRs that depend on this.

sogartar:
The CI is failing due to not having a recent enough python package if IREE, that includes the [multi-device work](https://github.com/iree-org/iree/pull/17987).

The release package of IREE is failing due to https://github.com/iree-org/iree/actions/runs/10294193487/job/28492076494

sogartar:
Actually we are not using nightly packages for the CI, but the ones published to `pypi.org`. I guess this is intentional? Then we need to update the IREE pypi packages.

stellaraccident:
Not intentional to be pinned like that. Feel free to pin to an appropriate nightly.

PR #21, opened 2024-05-20
Load pins from iree-requirements.txt in setup.py.

ScottTodd:
Without this, pip installs require the unqualified `iree-compiler` package.
With this, pip installs require the `iree-compiler>=[version pin]` package (note that `get_version_spec` here is inserting the `>=`, even if the requirements file has `==`)

This helps for https://github.com/nod-ai/sharktank/tree/main?tab=readme-ov-file#install-development-packages with these instructions:
```
# Clone and install editable iree-turbine dep in deps/
pip install -f https://iree.dev/pip-release-links.html --src deps \
  -e "git+https://github.com/iree-org/iree-turbine.git#egg=shark-turbine"
```

### DIFF
--- a/iree-requirements.txt
+++ b/iree-requirements.txt
@@ -1,3 +1,3 @@
---extra-index-url https://iree.dev/pip-release-links.html
+--find-links https://iree.dev/pip-release-links.html
 iree-compiler==20240514.893
 iree-runtime==20240514.893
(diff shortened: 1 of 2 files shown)


iree-requirements.txt
```diff
@@ -1,2 +1,3 @@
+--find-links https://iree.dev/pip-release-links.html
 iree-compiler==20240514.893
 iree-runtime==20240514.893
```
ScottTodd, line 3: Not sure how to make this work without either

A) adding `--find-links https://iree.dev/pip-release-links.html` to all install commands for iree-turbine
B) only pinning to versions here that are published to pypi (https://pypi.org/project/iree-compiler/#history)

ScottTodd:
Closing this since it's stale.

PR #22, opened 2024-05-20
Add dco.yml file allowing members to use commit signing.

ScottTodd:


ScottTodd:
@jpienaar Is there a way to configure this across the entire organization, without needing to add this file to each repo?

ScottTodd:
> @jpienaar Is there a way to configure this across the entire organization, without needing to add this file to each repo?

I didn't see a way to configure app settings in the UI, but maybe we can define a common file and then reference that file from other repos: https://github.com/probot/probot-config?tab=readme-ov-file#setup (might still require a new file in every repo? can test if .github works without that)

ScottTodd:
Ok, I think https://github.com/iree-org/.github/blob/main/.github/dco.yml will be picked up now.

PR #31, opened 2024-06-18
CI Fixes for test_models and SHARK testing

saienduri:
This commit makes the necessary changes to both the workflows to get the CI working properly. The SHARK workflow file is temporarily using a branch that I created in nod-ai/SHARK to resolve some versioning issues. I will land there and update this back to use the main branch once that happens.

### DIFF
--- a/.github/workflows/test_models.yml
+++ b/.github/workflows/test_models.yml
@@ -49,7 +49,11 @@ jobs:
           pip install --no-compile --pre --upgrade -r requirements.txt
           pip install --no-compile --pre -e .[testing]
           pip install --no-compile --pre --upgrade -e ${{ github.workspace }}/SHARK-Turbine/models -r ${{ github.workspace }}/SHARK-Turbine/models/requirements.txt
-          pip install --no-cache-dir -r iree-requirements.txt -f https://iree.dev/pip-release-links.html
+          pip install \
+            --find-links https://iree.dev/pip-release-links.html \
+            --upgrade \
+            iree-compiler \
+            iree-runtime
       - name: Show current free memory
         run: |
           free -mh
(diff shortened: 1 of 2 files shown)


.github/workflows/test_shark.yml
```diff
@@ -36,9 +36,9 @@ jobs:
       - name: "Checkout SHARK"
         uses: actions/checkout@v4
         with:
-          repository: "nod-ai/SHARK.git"
+          repository: "nod-ai/SHARK"
           path: SHARK
-          ref: "iree-turbine-switch"
+          ref: "diffusers-version"
```
harsh-nod, line 41: What is this diffusers-version? Is this your local branch?

saienduri: Yup, it's this: https://github.com/nod-ai/SHARK/pull/2157

saienduri: Ean just approved, so I can land there and change this to check out main

.github/workflows/test_models.yml
```diff
@@ -49,6 +49,11 @@ jobs:
           pip install --no-compile --pre --upgrade -r requirements.txt
           pip install --no-compile --pre -e .[testing]
           pip install --no-compile --pre --upgrade -e ${{ github.workspace }}/SHARK-Turbine/models -r ${{ github.workspace }}/SHARK-Turbine/models/requirements.txt
+          pip install \
+            --find-links https://iree.dev/pip-release-links.html \
```
harsh-nod, line 53: This could create some backwards compatibility issues if we are using the latest tree-compiler & runtime, but since you have verified this we can go with this for now.

harsh-nod (approved):
Thanks for getting this to work!

.github/workflows/test_models.yml
saienduri: Yeah, since it is just for model testing should be fine for now. But, I can pin it to a specific version here if we ever run into problems. Also, thought it would just be good for visibility to see when our models are not passing using the latest iree.

PR #35, opened 2024-06-24
Add pass for embedding simple metadata with iree.reflection attribute.

monorimet:


stellaraccident (approved):
Seems easy enough. We may want to also expose this as an argument on export if it gets use.

stellaraccident (approved):

PR #40, opened 2024-07-02
 Fixes metadata for multiple entry point functions.

monorimet:


shark_turbine/transforms/general/add_metadata.py
```diff
@@ -50,8 +50,10 @@ def parse_metadata_dict(metadata_dict: dict) -> DictAttr:
 
         metadata_dict_attr = parse_metadata_dict(self.inp_metadata)
         for func_op in self.funcs:
-            if func_op.op.attributes[1].attr.value == self.func_name:
-                func_op.op.attributes["iree.reflection"] = metadata_dict_attr
+            for attr in func_op.op.attributes:
```
stellaraccident, line 53: The better way to do this, versus looking at all attribute values for a match, would be to use `SymbolTable.get_symbol_name(func_op)`. That will return a StringAttr.

Example from somewhere else that does this:

```
        symbol_op = symbol_op.detach_from_parent()
        orig_symbol = SymbolTable.get_symbol_name(symbol_op)
        orig_symbol_name = StringAttr(orig_symbol).value
```

If doing a lot of these, it is better to instantiate a `SymbolTable` which scans everything once and then lets you look up by name in O(1). But for something like this, it is fine to traverse, I suspect.

stellaraccident (approved):

PR #41, opened 2024-07-08
Add TKW graph expansion

martin-luecke:
This adds the expansion of the original program depending on workgroup and hardware constraints.
The exact semantics can be seen well in the new [tests](https://github.com/iree-org/iree-turbine/blob/dedd438641072ac9e5f47dc60212a9306de55a00/lit_tests/kernel/wave/expansion.py).

Additionally, this introduces:
- `HardwareConstraint` that specifies the number of waves per block, number of threads per wave and mma instruction sizes. 
- New custom nodes for:
`Output`, `IterArg`
- Some new convenience features like updating the args of a node through the custom interface. (Ideally this would be accomplished through intercepting dataclass setters, but that interferes with the tracing and I did not yet find a workaround.)

In particular the handling of iterArgs should be improved in follow-up PRs. For instance, accessing them currently requires iteration over all nodes in a subgraph. We probably want to provide access to them through the corresponding `Reduction` node.

### DIFF
--- a/lit_tests/kernel/wave/expansion.py
+++ b/lit_tests/kernel/wave/expansion.py
@@ -1,0 +1,93 @@
+# RUN: python %s | FileCheck %s
+
+import logging
+from typing import Callable
+import unittest
+import shark_turbine.kernel as tk
+import shark_turbine.kernel.lang as tkl
+import shark_turbine.kernel.wave as tkw
+from shark_turbine.kernel.wave.expansion import expand_graph
+from shark_turbine.kernel._support.tracing import CapturedTrace
+from shark_turbine.kernel.ops.wave_ops import get_custom
+
+
+def run(func: Callable[[], None]) -> Callable[[], None]:
+    """Run a function as part of the test suite."""
+    if __name__ == "__main__":
+        func()
+
+    return func
+
+
+def print_trace(trace: CapturedTrace):
+    """
+    Prints all subgraphs of a trace starting with the root graph.
+    The graphs are printed first in the torch printing format and then using
+    our custom node format.
+    """
+    # The root graph is at the back so we print the subgraphs in reverse order
+    for subgraph in reversed(list(trace.region_graph.subgraphs.values())):
+        print(subgraph)
+        for node in subgraph.nodes:
+            print(get_custom(node))
+
+
+# Input sizes
+M = tkl.sym.M
+N = tkl.sym.N
+K = tkl.sym.K
+
+# Workgroup tile sizes
+BLOCK_M = tkl.sym.BLOCK_M
+BLOCK_N = tkl.sym.BLOCK_N
+BLOCK_K = tkl.sym.BLOCK_L
+
+# Address space (for GPU, shared(1) or global(0))
+ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
+
+
+@tkw.wave_trace_only()
+def read_write(
+    a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+    c: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f32],
+):
+    a_reg = tkw.read(a, elements_per_thread=4)
+    tkw.write(a_reg, c, elements_per_thread=4)
+
+
+def test_read_write():
+    with tk.gen.TestLaunchContext({}):
+        graph = read_write()
+        expand_graph(graph)
+        print_trace(graph)
+
+
+@tkw.wave_trace_only()
+def gemm(
+    a: tkl.Memory[M, K, ADDRESS_SPACE, tkl.f16],
+    b: tkl.Memory[N, K, ADDRESS_SPACE, tkl.f16],
+    c: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f32],
+):
+    c_reg = tkl.Register[M, N, tkl.f32](0.0)
+
+    @tkw.reduction(K, init_args=[c_reg])
+    def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
+        a_reg = tkw.read(a, elements_per_thread=4)
+        b_reg = tkw.read(b, elements_per_thread=4)
+        acc = tkw.mma(a_reg, b_reg, acc)
+        return acc
+
+    tkw.write(repeat, c, elements_per_thread=4)
+
+
+@run
+def test_gemm():
+    with tk.gen.TestLaunchContext({}):
+        graph = gemm()
+        expand_graph(graph)
+        print_trace(graph)
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.DEBUG)
+    unittest.main()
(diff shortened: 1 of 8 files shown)


shark_turbine/kernel/wave/expansion.py
```diff
@@ -0,0 +1,324 @@
...
+                # Temporarily replace the carried arg here to avoid duplicated expansion.
+                # Otherwise we have the following situation:
+                # Suppose we have:
+                #   mma_0_0_0(..., acc_0_0_0)
+                #   mma_0_0_1(..., mma_0_0_0)
+                # Expanding mma_0_0_1 to mma_0_0_2 will trigger expansion of its arg
+                # mma_0_0_0 in dims 0_0_2 as well, effectively duplicating the new node.
+                # To avoid this we temporarily replace the use of it with the original iter_arg
+                # Another option would be to pass a mask of which args to expand, but that
+                # complicates the expansion base case
```
harsh-nod, line 309: Let's chat about this more, I am not sure I fully understand why this is necessary.

lit_tests/kernel/wave/tracing.py
```diff
@@ -80,7 +80,7 @@ def test(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
     # CHECK-SAME: MemoryType[M, N].of(f16)
     # CHECK: read(memory=a
     # CHECK: write(register_=read, memory=a
-    # CHECK: unknown: output
```
harsh-nod, line 83: What is this unknown and why did we have this before?

lit_tests/kernel/wave/expansion.py
```diff
@@ -0,0 +1,342 @@
...
+):
+    a_reg = tkw.read(a, elements_per_thread=4)
+    tkw.write(a_reg, c, elements_per_thread=4)
+
+
+@run
+def test_read_write_equal_sizes():
+    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+    constraints += [tkw.HardwareConstraint(threads_per_wave=64, waves_per_block=[1, 1])]
```
harsh-nod, line 61: Should this be [1, 1, 1]? Also where is the mma type specified?

lit_tests/kernel/wave/expansion.py
```diff
@@ -0,0 +1,342 @@
...
+            BLOCK_N: 32,
+        }
+    ):
+        graph = read_write_same_size()
+        IndexingContext.current().finalize()
+        expand_graph(graph, constraints)
+        print_trace(graph)
+        # CHECK: %a
+        # CHECK: %c
+        # CHECK: %read
```
harsh-nod, line 75: why is this not read_0_0 ?

lit_tests/kernel/wave/expansion.py
```diff
@@ -0,0 +1,342 @@
...
+    ):
+        graph = read_write_different_dims()
+        IndexingContext.current().finalize()
+        expand_graph(graph, constraints)
+        print_trace(graph)
+        # CHECK: %a
+        # CHECK: %c
+        # CHECK: %read_0_0_0
+        # CHECK-SAME: (%a, 4)
+        # CHECK: %read_1_0_0
```
harsh-nod, line 127: if all the tile sizes are 32 and mma size is 16, shouldnt there be 4 reads and 4 writes?

harsh-nod (changes requested):
This is a great start at a more generalized expansion. Let's chat about it more tonight.

harsh-nod (changes requested):
This is a great start at a more generalized expansion. Let's chat about it more tonight.

lit_tests/kernel/wave/expansion.py
martin-luecke: fixed 👍

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -143,6 +167,40 @@ def _add_proxy_to_graph(self, region_graph: RegionGraph):
...
+            idx = idx_or_name
+        if isinstance(value, CustomOp):
+            value = value.fx_node
+        # Skip the fields defined by the abstract base class
+        if 0 <= idx < len(dataclass_fields):
+            self.node_args[idx] = value  # TODO: Does this make sense?
+            field_name = dataclass_fields[idx].name
+            # Set the new value for the field
+            setattr(self, field_name, value)
+            fx_val = value.fx_node if isinstance(value, CustomOp) else value
```
harsh-nod, line 190: In line 182, in the instance of a CustomOp, value is set to value.fx_node. Isn't that going to be a problem here because value.fx_node.fx_node is not defined?

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -143,6 +167,40 @@ def _add_proxy_to_graph(self, region_graph: RegionGraph):
             kwargs={},
         )
 
+    def update_arg(self, idx_or_name: int | str, value: Any):
+        """
+        Update the value of an argument in the node while keeping the
+        underlying fx.Node consistent.
+        """
+        dataclass_fields = fields(self)[5:]
```
harsh-nod, line 175: Can you add a comment about the magic number 5 here? Also, looks like we only use the name member variable of  dataclass_fields so we could just do
```
field_names = [x.name for x in fields(self)[5:]]
```

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -143,6 +167,40 @@ def _add_proxy_to_graph(self, region_graph: RegionGraph):
...
+        if isinstance(idx_or_name, str):
+            if idx_or_name not in [field.name for field in dataclass_fields]:
+                raise ValueError(f"Field {idx_or_name} not found")
+            idx = [field.name for field in dataclass_fields].index(idx_or_name)
+        else:
+            idx = idx_or_name
+        if isinstance(value, CustomOp):
+            value = value.fx_node
+        # Skip the fields defined by the abstract base class
+        if 0 <= idx < len(dataclass_fields):
```
harsh-nod, line 185: Pretty cool, didn't know you can chain inequalities like this :)

harsh-nod (approved):
Some minor comments, but should be good to go after those. Thanks again for getting this generalized expansion going. Much much better than the old "c_reg" way :)

shark_turbine/kernel/wave/expansion.py
```diff
@@ -0,0 +1,323 @@
+import itertools
+from sympy import Symbol
```
harsh-nod, line 2: is this import being used?

shark_turbine/kernel/ops/wave_ops.py
martin-luecke: Ah, I check whether value is a CustomOp twice.
This was no problem because at the second instance `isinstance(value, CustomOp)` will always evaluate to false because of line 182, where we set `value = value.fx_node` in that case. 
I removed the second check

shark_turbine/kernel/wave/expansion.py
martin-luecke: I removed it 👍

shark_turbine/kernel/ops/wave_ops.py
martin-luecke: Thanks for pointing me to this again. This magic number skips over the inherited fields from the `CustomOp` superclass. The problem is that this number needs to be adjusted when new fields are added to the superclass. This bugged me a few times already so I implemented this now without the need for a magic number.

Thanks for the `field_names` suggestion 👍

PR #42, opened 2024-07-12
Add shared memory promotion patterns

harsh-nod:
This PR adds patterns to promote global memory
reads to shared memory reads. Also adds a pattern
to hoist allocation nodes outside of reduction nodes.

### DIFF
--- a/lit_tests/kernel/wave/promotion.py
+++ b/lit_tests/kernel/wave/promotion.py
@@ -1,0 +1,226 @@
+# RUN: python %s | FileCheck %s
+
+import logging
+from typing import Callable
+import unittest
+import shark_turbine.kernel as tk
+import shark_turbine.kernel.lang as tkl
+import shark_turbine.kernel.wave as tkw
+from shark_turbine.kernel.wave.promotion import promote_node
+from shark_turbine.kernel.wave.hoisting import hoist_allocs
+from shark_turbine.kernel.wave.address_spaces import *
+from shark_turbine.kernel._support.tracing import CapturedTrace
+from shark_turbine.kernel._support.indexing import IndexingContext
+from shark_turbine.kernel.ops.wave_ops import *
+
+
+def run(func: Callable[[], None]) -> Callable[[], None]:
+    """Run a function as part of the test suite."""
+    if __name__ == "__main__":
+        func()
+        # Print a separator between tests
+        print("-----")
+    return func
+
+
+def get_read_nodes(graph: fx.Graph) -> list[fx.Node]:
+    nodes: list[fx.Node] = list(graph.nodes)
+    return [node for node in nodes if hasattr(node, "tkw_op") and node.tkw_op == Read]
+
+
+def print_trace(trace: CapturedTrace):
+    """
+    Prints all subgraphs of a trace starting with the root graph.
+    The graphs are printed first in the torch printing format and then using
+    our custom node format.
+    """
+    # The root graph is at the back so we print the subgraphs in reverse order
+    for subgraph in reversed(list(trace.region_graph.subgraphs.values())):
+        print(subgraph)
+
+
+# Input sizes
+M = tkl.sym.M
+N = tkl.sym.N
+K = tkl.sym.K
+
+# Workgroup tile sizes
+BLOCK_M = tkl.sym.BLOCK_M
+BLOCK_N = tkl.sym.BLOCK_N
+BLOCK_K = tkl.sym.BLOCK_K
+
+# Address space (for GPU, shared(1) or global(0))
+ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
+
+
+@tkw.wave_trace_only()
+def read_write_same_size(
+    a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+    c: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f32],
+):
+    a_reg = tkw.read(a, elements_per_thread=4)
+    tkw.write(a_reg, c, elements_per_thread=4)
+
+
+@run
+def test_read_write_equal_sizes():
+    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+    constraints += [
+        tkw.HardwareConstraint(
+            threads_per_wave=64,
+            waves_per_block=(1, 1, 1),
+            mma_type=tkw.MMAType.F32_16x16x16_F16,
+        )
+    ]
+
+    with tk.gen.TestLaunchContext(
+        {
+            BLOCK_M: 32,
+            BLOCK_N: 32,
+            ADDRESS_SPACE: GLOBAL_ADDRESS_SPACE,
+        }
+    ):
+        trace: CapturedTrace = read_write_same_size()
+        graph: fx.Graph = trace.get_root_graph()
+        read_node = get_read_nodes(graph)[0]
+        IndexingContext.current().finalize()
+        promote_node(read_node, graph, SHARED_ADDRESS_SPACE)
+        print_trace(trace)
+        # CHECK: %a
+        # CHECK-NEXT: %c
+        # CHECK-NEXT: %read_0_0
+        # CHECK-SAME: (%a, 4)
+        # CHECK-NEXT: %read_1_1
+        # CHECK-SAME: (%a, 4)
+        # CHECK-NEXT: %read_1_0
+        # CHECK-SAME: (%a, 4)
+        # CHECK-NEXT: %read_0_1
+        # CHECK-SAME: (%a, 4)
+        # CHECK-NEXT: %write_0_0
+        # CHECK-SAME: (%read, %c, 4)
+        # CHECK-NEXT: %write_1_1
+        # CHECK-SAME: (%read_1_1, %c, 4)
+        # CHECK-NEXT: %write_1_0
+        # CHECK-SAME: (%read_1_0, %c, 4)
+        # CHECK-NEXT: %write_0_1
+        # CHECK-SAME: (%read_0_1, %c, 4)
+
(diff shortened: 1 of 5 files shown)


martin-luecke (changes requested):
Thank you for the PR! 

Just a few comments about use of the interfaces. Ideally, I think we should aim to provide a base foundation where we only pass CustomOps in the interfaces

shark_turbine/kernel/wave/promotion.py
```diff
@@ -0,0 +1,46 @@
...
+            memory, elements_per_thread
+        ) if memory.type.address_space != allocate_node.address_space:
+            promoted_write = Write(
+                custom_node.fx_node, allocate_node.fx_node, elements_per_thread
+            ).add_to_graph(graph)
+            promoted_read = Read(
+                allocate_node.fx_node, elements_per_thread
+            ).add_to_graph(graph)
+            promoted_nodes = [promoted_write, promoted_read]
+            custom_node.fx_node.replace_all_uses_with(promoted_read)
```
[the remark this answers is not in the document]
harsh-nod: Good point. I realized that list wasn't being used for anything so I just removed it.

lit_tests/kernel/wave/promotion.py
```diff
@@ -0,0 +1,178 @@
...
+    if __name__ == "__main__":
+        func()
+        # Print a separator between tests
+        print("-----")
+    return func
+
+
+def get_read_nodes(graph: fx.Graph) -> list[fx.Node]:
+    custom_nodes: list[CustomOp] = [get_custom(node) for node in graph.nodes]
+    return [node for node in custom_nodes if isinstance(node, Read)]
```
martin-luecke, line 28: The return type annotation should be `list[CustomOp]`

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -198,6 +206,25 @@ def copy(self, new_name: Optional[str] = None) -> Self:
...
 
+    def copy_to_new_graph(
+        self, new_graph: fx.Graph, new_name: Optional[str] = None
+    ) -> Self:
+        """Returns a duplicate of this node."""
+        new_node = new_graph.node_copy(self.fx_node)
+        new_node.tkw_op = self
+        if new_name:
+            new_node.name = new_name
+        return get_custom(new_node)
```
martin-luecke, line 217: Would it make sense to add the `new_graph` as optional argument to our existing `copy` function above?

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -198,6 +206,25 @@ def copy(self, new_name: Optional[str] = None) -> Self:
...
+
+    def replace_all_uses_with(self, new_node: CustomOp):
+        """Replace all uses of the current node with the new node."""
+        for user in self.users:
+            user.update_arg(user.node_args.index(self), new_node)
+
+    def erase(self):
+        """Erase the current node from the graph where it exists."""
+        self.graph.erase_node(self.fx_node)
```
martin-luecke, line 226: Eventually we should add additional safety here, e.g. printing a warning if this node still has life uses

shark_turbine/kernel/wave/hoisting.py
```diff
@@ -0,0 +1,31 @@
...
+
+logger = get_logger("turbine.wave.hoisting")
+
+
+def get_allocs_(graph: fx.Graph) -> list[CustomOp]:
+    return [
+        custom_node
+        for node in graph.nodes
+        if isinstance((custom_node := get_custom(node)), Allocate)
+    ]
```
martin-luecke, line 15: nit: If we provide a return type annotation it should be as strict as possible, i.e: `list[Allocate]`

Why does the name have a trailing `_`?

shark_turbine/kernel/wave/pattern_rewriter.py
martin-luecke, line 1: Is this empty file required as part of this PR?

shark_turbine/kernel/wave/promotion.py
```diff
@@ -0,0 +1,41 @@
...
+from shark_turbine.kernel._support.tracing import CapturedTrace
+import torch.fx as fx
+from ..ops.wave_ops import *
+from .address_spaces import *
+import shark_turbine.kernel.lang as tkl
+
+logger = get_logger("turbine.wave.promotion")
+
+
+def apply_promotion_pattern_(custom_node: Read | Write, allocate_node: Allocate):
```
martin-luecke, line 11: nit: Why does the name have a trailing _?

shark_turbine/kernel/wave/promotion.py
```diff
@@ -0,0 +1,41 @@
...
+    match custom_node:
+        case Read(
```
martin-luecke, line 13: We could add a `case _: ` to report that we encountered a node we do not know how to do promotion for.
Is it correct that we expect `custom_node: Read | Write` but do not handle the `Write` case in the match?

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -198,6 +206,25 @@ def copy(self, new_name: Optional[str] = None) -> Self:
...
+        new_node = new_graph.node_copy(self.fx_node)
+        new_node.tkw_op = self
+        if new_name:
+            new_node.name = new_name
+        return get_custom(new_node)
+
+    def replace_all_uses_with(self, new_node: CustomOp):
+        """Replace all uses of the current node with the new node."""
+        for user in self.users:
+            user.update_arg(user.node_args.index(self), new_node)
```
martin-luecke, line 222: The `update_arg` works both with instances of `CustomOp` and `fx.Node` as new values. Hence, we could change the signature to `def replace_all_uses_with(self, new_node: CustomOp | fx.Node):` and explicitly support that as well. In fact, we use this below already in `custom_node.replace_all_uses_with(promoted_read)` where `promoted_read` is an instance of `fx.Node`

shark_turbine/kernel/wave/promotion.py
```diff
@@ -0,0 +1,41 @@
...
+def promote_node(node: CustomOp, address_space: IndexSymbol):
+    """Promotes the given operand in the provided graph
+    to the specified address space.
+
+    The process of promotion involves allocating memory
+    in the new address space and writing to the new
+    memory location and subsequent uses reading from there.
+    """
+
+    assert isinstance(node, Read) or isinstance(node, Write)
```
martin-luecke, line 35: This assert could also be reflected in the signature, e.g. `def promote_node(node: Read | Write, address_space: IndexSymbol):`

lit_tests/kernel/wave/promotion.py
```diff
@@ -0,0 +1,178 @@
...
+        # CHECK: %a
+        # CHECK-NEXT: %b
+        # CHECK-NEXT: %c
+        # CHECK-NEXT: %register
+        # CHECK-NEXT: %allocate
+        # CHECK-SAME: ((M, K), f16, SHARED_ADDRESS_SPACE)
+        # CHECK-NEXT: %allocate_1
+        # CHECK-SAME: ((N, K), f16, SHARED_ADDRESS_SPACE)
+        # CHECK-NEXT: reduction
+
```
martin-luecke, line 157: After the reduction I have an additional `%write(%reduction, %c, 4)`
If that is expected we should add:
```suggestion
        # CHECK-SAME: ((N, K), f16, SHARED_ADDRESS_SPACE)
        # CHECK-NEXT: reduction
        # CHECK-NEXT: %write
        # CHECK-SAME: (%reduction, %c, 4)

```

martin-luecke (approved):
LGTM after addressing the remaining few questions and nits

lit_tests/kernel/wave/promotion.py
harsh-nod: good catch! Thanks!

shark_turbine/kernel/ops/wave_ops.py
harsh-nod: That's also what I was wondering but decided to keep it separate to hear back from you during review. Thanks for the clarification. I will update that function and remove this new one.

shark_turbine/kernel/ops/wave_ops.py
harsh-nod: I can add a warning to handle that case.

shark_turbine/kernel/wave/hoisting.py
harsh-nod: Good point, will update. I was trying to follow your convention which I thought was to mark functions not to be exposed with a trailing underscore, but maybe that wasn't the intention. What was your convention? I can remove the underscore here.

shark_turbine/kernel/wave/pattern_rewriter.py
harsh-nod: I didn't even realize it was here. Will remove it.

shark_turbine/kernel/wave/promotion.py
harsh-nod: Same as above, will change.

shark_turbine/kernel/wave/promotion.py
harsh-nod: Yes, that's correct. Writes can also be promoted, I just haven't implemented that yet. I will add a case to handle that.

shark_turbine/kernel/ops/wave_ops.py
harsh-nod: Cool, that's a nice side effect :) Will update the signature.

lit_tests/kernel/wave/promotion.py
harsh-nod: Yes that is expected, will add that in. Thanks!

shark_turbine/kernel/wave/hoisting.py
martin-luecke: Ah, I see! I use trailing `_` only when I need it to avoid a name clash with an existing Python keyword.
The Python convention is that private functions should be prefixed with `_`.

PR #43, opened 2024-07-12
Add f8 types to AOT tracing

rsuderman:


stellaraccident (approved):
Let's see if this breaks anything. I think these are fine, but sometimes new types need to be conditionally added to not break old versions. But these have been here for a while

PR #44, opened 2024-07-16
Add attention example in tracing test

harsh-nod:
This PR adds an FP8 attention kernel to the
tracing test to validate that it traces correctly.

### DIFF
--- a/lit_tests/kernel/wave/tracing.py
+++ b/lit_tests/kernel/wave/tracing.py
@@ -233,3 +233,129 @@ def test_trace_gemm():
     # CHECK-NEXT: register
     # CHECK-NEXT: mma
     # CHECK-NEXT: output
+
+
+# Input sizes
+B = tkl.sym.M
+D = tkl.sym.N
+S = tkl.sym.K
+
+# Workgroup tile sizes
+BLOCK_B = tkl.sym.BLOCK_B
+BLOCK_S = tkl.sym.BLOCK_S
+
+
+@run
+def test_attention():
+    @tkw.wave_trace_only()
+    def attention(
+        query: tkl.Memory[B, N, D, ADDRESS_SPACE, tkl.f8e4m3fnuz],
+        key: tkl.Memory[B, S, D, ADDRESS_SPACE, tkl.f8e4m3fnuz],
+        value: tkl.Memory[B, S, D, ADDRESS_SPACE, tkl.f8e4m3fnuz],
+        output: tkl.Memory[B, S, D, ADDRESS_SPACE, tkl.f32],
+    ):
+        s_reg = tkl.Register[B, N, S, tkl.f32](0.0)
+        l_reg = tkl.Register[B, N, tkl.f32](0.0)
+        m_reg = tkl.Register[B, N, tkl.f32](-1e3)
+        o_reg = tkl.Register[B, N, D, tkl.f32](0.0)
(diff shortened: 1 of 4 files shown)


shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -38,6 +38,26 @@ def allocate(
     ...
 
 
+def cast(register: "Register", type: DataType) -> "Register":
+    ...
+
+
+def div(lhs: "Register", rhs: "Register") -> "Register":
+    ...
```
martin-luecke, line 46: The predeclarations here do not have CustomOps associated with them
We should map the `div` here to the Python operator like [here](https://github.com/iree-org/iree-turbine/blob/44c7634c3a0d4e26a8b31fc1ffa5bbb2cc8fbfcd/shark_turbine/kernel/ops/wave_ops.py#L310). The branch needs to be rebased for that

lit_tests/kernel/wave/tracing.py
```diff
@@ -233,3 +233,129 @@ def repeat(acc) -> tkl.Register[M, N, tkl.f32]:
...
+    # CHECK-NEXT: placeholder
+    # CHECK-NEXT: placeholder
+    # CHECK-NEXT: read
+    # CHECK-NEXT: placeholder
+    # CHECK-NEXT: placeholder
+    # CHECK-NEXT: mma
+    # CHECK-NEXT: unknown: sub
+    # CHECK-NEXT: unknown: sub_1
+    # CHECK-NEXT: unknown: mul
+    # CHECK-NEXT: unknown: add
```
martin-luecke, line 357: These will become proper sub, mul and add after rebasing on main

lit_tests/kernel/wave/tracing.py
```diff
@@ -233,3 +233,129 @@ def repeat(acc) -> tkl.Register[M, N, tkl.f32]:
...
+    # CHECK-NEXT: placeholder
+    # CHECK-NEXT: register
+    # CHECK-NEXT: register
+    # CHECK-NEXT: register
+    # CHECK-NEXT: register
+    # CHECK-NEXT: read
+    # CHECK-NEXT: reduction
+    # CHECK-NEXT: unknown: getitem
+    # CHECK-NEXT: unknown: getitem_1
+    # CHECK-NEXT: unknown: getitem_2
```
martin-luecke, line 322: These `getitem` nodes stem from the tuple unpacking here: `final_sum, _, final_output = repeat`
We should think about how we want to handle these. I think for now it might make sense to intercept tracing for `getitem` on a CustomOp and emit a `getResult` op instead. 
At least that makes sense in this case where the results of the reduction are accessed. I don't think we want to support this unpacking for general nodes so we should add further checks on when this is valid.

Anyway, I like the syntax you have here. Feels Pythonic to me

martin-luecke (changes requested):
This is a good demonstrator for how attention could look in TKW.

This needs more iteration and requires features we do not yet support like tuple unpacking.
I think we do not need to have an end-to-end flow as prerequisite to add this tracing test, but we should at least rebase to use the Python operators we have and have a rough plan for supporting the assignments here through expansion + scheduling.

Let's for now convert this to a draft PR

harsh-nod:
This is outdated with all the latest changes in attention.

PR #46, opened 2024-07-18
TKW: Add support for representing python operators as CustomOps

martin-luecke:
This overloads python operators we are interested in in the fx.Proxy class to control their tracing.
Registering a python operator using the new `define_py_op` ensures that it is mapped to a corresponding subclass of e.g. `BinaryPyOp` for `+`. These subclasses are created on the fly for the operators we register. Adding new operators requires simply adding an additional decorator to `BinaryPyOp` or `UnaryPyOp` (depending on number of args) and writing the appropriate codegen handler. 
These operations are then properly handled by the rest of the system (e.g. expansion).

This PR also adds tests for tracing, expansion and preliminary codegen tests (to check that the codegen handlers link up properly).

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -7,10 +7,12 @@ import shark_turbine.kernel.lang as tkl
 import shark_turbine.kernel.wave as tkw
 import torch
 
-
 M = tkl.sym.M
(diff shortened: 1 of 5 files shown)


martin-luecke:
Superseded by: https://github.com/iree-org/iree-turbine/pull/54

PR #53, opened 2024-07-26
Add TilingConstraint to represent non-distributed tiled loops

harsh-nod:
This PR adds a tiling constraint primitive to the language. This is used to represent constraints on how a dimension is to be tiled. In contrast to the workgroup constraints, dimensions with a tiling constraint on them are not distributed and materialized to sequential loops.

This PR also adds support for a third workgroup dimension and deduces the mapping of user dimensions to MMA dimensions based on the operands of the tkw.mma operator.

### DIFF
--- a/shark_turbine/kernel/wave/constraints.py
+++ b/shark_turbine/kernel/wave/constraints.py
@@ -46,6 +46,7 @@ class HardwareConstraint(Constraint):
     threads_per_wave: int
     waves_per_block: Optional[tuple[int, int, int]] = None
     mma_type: Optional[MMAType] = MMAType.F32_16x16x16_F16
+    vector_shapes: Optional[dict[IndexExpr, int]] = None
 
     @property
     def mma_matrix_shapes(self):
@@ -79,11 +80,13 @@ class WorkgroupConstraint(Constraint):
     def apply(self) -> IndexExpr:
         match self.workgroup_dim:
             case 0:
-                wg_dim = sym.WG0
+                wg_dim = tkl.sym.WG0
             case 1:
-                wg_dim = sym.WG1
+                wg_dim = tkl.sym.WG1
+            case 2:
+                wg_dim = tkl.sym.WG2
             case _:
-                raise ValueError("Invalid workgroup dimension. Expected 0 or 1.")
+                raise ValueError("Invalid workgroup dimension. Expected 0, 1 or 2.")
         return wg_dim * self.tile_size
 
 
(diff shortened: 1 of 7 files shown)


shark_turbine/kernel/wave/constraints.py
```diff
@@ -101,3 +104,31 @@ def get_grid_shape(wg_constraints: list[WorkgroupConstraint]) -> list[IndexExpr]
...
+
+    dim: IndexExpr
+    tile_size: IndexExpr
+    induction_var: Optional[IndexExpr] = None
+
+    def iterations(self) -> IndexExpr:
+        """
+        Returns an expression for the number of iterations in the loop.
+        """
+        return self.dim / self.tile_size
```
Hardcode84, line 127: I'm not very familiar with `tkw` internals yet, are you sure it should be `/` and not `//` or `ceilDiv`

harsh-nod: Good catch. Yes, I believe ceilDiv would be most appropriate here. Since sympy does not have this, I can wrap the above code with ceiling to give something like
```
return ceiling(self.dim / self.tile_size)
```
I will also add a test for this.

Hardcode84 (approved):

shark_turbine/kernel/wave/constraints.py
```diff
@@ -46,6 +45,7 @@ class HardwareConstraint(Constraint):
     threads_per_wave: int
     waves_per_block: Optional[tuple[int, int, int]] = None
     mma_type: Optional[MMAType] = MMAType.F32_16x16x16_F16
+    vector_shapes: Optional[dict[IndexExpr, int]] = None
```
raikonenfnu, line 48: Can we add description on the vector_shape to the comment on Hardwareconstraint? Is it vector shape of matmul per wave?

shark_turbine/kernel/wave/constraints.py
```diff
@@ -101,3 +103,31 @@ def get_grid_shape(wg_constraints: list[WorkgroupConstraint]) -> list[IndexExpr]
...
+
+    dim: IndexExpr
+    tile_size: IndexExpr
+    induction_var: Optional[IndexExpr] = None
+
+    def iterations(self) -> IndexExpr:
+        """
+        Returns an expression for the number of iterations in the loop.
+        """
+        return ceiling(self.dim / self.tile_size)
```
raikonenfnu, line 126: Would we handle the non-aligned-ness/padding of it somewhere else?

shark_turbine/kernel/wave/constraints.py
harsh-nod: Sure will add in a separate PR.

shark_turbine/kernel/wave/constraints.py
harsh-nod: Yes, this just gives the number of iterations.

shark_turbine/kernel/wave/constraints.py
harsh-nod: It is the shape of the matmul per wave if there is an mma operation in the graph. Otherwise it is user-settable.

PR #54, opened 2024-07-28
Add CustomOp support for python operators

harsh-nod:


lit_tests/kernel/wave/codegen.py
```diff
@@ -49,4 +56,84 @@ def test(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
...
+    constraints: list[tkw.Constraint] = [
+        tkw.HardwareConstraint(threads_per_wave=64, waves_per_block=(1, 1, 1))
+    ]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+
+    @tkw.wave(constraints)
+    def test(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
+        res = a[0]
+        tkw.write(res, a, elements_per_thread=4)
```
Hardcode84, line 130: So, what the exact semantics when we write lower-dimensional array (or scalar) into higher dimensional? Also, as we now have a simulator, we should probably test this case under simulator too (we can do it in separate PR).

Hardcode84 (approved):

lit_tests/kernel/wave/codegen.py
harsh-nod: I believe we want to follow numpy broadcasting semantics (https://numpy.org/doc/stable/user/basics.broadcasting.html). Even pytorch seems to be adhering to these. So specifically in this case, we extract a N-dim vector and when we write it back it is broadcasted to MxN. Regarding the simulator, makes sense. Would you like to try this or do you want me to?

Hardcode84: I already have tests for elementwise in my branch, so yeah, I think I can add tests for broadcasting too.

harsh-nod: thanks!

PR #55, opened 2024-07-28
Wave sim mapping support and igemm kernel

Hardcode84:
Add `mapping` argument to `read` and `write` ops and add support to the simulator.

* Add `IndexMapping` class which describes how to map between 2 sets of indices. Mapping consists of set of iterators and input/output dicts, which describe, how to map symbolic dimension to the specific iterator. Mapping is similar to `linalg.generic` index maps, where it's iterators corresponds to linalg `parallel` iterator. The difference is that we are binding iterator to dimensions symbols instead of dim index. E.g. if we have mapping like this:
```
    i = tkw.IndexMapping.iterator(0)
    j = tkw.IndexMapping.iterator(1)
    mapping = tkw.IndexMapping(
        num_iterators=2, inputs={N: i, M: j}, outputs={N: i, M: j}
    )
```
If the input array has shape `[N, M]`, we will copy a contiguous slice, but for the input shape `[M, N]` we will transpose input data.
If operation doesn't have output symbols by itself (i.e. `read`) the output will have same shape as `outputs` keys (e.g. `[N,M]` for prev example).
* Add simulator tests (simple transposes and implicit gemm conv example)

### DIFF
--- a/lit_tests/kernel/wave/expansion.py
+++ b/lit_tests/kernel/wave/expansion.py
@@ -88,13 +88,13 @@ def test_read_write_equal_sizes():
         # CHECK: %a
         # CHECK-NEXT: %c
         # CHECK-NEXT: %read_0_0
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_1_1
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_1_0
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_0_1
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %write_0_0
         # CHECK-SAME: (%read_0_0, %c, 4, None)
         # CHECK-NEXT: %write_1_1
@@ -166,9 +166,9 @@ def test_read_write():
         # CHECK: %a
         # CHECK-NEXT: %c
         # CHECK-NEXT: %read_0_0_0
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_1_0_0
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %write_0_0_0
         # CHECK-SAME: (%read_0_0_0, %c, 4, None)
         # CHECK-NEXT: %write_1_0_1
@@ -295,23 +295,23 @@ def test_gemm():
 
         # CHECK-NEXT: %a
         # CHECK-NEXT: %read_0_0_0
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_0_0_1
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_1_0_0
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_1_0_1
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
 
         # CHECK-NEXT: %b
         # CHECK-NEXT: %read_0_0_0
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_0_1
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_1_0
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_1_1
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
 
         # CHECK-NEXT: %mma_0_0_0
         # CHECK-SAME: (%read_0_0_0, %read_0_0_0, %acc_0_0_0)
@@ -436,31 +436,31 @@ def test_gemm_reduction_expansion_only():
 
         # CHECK-NEXT: %a
         # CHECK-NEXT: %read_0_0_0
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_0_0_1
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_0_0_2
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
         # CHECK-NEXT: %read_0_0_3
-        # CHECK-SAME: (%a, 4, None, None)
+        # CHECK-SAME: (%a, 4, None)
 
         # CHECK-NEXT: %b
         # CHECK-NEXT: %read_0_0_0
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_0_1
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_0_2
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_0_3
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_1_0
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_1_1
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_1_2
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
         # CHECK-NEXT: %read_0_1_3
-        # CHECK-SAME: (%b, 4, None, None)
+        # CHECK-SAME: (%b, 4, None)
 
         # CHECK-NEXT: %mma_0_0_0
         # CHECK-SAME: (%read_0_0_0, %read_0_0_0, %acc_0_0_0)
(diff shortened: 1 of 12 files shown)


tests/kernel/wave/wave_sim_test.py
```diff
@@ -210,3 +210,178 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+        we: tkl.Memory[NF, C, HF, WF, ADDRESS_SPACE, tkl.f16],
+        out: tkl.Memory[N, NF, H_OUT, W_OUT, ADDRESS_SPACE, tkl.f32],
+    ):
+        c_reg = tkl.Register[M, NF, tkl.f32](0.0)
+
+        @tkw.reduction(K, init_args=[c_reg])
+        def repeat(acc: tkl.Register[M, NF, tkl.f32]) -> tkl.Register[M, NF, tkl.f32]:
+            a_reg = tkw.read(
+                x,
+                mapping=x_mapping,
```
harsh-nod, line 368: Something that is a little confusing here, we have x_mapping which maps a 2d index to a 4d index but the shape is only 2D. What are the semantics of mapping and shape?

tests/kernel/wave/wave_sim_test.py
```diff
@@ -210,3 +210,178 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+
+    # Expose user-constraints
+    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+
+    constraints += [
+        tkw.HardwareConstraint(threads_per_wave=64, waves_per_block=(1, 1, 1))
+    ]
+
+    mapping = tkw.IndexMapping(lambda i, j: (j, i))
```
harsh-nod, line 236: I like this example. It is a good showcase of mapping. In this situation, do we need to specify the shape? Maybe this is a general question, but with the mapping specification isn't the shape information not required?

lit_tests/kernel/wave/expansion.py
```diff
@@ -85,21 +85,21 @@ def test_read_write_equal_sizes():
         # CHECK: %a
         # CHECK-NEXT: %c
         # CHECK-NEXT: %read_0_0
```
harsh-nod, line 87: Can we avoid printing the mapping and shape when it is None?

shark_turbine/kernel/lang/wave_types.py
```diff
@@ -109,3 +110,11 @@ def __class_getitem__(
             symbolic_shape=shape,
             dtype=dtype,
         )
+
+
+class IndexMapping:
+
+    mapping_func: ClassVar[Callable[..., Any]]
```
harsh-nod, line 117: Won't this always be a mapping from a tuple of IndexExpr to a tuple of IndexExpr?

shark_turbine/kernel/wave/wave_sim.py
```diff
@@ -184,14 +190,38 @@ def decorator(func: Callable[..., Any]) -> Any:
...
-    memory: "Memory", elements_per_thread: Optional[IndexExpr] = None
+    memory: "Memory",
+    elements_per_thread: Optional[IndexExpr] = None,
+    mapping: Optional[IndexMapping] = None,
+    shape: Optional[tuple[IndexExpr, ...]] = None,
 ) -> "Register":
+    if mapping:
+        assert shape
+        mapping_func = mapping.mapping_func
+        res = torch.zeros(shape)
```
harsh-nod, line 201: So for example here, couldn't the shape be inferred from the shape of memory and the mapping_func ?

harsh-nod (commented):
This looks good for the most part. I'm a little unclear on the semantics of mapping and shape (2D to 4D) and whether we need to specify shape at all ?

lit_tests/kernel/wave/expansion.py
Hardcode84: Yeah, good idea.

shark_turbine/kernel/lang/wave_types.py
Hardcode84: Yeah, will update

tests/kernel/wave/wave_sim_test.py
Hardcode84: In this case, we are mapping from 4d (N, C, H, W) input to 2d (M, K) output, so mapping func takes 2 arguments (output indices) and returns 4 results (input indices). It's a good question, actually, do we want to map indices in `out->in` or `in->out` order, I choose current order because I think it will be more straightforward to handle in simulator and to translate to `vector.gather`.

Regarding question if we need an explicit shape argument, we will need to invert this mapping function to find output shape from the input shape, which should be possible in most cases, but not always.

One interesting case is manual broadcasting specific row/column:
```python
mapping = IndexMapping(lambda i, j: (i, 2))
a_reg = read(a, mapping=mapping, shape=...)
```
In this case we cannot infer resulting shape from input and mapping alone.
(We don't need this for the igemm, though)

lit_tests/kernel/wave/expansion.py
Hardcode84: I can update `CustomOp` printing, but I can't do anything with `%read_3 : [num_users=1] = call_function[target=shark_turbine.kernel.ops.wave_ops.read](args = (%allocate_1, 4, None, None), kwargs = {})` form.

Hardcode84: https://github.com/iree-org/iree-turbine/pull/64

tests/kernel/wave/wave_sim_test.py
harsh-nod: Okay that makes sense. I think in general indices from in -> out would make more sense, but I can see how some expressions would be easier to write from out -> in than the other way around. Okay thinking about this more, I was wondering whether you could incorporate the shape information within the IndexMapping class as a type which specifies what the index mapping does: takes indices from shape [A, B] -> [U, V, W, X] and this way we could kill two birds with one stone. You can deduce in -> out or out -> in from the types and we avoid the explicit shape argument in the read, instead keeping it inside the index mapping. What do you think?

lit_tests/kernel/wave/expansion.py
harsh-nod: Thanks!

tests/kernel/wave/wave_sim_test.py
Hardcode84: Updated, mapping API following offline discussion, explicit `shape` is no longer needed as it encoded directly in mapping object now.

harsh-nod: Awesome, thanks! Taking a look now.

tests/kernel/wave/wave_sim_test.py
```diff
@@ -210,3 +210,192 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+    W_OUT = (W + 2 * padding - WF) // stride + 1
+    SZ_OUT = H_OUT * W_OUT
+
+    K = HF * WF * C
+    M = SZ_OUT * N
+
+    i = tkw.IndexMapping.iterator(0)
+    j = tkw.IndexMapping.iterator(1)
+
+    x_mapping = tkw.IndexMapping(
```
harsh-nod, line 327: Just confirming here that the reason we need to specify num_dims is because in general we cannot deduce which variables are iterators and which ones are not?

shark_turbine/kernel/wave/wave.py
```diff
@@ -15,7 +15,7 @@
 )
 from .codegen import WaveEmitter
 from .expansion import expand_graph
-from ..lang import Grid
+from ..lang import Grid, IndexMapping
```
harsh-nod, line 18: is this import used anywhere here?

shark_turbine/kernel/wave/wave_sim.py
```diff
@@ -33,12 +35,16 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
             global _api_subs
+            global _symbolic_shapes
             subs = copy.copy(_api_subs)
             if args_handler:
                 args_handler(args, subs)
 
             new_func = _resolve_symbols(f, subs)
-            return new_func(*args)
+            try:
+                return new_func(*args)
```
harsh-nod, line 45: Why would returning new_func(*args) fail?

shark_turbine/kernel/wave/wave_sim.py
```diff
@@ -184,19 +221,71 @@ def decorator(func: Callable[..., Any]) -> Any:
...
+            dst[res_mapped] = src[inp_mapped]
+
+        return
+
     dst[:] = src
 
 
 def _mma_proxy(a: "Register", b: "Register", acc: "Register") -> "Register":
-    return torch.matmul(a, b.T) + acc
+    a_shape = _get_symbolic_shape(a)
```
harsh-nod, line 284: Do we need to set a_shape, b_shape and res shape? Doesn't seem like they are used here?

lit_tests/kernel/wave/expansion.py
```diff
@@ -88,21 +88,21 @@ def test_read_write_equal_sizes():
...
-        # CHECK-SAME: (%a, 4)
+        # CHECK-SAME: (%a, 4, None)
```
harsh-nod, line 91: I thought we didn't need to print None anymore?

harsh-nod (approved):
Minor comments, but overall looks good. I like the new index mapping and think this could be a great new powerful construct for the language.

harsh-nod (approved):
Minor comments, but overall looks good. I like the new index mapping and think this could be a great new powerful construct for the language.

tests/kernel/wave/wave_sim_test.py
Hardcode84: Technically, we can remove it and just enumerate all iterators both in `input` and `output` instead (as each iterator must be used at least once), but I think it's more user readable this way. (Also, after some thought, `num_iterators` will be a better name).

shark_turbine/kernel/wave/wave_sim.py
Hardcode84: We are calling `new_func` here, so I want cleanup to happen even if it fails for whatever reason.

shark_turbine/kernel/wave/wave_sim.py
Hardcode84: Th reason I need those `_get/_set_symbolic_shape` calls, because memory/register arrays are represented by `torch.tensor` and getting shape from it will return actual shape, but I also want to know original symbolic shapes that were associated with those arrays. For now, I'm just maintaining a separate map `id(object) -> symbolic_shape` for simplicity. Another approach will be to create a wrapper on top of `torch.tensor` and keep symbolic shape there.

lit_tests/kernel/wave/expansion.py
Hardcode84: We don't print `None`s when we are pretty-printing `CustomOp`s, but those tests are printing IR in `%read_3 : [num_users=1] = call_function[target=shark_turbine.kernel.ops.wave_ops.read](args = (%allocate_1, 4, None, None), kwargs = {})` form.

shark_turbine/kernel/wave/wave.py
Hardcode84: It needed for `tkw.IndexMapping` to work.

shark_turbine/kernel/wave/wave_sim.py
harsh-nod: Sounds good, thanks!

lit_tests/kernel/wave/expansion.py
harsh-nod: Ah makes sense. Thanks!

PR #56, opened 2024-07-29
Add elementwise and broadcast simulator tests

Hardcode84:
As I'm using `torch.tensor` in simulator, all elementwise ops and broadcasting should work out of the box.

### DIFF
--- a/tests/kernel/wave/wave_sim_test.py
+++ b/tests/kernel/wave/wave_sim_test.py
@@ -6,6 +6,152 @@ from shark_turbine.kernel.wave.wave_sim import wave_sim
 from numpy.testing import assert_allclose
 
 
+def test_eltwise():
+    # Input sizes
+    M = tkl.sym.M
+    N = tkl.sym.N
+    # Workgroup tile sizes
+    BLOCK_M = tkl.sym.BLOCK_M
+    BLOCK_N = tkl.sym.BLOCK_N
+    # Address space (for GPU, shared(1) or global(0))
+    ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
+    # Other hyperparameters
+    LOAD_ELEMS_PER_THREAD = tkl.sym.LOAD_ELEMS_PER_THREAD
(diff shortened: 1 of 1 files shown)


tests/kernel/wave/wave_sim_test.py
```diff
@@ -6,6 +6,152 @@
...
+    # Other hyperparameters
+    LOAD_ELEMS_PER_THREAD = tkl.sym.LOAD_ELEMS_PER_THREAD
+    STORE_ELEMS_PER_THREAD = tkl.sym.STORE_ELEMS_PER_THREAD
+
+    # Expose user-constraints
+    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+
+    constraints += [
+        tkw.HardwareConstraint(threads_per_wave=64, waves_per_block=(1, 1, 1))
```
harsh-amd, line 27: I added a vector_shapes arg which specifies the "intrinsic" tile size that will probably be useful for the simulator when we simulate things at that level. Right now, I think the simulator ignores that, but just fyi.

tests/kernel/wave/wave_sim_test.py
```diff
@@ -6,6 +6,152 @@
...
+
+    a = torch.randn(128, 256, dtype=torch.float32)
+    b = torch.randn(128, 256, dtype=torch.float32)
+    c = torch.zeros(128, 256, dtype=torch.float32)
+    eltwise(a, b, c)
+    assert_allclose(c, a + b)
+
+
+def test_broadcast_1():
+    # Input sizes
```
harsh-amd, line 48: Nice!

harsh-amd (approved):
Thanks for adding these tests. Great that broadcasting just works out of the box.

tests/kernel/wave/wave_sim_test.py
Hardcode84: Yeah, at this point simulator completely ignores the `constraints` arg

harsh-nod (approved):
Thanks! LGTM!

PR #59, opened 2024-07-31
Add indexing to nodes

harsh-nod:
This primary purpose of this PR is to annotate nodes with their access patterns based on the workgroup, tiling and MMA constraints. This is accomplished prior to expansion and propagates through expansion to the expanded nodes.

### DIFF
--- a/shark_turbine/kernel/ops/wave_ops.py
+++ b/shark_turbine/kernel/ops/wave_ops.py
@@ -22,6 +22,7 @@ from .._support.indexing import IndexExpr, IndexSymbol
 from .._support.dtype import DataType
 from .._support.regions import RegionGraph
 from .base import OpDispatcher
+import shark_turbine.kernel.lang as tkl
 
 T = TypeVar("T", bound=Type[Any])
 AccT = TypeVar("AccT")
@@ -171,13 +172,15 @@ class CustomOp(ABC):
     fx_node: Optional[fx.Node] = field(default=None, init=False)
     tkw_op_name: str = field(default="unknown", init=False)
     _tracing_function: Optional[Callable[..., Any]] = field(default=None, init=False)
-    index: Optional[IndexExpr] = field(default=None, init=False)
+    index: Optional[dict[IndexSymbol, IndexSequence]] = field(default=None, init=False)
 
     @classmethod
     def from_fx_node(cls: Type[CustomOpT], node: fx.Node) -> CustomOpT:
         instance = cls(*node.args)
         instance.fx_node = node
         instance.graph = node.graph
+        if hasattr(node, "index"):
+            instance.index = node.index
         return instance
 
     def __post_init__(self):
@@ -198,7 +201,14 @@ class CustomOp(ABC):
 
     def custom_string(self, value_map: dict[str, str]) -> str:
         # print all variables of the node apart from graph and fx_node
-        vars_list = [f"{key}={value}" for key, value in vars(self).items()][:-2]
+        ignore_list = ["fx_node", "graph"]
+        if self.index is None:
+            ignore_list += ["index"]
+        vars_list = [
+            f"{key}={value}"
+            for key, value in vars(self).items()
+            if key not in ignore_list
+        ]
         vars_str = ", ".join(vars_list)
         return f"{self.tkw_op_name}({vars_str})"
 
@@ -213,6 +223,7 @@ class CustomOp(ABC):
         )
         self.fx_node.tkw_op = self.__class__
         self.fx_node.tkw_op_name = self.tkw_op_name
+        self.fx_node.index = None
         return self.fx_node
 
     def _add_proxy_to_graph(self, region_graph: RegionGraph):
@@ -259,6 +270,7 @@ class CustomOp(ABC):
             graph.inserting_after(self.fx_node)
         new_node = graph.node_copy(self.fx_node)
         new_node.tkw_op = self
+        new_node.index = self.fx_node.index
         if new_name:
             new_node.name = new_name
         return get_custom(new_node)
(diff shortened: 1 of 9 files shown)


shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -512,6 +524,40 @@ def rhs_type(self) -> Memory:
...
+        return self.operand_index(operand_map, self.lhs_type.symbolic_shape)
+
+    @property
+    def rhs_index(self) -> list[IndexSequence]:
+        operand_map = {tkl.sym.MMA_LHS: 0, tkl.sym.MMA_RHS: 1, tkl.sym.MMA_ACC: 0}
+        return self.operand_index(operand_map, self.rhs_type.symbolic_shape)
+
+    @property
+    def acc_index(self) -> list[IndexSequence]:
+        operand_map = {tkl.sym.MMA_LHS: 0, tkl.sym.MMA_RHS: 0, tkl.sym.MMA_ACC: 1}
```
raikonenfnu, line 547: Since we are subbing IndexExpr.sub by an operand_map, would the symbolic expression already have the `tkl.sym.MMA_LHS`, `tkl.sym.MMA_RHS`, and/or `tkl.sym.MMA_ACC` as part of the expression?

shark_turbine/kernel/wave/distribution_symbols.py
```diff
@@ -0,0 +1,9 @@
+import shark_turbine.kernel.lang as tkl
+
+WORKGROUP_0 = tkl.sym.WG0
```
raikonenfnu, line 3: Does WG0, WG1, Wg2 map to WG_X, WG_Y, WG_Z ?

shark_turbine/kernel/wave/indexing.py
```diff
@@ -0,0 +1,43 @@
...
+logger = get_logger("turbine.wave.indexing")
+
+
+@dataclass
+class IndexSequence:
+    start: IndexExpr | int
+    size: IndexExpr | int
+    stride: Optional[IndexExpr | int] = 1
+
+    def __add__(self, other: Any) -> Any:
```
raikonenfnu, line 16: Not really intuitive to me, when would we add index_sequence together? Can we add a small comment on that?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -120,6 +121,50 @@ def is_mma(node: fx.Node) -> bool:
...
+                    custom.index = {
+                        dim: IndexSequence(0, 1) for dim in custom.indexing_dims
+                    }
+                if isinstance(constraint, HardwareConstraint):
+                    if dim in mma_index and isinstance(custom, MMA):
+                        custom.index[dim] += constraint.apply(mma_index[dim])
+                elif dim == constraint.dim:
+                    custom.index[dim] += constraint.apply()
+        if custom.index:
+            setattr(custom.fx_node, "index", custom.index)
```
raikonenfnu, line 162: Does this mean during handle_op, we cshould be able to just access this "index" attribute? do we still need to take the size from elem_per_thread, or would it be handled somewhere here as well?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -120,6 +121,50 @@ def is_mma(node: fx.Node) -> bool:
...
+        for dim in custom.indexing_dims:
+            for constraint in constraints:
+                if (
+                    not isinstance(constraint, HardwareConstraint)
+                    and dim != constraint.dim
+                ):
+                    continue
+                if not custom.index:
+                    custom.index = {
+                        dim: IndexSequence(0, 1) for dim in custom.indexing_dims
```
raikonenfnu, line 154: Do we expect the handling of the not custom index and loop over `custom.indexing_dims` in this for loop? would it not make sense to handle this outside of `for dim in custom.indexing_dims:`?

shark_turbine/kernel/wave/wave.py
```diff
@@ -93,15 +128,20 @@ def _trace_and_get_kernel_signature(
...
         idxc = IndexingContext.current()
         idxc.finalize()
 
         # Expansion
         expand_graph(graph, self.constraints)
 
         kernel_sig = kernel_codegen.KernelSignature()
-        # Fixed values for now, will be determined through constraints
-        self.grid_type.dims = [32, 32]  # Will be determined by constraints
+        self.grid_type.dims = [1, 1, 1]
```
raikonenfnu, line 140: Nice! :)

shark_turbine/kernel/ops/wave_ops.py
harsh-nod: Yes, so this what I was talking about in the meeting. For each operator, we specify a dimension-specific index. So for MMA, a separate index of M, N and K. Rather than partition these dimensional indices further into operand specific indices, I have them represented as a piecewise function where the conditions depend on MMA_{LHS/RHS/ACC} (In the current PR, just ACC). So in order to extract the operand specific parts, we just substitute the appropriate values as above. The advantage of this piecewise function approach is it allows you to see where the dimensional mapping bifurcates and for which operands and allows you to reason about "layout changes". (For example, you could ask questions like - what setting of LHS, RHS and ACC would make the indices of MMA_0 be the same as that of the LHS of MMA1?)

shark_turbine/kernel/wave/distribution_symbols.py
harsh-nod: WG[ i ]  maps to stream dialect dispatch workgroup id [ i ] . I am not sure how stream dialect maps 0, 1, 2 to x,y,z. I think it maybe reversed?

shark_turbine/kernel/wave/indexing.py
harsh-nod: The index sequence represents an access pattern and this is attempting to define arithmetic on access patterns. Currently, this is only being used when computing indices from constraints. Not sure whether it will be used for other things. But the idea is this: Say our access pattern is for workgroup i to read from offset b * i and it reads 1 tile of size b with a stride of b. Next, wave j reads from offset c * j and it reads 4 elements with stride of 1. So when we combine these access patterns, we get that wave j needs to read 4 elements from b * i + c * j, with a stride of 1. But I do agree this is non-intuitive, so let me try and see if we can remove this altogether.

shark_turbine/kernel/wave/expansion.py
harsh-nod: My intention here was to only set the index attribute on operators that were affected by the constraints. The reason for this is to distinguish between the following 3 scenarios: an operator that has no index, an operator that has an index but the index is None and an operator that has an index that is not None. I dont think we need to distinguish between the first 2 options - so we can always set the index, but the index could be None. So with that I could move the initialization of custom_index outside the loop.

shark_turbine/kernel/wave/expansion.py
harsh-nod: Yes, you should be able to just access the index attribute, but you will have to check if it is None. You will still need to get the size from elem_per_thread. The index sequence tells you how many you can load but that could be different from how many the user has requested to load.

shark_turbine/kernel/wave/indexing.py
harsh-nod: Okay got rid of it in the latest patch.

raikonenfnu (approved):
LGTM!

shark_turbine/kernel/wave/distribution_symbols.py
```diff
@@ -0,0 +1,9 @@
...
+WORKGROUP_1 = tkl.sym.WG1
+WORKGROUP_2 = tkl.sym.WG2
+
+THREAD_0 = tkl.sym.T0
```
Hardcode84, line 7: General comment - I think, defining internal symbols in that manner may not be safe as user can accidentally define symbols with the same name in his code and break things in unexpected ways (especially for short and common names).
We probably need separate facility to create internal symbols while will never overlap with `tkl.sym` (e.g. by prepending sympy symbol name with "$").

harsh-nod: Good point. I will update this in the next PR.

Hardcode84 (approved):

PR #60, opened 2024-07-31
Add Codegen e2e flow

raikonenfnu:


shark_turbine/kernel/wave/codegen.py
```diff
@@ -43,6 +104,120 @@ def _emit_function_call_node(self, node: fx.Node):
 
         handler(self, node)
 
+    def lookup_node_values(self, node: fx.Node) -> List[Value]:
+        assert NDEBUG or isinstance(node, fx.Node)
```
harsh-nod, line 108: what is this NDEBUG assert for?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -1,13 +1,57 @@
...
+    # node attrs.
+    unsigned: bool = False
+
+    @staticmethod
+    def load(py_value) -> "NodeAttrs":
+        if isinstance(py_value, fx.Node):
+            return NodeAttrs(unsigned=bool(py_value.meta.get("unsigned")))
+        return NodeAttrs()
+
+    def store(self, node: fx.Node):
```
harsh-nod, line 53: Also instead of using the fx.Node here and below, can we use the CustomNode? The reason is that the CustomNode contains more information that is easily queryable as opposed to the fx.node.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -43,6 +104,120 @@ def _emit_function_call_node(self, node: fx.Node):
...
+    # hardware constraints later on.
+    def offset_fn(self, i):
+        return 0
+
+
+def gen_sympy_index(emitter: WaveEmitter, expr: sympy.Expr, stage: int) -> OpResult:
+    stack: list[OpResult] = []
+
+    # Induction var is accessed outside of the loop. Sets value to value of stage.
+    induction_var = sympy.symbols("ARG0")
```
harsh-nod, line 137: This is not going to be the induction variable and in general there could be many induction variables. I think you can get that information from the class in wave.py.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -43,6 +104,120 @@ def _emit_function_call_node(self, node: fx.Node):
...
+    induction_var_value = [arith_d.constant(IndexType.get(), stage)]
```
harsh-nod, line 138: Don't need stage for now.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -43,6 +104,120 @@ def _emit_function_call_node(self, node: fx.Node):
...
+    )
+
+    idxc = IndexingContext.current()
+    # Substitute in frozen vars to simplify expression.
+    if not isinstance(expr, sympy.Expr):
+        expr = sympy.sympify(expr)
+    expr = expr.subs(idxc.subs)
+    # Why affine, for now simply create indexing expressions.
+    # This can easily be adapted to affine expressions later.
+    for term in sympy.postorder_traversal(expr):
```
harsh-nod, line 156: I think this is okay for now, but let's think about how we could do this better. (in another PR)

harsh-nod (changes requested):
Awesome work! It's so nice to see IR being generated. Thanks for getting us here. Left a few comments. I think they are pretty minor.

shark_turbine/kernel/wave/codegen.py
raikonenfnu: Yeah, this is the part of the code I'd like to refacotr/clean up the most. :)

shark_turbine/kernel/wave/codegen.py
```diff
@@ -86,7 +286,33 @@ def handle_mma(emitter: WaveEmitter, node: fx.Node):
...
+        lhs, rhs = node.args
+    except ValueError as e:
+        raise ValidationError("Malformed arguments") from e
+    lhs = cast_py_value(emitter, lhs)
+    rhs = cast_py_value(emitter, rhs)
+
+    def get_type(value: IRProxyValue):
+        return value.ir_value.type
+
+    # TODO: this is too hard coded
```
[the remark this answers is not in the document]
raikonenfnu: sounds good! perhaps, I'll just catch and throw error then in case it manage to slip from FX. :)

shark_turbine/kernel/wave/codegen.py
```diff
@@ -1,13 +1,57 @@
...
+    cast_py_value,
+    cast_vector,
+)
+
+# Indexing imports.
+from .._support.indexing import IndexingContext
+
+
+@dataclass
+class NodeAttrs:
```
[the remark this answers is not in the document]
raikonenfnu: It used to be to keep track if something is unsigned or not, I don't think we need it for now, will remove.

shark_turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -189,6 +190,49 @@ def add_grid(self, grid: Type[Grid]):
...
+            for i, binding in enumerate(self.bindings):
+                if binding.reference[1] == node:
+                    index = i
+                    break
+            if index == None:
+                continue
+            # TODO: remove this hack, this is just to make things pass
+            # I did not investigate yet why it does not correctly determine the
+            # buffer to only have read dependencies, even though that is the case
+            usage = KernelBufferUsage.INPUT
```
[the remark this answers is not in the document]
raikonenfnu: Original hack from Martin is here https://github.com/harsh-nod/SHARK-Turbine/commit/767f7ecbabace15aa253437b786a89a256fe4d3f#diff-ed8153028c8d651d74c3da869ae799a720faa509ea86828e2beaec1a4fd42ac6R240-R244 

Seems like eventually he added a guard but still default to usage = KernelBufferUsage.INPUT. But I think the right thing to do is set bufferUsage to `KernelBufferUsage.NONE` if it is transient, and if it has both inputs and outputs then it may be transient/temporary buffer. 

Let me double check with Stella on this though.

shark_turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -189,6 +190,49 @@ def add_grid(self, grid: Type[Grid]):
...
             )
 
+    def determine_input_output_buffers(self, graph: fx.Graph):
+        placeholder_nodes: list[fx.Node] = []
+        for node in graph.nodes:
+            if node.op != "placeholder":
+                continue
+            placeholder_nodes.append(node)
+
+        def only_read_dependencies(node):
```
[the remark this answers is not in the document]
raikonenfnu: Done!

shark_turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -189,6 +190,49 @@ def add_grid(self, grid: Type[Grid]):
                 )
             )
 
+    def determine_input_output_buffers(self, graph: fx.Graph):
+        placeholder_nodes: list[fx.Node] = []
+        for node in graph.nodes:
+            if node.op != "placeholder":
+                continue
+            placeholder_nodes.append(node)
```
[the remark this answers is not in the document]
raikonenfnu: Done!

shark_turbine/kernel/wave/codegen.py
```diff
@@ -18,16 +62,33 @@ class WaveEmitter:
...
     OP_HANDLERS: ClassVar[dict[str, Callable[["WaveEmitter", fx.Node], None]]] = {}
+    _node_values: ClassVar[dict[fx.Node, List[IRProxyValue]]] = {}
 
     def __post_init__(self):
         self.ip = InsertionPoint(self.root_sig.entry_block)
 
+    def emit_program_invariants(self):
+        self.workgroup_ids = [
+            stream_d.dispatch_workgroup_id(IntegerAttr.get(IndexType.get(), 0)),
+            stream_d.dispatch_workgroup_id(IntegerAttr.get(IndexType.get(), 1)),
```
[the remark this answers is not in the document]
raikonenfnu: done!

raikonenfnu:
@harsh-nod I think I addressed most of your comments. 

WRT "the hack", I think it should be OK, but Martin used to set everything except for output as input bindings, which I am not sure is correct, I am double checking the info, but if it is placeholder => typically used for allocations, then I reckon it may just be a transient/temporary buffer if it is not purely used as input or output.

raikonenfnu:
Actually upon further inspection of our codebase, seems like we are only using it as an ordering for Dispatch Entry Point ([link](https://github.com/iree-org/iree-turbine/blob/3b8310aeecb718d191838af121073f649b1f6c53/shark_turbine/kernel/compiler/dispatch_codegen.py#L118-L138)) and for test function call ([link](https://github.com/iree-org/iree-turbine/blob/3b8310aeecb718d191838af121073f649b1f6c53/shark_turbine/kernel/compiler/host_codegen.py#L35-L39)).

I guess these other types None and Temporary was made by Stella a while back, and the original intention was probably to map to buffer types in IREE. I think maybe good to discuss tomorrow on what a good interface going forward is.

shark_turbine/kernel/wave/codegen.py
harsh-nod: thanks

shark_turbine/kernel/compiler/kernel_codegen.py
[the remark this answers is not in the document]
harsh-nod: thanks!

shark_turbine/kernel/compiler/kernel_codegen.py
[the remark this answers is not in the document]
harsh-nod: Thanks!

shark_turbine/kernel/compiler/kernel_codegen.py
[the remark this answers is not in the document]
harsh-nod: thanks!

shark_turbine/kernel/wave/codegen.py
harsh-nod: Sounds good thanks

shark_turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -150,12 +159,17 @@ def kernel_buffer_temporary_bindings(self) -> list[BindingDesc]:
...
-        placeholder_nodes: list[fx.Node] = []
-        for node in graph.nodes:
-            if node.op != "placeholder":
-                continue
-            placeholder_nodes.append(node)
 
+        # Extract all placeholder nodes.
+        def is_placeholder(node: fx.Node):
+            custom = get_custom(node)
+            if isinstance(custom, Placeholder):
```
harsh-nod, line 166: how about ``` return isinstance(custom, Placeholder) ```?

shark_turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -191,19 +205,31 @@ def add_grid(self, grid: Type[Grid]):
             )
 
     def determine_input_output_buffers(self, graph: fx.Graph):
-        placeholder_nodes: list[fx.Node] = []
-        for node in graph.nodes:
-            if node.op != "placeholder":
-                continue
-            placeholder_nodes.append(node)
+        # Extract all placeholder nodes.
+        def is_placeholder(node: fx.Node):
```
harsh-nod, line 209: is this function duplicated?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -293,24 +256,11 @@ def handle_add(emitter: WaveEmitter, node: fx.Node):
...
-            lhs = lhs.ir_value
-            rhs = arith_d.truncf(
-                VectorType.get(rhs_type.shape, F16Type.get()), rhs.ir_value
-            )
-    else:
-        lhs = lhs.ir_value
-        rhs = rhs.ir_value
+    lhs = lhs.ir_value
+    rhs = rhs.ir_value
     result = arith_d.addf(lhs, rhs)
```
harsh-nod, line 265: How about support for addi? I think those are the only 2 possibilities?

harsh-nod:
Thanks! It looking pretty good, just a few minor tweaks and should be good to go. Regarding the buffer types, I believe the order is input memrefs first and output memref last. I dont think we need the None and Temporary types. We could ignore them for now and revisit later.

harsh-nod (changes requested):

shark_turbine/kernel/compiler/kernel_codegen.py
raikonenfnu: Ah, that makes sense, silly me :)

shark_turbine/kernel/compiler/kernel_codegen.py
raikonenfnu: Yes, sorry, I am still not sure on the decorum on how/where to put these util/helper function. At first I was thinking of using it similar to how we use lambda functions in C++. 

De-duplicated though, LMK if you think there is a more apt spot. :)

shark_turbine/kernel/wave/codegen.py
raikonenfnu: Added, good catch, thanks!

harsh-nod: thanks!

harsh-nod (approved):
Thanks for pushing this! Just give us a good foundation to build on.

raikonenfnu:
> Thanks for pushing this! Just give us a good foundation to build on.

Thanks for the guidance Harsh! It was a good experience to touch different parts of the stack.
Going to fruther write handling for write and mma next to enable gemm e2e running here.  :)

PR #62, opened 2024-08-01
Make CI lit verbose

Hardcode84:
Pass `-v` option to lit to enable verbose output so we can actually see what failed exactly.


harsh-nod (approved):
thanks!

PR #63, opened 2024-08-01
Move globals to separate file and prefix with $

harsh-nod:
This PR adds a dollar prefix to all private globals to avoid namespace collisions with user defined
variables.

### DIFF
--- a/lit_tests/kernel/wave/expansion.py
+++ b/lit_tests/kernel/wave/expansion.py
@@ -106,21 +106,21 @@ def test_read_write_equal_sizes():
         # CHECK-NEXT: placeholder(_name=a
         # CHECK-NEXT: placeholder(_name=c
         # CHECK-NEXT: read(memory=a
-        # CHECK-SAME: index={M: BLOCK_M*WG0, N: BLOCK_N*WG1}
+        # CHECK-SAME: index={M: $WG0*BLOCK_M, N: $WG1*BLOCK_N}
         # CHECK-NEXT: read(memory=a
-        # CHECK-SAME: index={M: BLOCK_M*WG0, N: BLOCK_N*WG1}
+        # CHECK-SAME: index={M: $WG0*BLOCK_M, N: $WG1*BLOCK_N}
         # CHECK-NEXT: read(memory=a
-        # CHECK-SAME: index={M: BLOCK_M*WG0, N: BLOCK_N*WG1}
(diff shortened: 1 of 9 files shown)


shark_turbine/kernel/lang/global_symbols.py
```diff
@@ -0,0 +1,21 @@
...
+WORKGROUP_0 = index_symbol("$WG0")
+WORKGROUP_1 = index_symbol("$WG1")
+WORKGROUP_2 = index_symbol("$WG2")
+
+THREAD_0 = index_symbol("$T0")
+THREAD_1 = index_symbol("$T1")
+THREAD_2 = index_symbol("$T2")
+
+# MMA symbols
+MMA_LHS = index_symbol("$MMA_LHS")
```
raikonenfnu, line 19: Is there some intuition to why do we change these but not `BLOCK_M`?

harsh-nod: Good question. BLOCK_M is a user-defined parameter. The user is free to choose any symbol for the tile size. But the symbols in this file are used in the codegen pipeline. (Think of them as reserved keywords in a language) and that's why we distinguish them.

Hardcode84 (approved):

lit_tests/kernel/wave/expansion.py
```diff
@@ -106,21 +106,21 @@ def test_read_write_equal_sizes():
         # CHECK-NEXT: placeholder(_name=a
         # CHECK-NEXT: placeholder(_name=c
         # CHECK-NEXT: read(memory=a
-        # CHECK-SAME: index={M: BLOCK_M*WG0, N: BLOCK_N*WG1}
+        # CHECK-SAME: index={M: $WG0*BLOCK_M, N: $WG1*BLOCK_N}
```
Hardcode84, line 109: Interesting, I assume sympy sorts vars in commutative exprs based on name.

shark_turbine/kernel/lang/global_symbols.py
raikonenfnu: Makes sense, thanks!

lit_tests/kernel/wave/expansion.py
harsh-nod: yeah I was surprised by that as well.

PR #65, opened 2024-08-02
Add wave contraints

harsh-nod:
This PR adds wave constraints which specify how to distribute a tensor dimension among waves. The assumption is that the dimension has already been distributed among workgroups and is further being distributed among waves.

### DIFF
--- a/shark_turbine/kernel/wave/constraints.py
+++ b/shark_turbine/kernel/wave/constraints.py
@@ -176,3 +176,41 @@ class TilingConstraint(Constraint):
                 "Index is being computed without setting induction variable"
             )
         return IndexSequence(self.induction_var * self.tile_size, 1)
+
+
+@dataclass
+class WaveConstraint(Constraint):
+    """
+    A constraint of the form `tkw.WaveConstraint(K, WAVE_K)` specifies
+    that we want distribute the K dimension among multiple waves which
+    each wave operating on a tile size of WAVE_K. The assumption is
+    that the K dimension has already been distributed among workgroups.
+    If the K dimension has been distributed among workgroups with a
+    tile size of BLOCK_K, then the number of waves along the K dimension
+    is given by BLOCK_K // WAVE_K.
+
+    This constraint adds an index constraint to the K-th dimension of a
+    a tensor of the form WAVE_K * wave_id. The index of the wave
+    is determined by the following mapping:
+    workgroup id 0 -> wave/thread id x
+    workgroup id 1 -> wave/thread id y
+    workgroup id 2 -> wave/thread id z
+    (If the tensor dimension has been distributed along workgroup dimension
+    {0, 1, 2}, then the corresponding thread id is {x, y, z}).
+
+    Because we represent the number of threads per block as
+    [wave_id_0 * threads_per_wave, wave_id_1, wave_id_2], special care is
+    required when computing wave_id_0. Specifically,
+    wave_id_0 = floor(thread_id_0 / threads_per_wave)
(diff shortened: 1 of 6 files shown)


shark_turbine/kernel/wave/constraints.py
```diff
@@ -176,3 +176,41 @@ def apply(self) -> IndexSequence:
                 "Index is being computed without setting induction variable"
             )
         return IndexSequence(self.induction_var * self.tile_size, 1)
+
+
+@dataclass
+class WaveConstraint(Constraint):
```
raikonenfnu, line 182: @harsh-nod is this the lowest level of constraint we expose to users? Since based on this we can determine indexing based of threadIdx.

If the above is true, is that why we called it TKW?

harsh-nod: Yes this is the lowest level of constraint we expose to users. The reason the lang is called tkw is because the kernels are authored at the wave level, but the reason why we need this constraint is because the author needs to specify how many waves will operate along each dimension (or more specifically what is the tile size that each wave is operating on, which implicitly translates to how many waves operate along each dimension).

shark_turbine/kernel/wave/wave.py
```diff
@@ -118,6 +137,25 @@ def is_reduction(node: fx.Node):
                 if tiling_constraint.dim == custom.axis:
                     tiling_constraint.induction_var = self.induction_vars[custom]
 
+    def initialize_wave_constraints(self, trace: CapturedTrace) -> None:
+        """
+        For each wave constraint, determines the appropriate wave id by looking
+        for workgroup constraints along the same dimension and using information
+        from the hardware constraints.
```
raikonenfnu, line 144: >    Because we represent the number of threads per block as
    [wave_id_0 * threads_per_wave, wave_id_1, wave_id_2], special care is
    required when computing wave_id_0. Specifically,
    wave_id_0 = floor(thread_id_0 / threads_per_wave)
    wave_id_1 = thread_id_1
    wave_id_2 = thread_id_2

Maybe should move this comment here

shark_turbine/kernel/wave/wave.py
```diff
@@ -118,6 +137,25 @@ def is_reduction(node: fx.Node):
...
+
+        """
+
+        hardware_constraint = self.hardware_constraints[0]
+        thread_ids = [THREAD_0, THREAD_1, THREAD_2]
+        for wave_constraint in self.wave_constraints:
+            for workgroup_constraint in self.workgroup_constraints:
+                if wave_constraint.dim == workgroup_constraint.dim:
+                    wave_constraint.wave_id = thread_ids[
+                        workgroup_constraint.workgroup_dim
```
raikonenfnu, line 154: Seems like we can infer the wave_id here, is there any reason to want to specify it from users? also wouldn't the user-specified wave_id actually be overriden if we do testExecute or invoke the `_trace_and_get_kernel_signature`?

shark_turbine/kernel/wave/wave.py
harsh-nod: Sure I can move it here.

shark_turbine/kernel/wave/wave.py
harsh-nod: We don't ever want the user to specify the wave id. I use it only for testing. It should always be inferred.

raikonenfnu: I se I see, that makes sense

raikonenfnu (approved):

PR #66, opened 2024-08-02
Prefix induction variables with $

harsh-nod:
This avoids an possible namespace collision
with user defined variables. Also does minor
refactors.


raikonenfnu (approved):
Nice cleanups thanks!

Hardcode84 (approved):

PR #39, opened 2024-06-28
Changes required for a successful build that passes tests 

saienduri:
This commit makes the necessary changes to get out a successful release that passes all tests.
Also added some scripts to help with release validation.
New build release CI to make sure no commits break iree-turbine release build.

PR #67, opened 2024-08-02
[dep] pin IREE versions

raikonenfnu:
We had experience of tests suddenly failing, we should make iree-compiler/iree-runtime pip package updates more delibrate.

### DIFF
--- a/iree-requirements.txt
+++ b/iree-requirements.txt
@@ -1,2 +1,2 @@
-iree-compiler
-iree-runtime
+iree-compiler==20240724.964
+iree-runtime==20240724.964


iree-requirements.txt
```diff
@@ -1,2 +1,2 @@
-iree-compiler
-iree-runtime
+iree-compiler==20240724.964
+iree-runtime==20240724.964
```
ScottTodd, line 2: These were pinned before https://github.com/iree-org/iree-turbine/pull/39, FYI

We might also want the `setup.py` changes from https://github.com/iree-org/iree-turbine/pull/21 🤔

raikonenfnu: yeah, weird that we unpin haha

> we might also want setup.py changes

btw, did you mean `load_requirement_pins("iree-requirements.txt")` ?

saienduri (approved):

PR #68, opened 2024-08-02
Add indexing during expansion

harsh-nod:
This PR updates the index during expansion. Since the operations are performed at the granularity of the instruction specific or user specified vector shape, the index needs to be updated based on this vector shape.


shark_turbine/kernel/wave/expansion.py
```diff
@@ -208,14 +214,40 @@ def expand_graph(
...
-            _expand_node(node, trace, expand_dims, dim_scaling, expansion_context)
+            _expand_node(
+                node, trace, expand_dims, dim_scaling, dim_tile_size, expansion_context
+            )
+
+    # Replace symbolic tile sizes with actual tile sizes.
+    def replace_index(node: fx.Node) -> bool:
+        if node.index is not None:
+            for dim in node.index:
+                node.index[dim].start = node.index[dim].start.subs(tile_size_mapping)
```
raikonenfnu, line 225: Any thoughts on setting the key-value pair for `tile_size_mapping` into indexing_context? Then we can just do all the substitution in one go during codegen. :)

shark_turbine/kernel/wave/expansion.py
```diff
@@ -208,14 +214,40 @@ def expand_graph(
...
+    restricted_dims: dict[IndexSymbol, int],
+    dim_tile_size: dict[IndexSymbol, int],
+    new_node: fx.Node,
+) -> None:
+    for dim, index in restricted_dims.items():
+        if new_node.fx_node.index is not None and dim in new_node.fx_node.index:
+            for tile_size in dim_tile_size.values():
+                new_node.fx_node.index[dim].start = new_node.fx_node.index[
+                    dim
+                ].start.subs({tile_size: 0})
```
raikonenfnu, line 241: I am not sure if I understand this correctly, are we subbing in the value of the dim_tile_size(i.e the int values inside vector shape) by 0?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -208,14 +214,40 @@ def expand_graph(
...
+    def replace_index(node: fx.Node) -> bool:
+        if node.index is not None:
+            for dim in node.index:
+                node.index[dim].start = node.index[dim].start.subs(tile_size_mapping)
+        return False
+
+    trace.walk(replace_index)
+
+
+def update_index(
```
raikonenfnu, line 231: No action for now (More of a wish list really), Any thoughts on somehow combining `update_index` and `compute_index` in a single location/function?

shark_turbine/kernel/wave/expansion.py
harsh-nod: Good question. So let me explain what is going on here. So during expansion, we expand a node or unroll it. So for example if we had a read with shape [M, N] and M and N each had a dim_scaling of 2, we would produce read_0_0, read_0_1, read_1_0 and read_1_1 (not necessarily in that order). Now for indexing we want to multiply the current dimension value with the tile size. So in theory, we could just get the value of the dims, so for example for read_1_0, M = 1, N = 0 and then multiply by tile sizes and we are done. This works but the problem is that we visit the same node multiple dims during the DFS. As a result, if we just did the naive approach, we would end up double counting (or more depending on the structure of the graph). In order to avoid this, we use a symbolic tile size and before setting the index we zero out the symbolic tile size to avoid over-counting and then apply the symbolic tile size times the value of M/N. Then as a final step we replace the symbolic value with the actual values. In general, I don't think we can know how many times a node will be traversed so this seemed like a foolproof approach to me. @martin-luecke might have some thoughts on this as well. But the reason why we dont want to add this into indexing_context is because this is temporary index only required during the expansion phase.

shark_turbine/kernel/wave/expansion.py
harsh-nod: The reason why they are separate is because I wanted to avoid redundant computation. We can put compute_index into update_index but that means that we would have to repeat the constraint based index computation for every node in the expansion. By separating them out, we pay a one time cost to compute the constraint based indices and then only compute the expansion based indices during expansion.

shark_turbine/kernel/wave/expansion.py
harsh-nod: See other comment, but yes, we are erasing all previous instances of tile size to avoid overcounting and then applying the index offset.

raikonenfnu: I see I see, but wouldn't the index still be symbolic at this stage, since we only call `replace_index` after we finish expanding all the nodes and there won't be any more calls to `update_index`?

shark_turbine/kernel/wave/expansion.py
raikonenfnu: I see, yeah that makes sense, would've been nice for reading ergonomics though haha. 

Although, I think if we made the expand nodes less pinned on dimension-to-dimension, by trying to expand all the dims in one go and just move from node to node, then we may be able to combine both index compute and still save compute here. Although, I think it's a bit hard to convey what I am saying over message, we can chat more about this on Monday with some example IRs I was thinking about. :)

shark_turbine/kernel/wave/expansion.py
```diff
@@ -208,14 +214,40 @@ def expand_graph(
...
+            )
+
+    # Replace symbolic tile sizes with actual tile sizes.
+    def replace_index(node: fx.Node) -> bool:
+        if node.index is not None:
+            for dim in node.index:
+                node.index[dim].start = node.index[dim].start.subs(tile_size_mapping)
+        return False
+
+    trace.walk(replace_index)
```
martin-luecke, line 228: Why is this part of the expansion? Is there no case where we might want these symbolic later?
Does this have to be done by walking all nodes or could this be done when we visit the nodes individually?

shark_turbine/kernel/wave/expansion.py
martin-luecke: This feels like a setter for the index attribute of CustomOp. Could we add this as a property there?
e.g.
```Python
    @property
    def index(self):
        return self._index

    @index.setter
    def index(self, new_index):
        // [...]
        self._index = ...
```

shark_turbine/kernel/wave/expansion.py
```diff
@@ -383,14 +427,17 @@ def get_dim_scaling(
                     "Tile size must be divisible by wave count and vector size"
                 )
             dim_scaling[constraint.dim] = tile_size // wave_count // vector_size
-    return dim_scaling
+            dim_tile_size[constraint.dim] = index_symbol(f"$TILE_SIZE_{constraint.dim}")
+            tile_size_mapping[dim_tile_size[constraint.dim]] = vector_size
+    return (dim_scaling, dim_tile_size, tile_size_mapping)
```
martin-luecke, line 432: I am not sure I understand the name of `tile_size_mapping`. This maps a tile_size symbol to the corresponding vector size. Does it make sense to name it `dim_vector_size` then? Maybe there is another good name that I am not seeing right now

shark_turbine/kernel/wave/expansion.py
```diff
@@ -178,17 +179,22 @@ def compute_index(node: fx.Node) -> bool:
...
     Create a graph that represents the expanded version of the wave function.
     The expansion is done in the dimensions specified by the constraints.
     """
+    tile_size_mapping = {}
     if isinstance(constraints_or_scaling, dict):
         dim_scaling = constraints_or_scaling
+        dim_tile_size = tile_sizes
     else:
         mma_index = get_mma_dimensional_mapping(trace)
         set_node_indices(trace, constraints_or_scaling, mma_index)
```
martin-luecke, line 194: `set_node_indices` is not part of this PR, but I'll add the thought anyway:
It does an independent walk of all nodes and performs an action on them. This is required before the expansion but I am not sure it is really a part of it. Feels like a preprocessing step.

I think it will serve visibility in the future if we factor out these individual bits and call them from `wave.py`. Otherwise we will end up with huge files (this one is already almost 500 loc) that perform more than what their name indicates.

shark_turbine/kernel/wave/expansion.py
```diff
@@ -12,6 +12,7 @@
 from .._support.indexing import IndexingContext
 from ...support.logging import get_logger
 from .._support.tracing import CapturedTrace
+from .._support.indexing import index_symbol
 from .indexing import IndexSequence
```
martin-luecke, line 16: Sorry, regarding a previous PR:
Why is `IndexSequence` not part of `.._support.indexing`. 
Also its import in `wave_ops.py` is missing.

martin-luecke (commented):
Thank you for the PR!
I mainly have some questions on how to layer stuff, what should be part of expansion and what could be factored out into more individual steps so others can aproach the system easier with better overview

shark_turbine/kernel/wave/expansion.py
harsh-nod: Yes that's right. The index will still be symbolic. So it would be something like `symbolic_tile_size * 2` and then we would zero it out and then set it to say `symbolic_tile_size * 3`. (If we didn't zero it out, we would end up with `symbolic_tile_size * 2 + symbolic_tile_size * 3`.

shark_turbine/kernel/wave/expansion.py
harsh-nod: Sure I can add it there.

shark_turbine/kernel/wave/expansion.py
harsh-nod: Good question. Hadn't look much into that code but seems like it could fit there. Will move and add import into wave_ops.py

shark_turbine/kernel/wave/expansion.py
harsh-nod: Yes that's right. I was trying to avoid recomputing the mma_dimensional_mapping, but that is also a preprocessing step. Will move them both out of here into a separate file.

shark_turbine/kernel/wave/expansion.py
raikonenfnu: should we not be iterating over keys instead of values then? 

as in `for tile_size in dim_tile_size.keys():` instead of `for tile_size in dim_tile_size.values():`?

shark_turbine/kernel/wave/expansion.py
raikonenfnu: Style NIT on that, maybe we should name the index setter function as `set_index` for better clarity.

shark_turbine/kernel/wave/expansion.py
raikonenfnu: I had the same question since I was expecting this to be part of the `codegen.py`, but here is Harsh's response:

>  But the reason why we dont want to add this into indexing_context is because this is temporary index only required during the expansion phase.

harsh-nod: So this brings up a good question. My naive attempt at indexing was to just update the index but when I did this I saw overcounting happening which implied that nodes are being visited multiple times. Do you know why this happens? So my understanding is that if M = [0, 1, 2] and N = [0, 1], then we should visit the node only 3 x 2 = 6 times, but we are visiting it more than that. Why is this the case? 

The reason why this is part of the expansion is because I dont think it makes sense to keep the tile sizes symbolic after expansion is materialized. Not sure what the benefit of that would be. I dont think we can do this by visiting the nodes individually unless we figure out the first question - why do we multiple a node more than its dim scaling requires it to?

shark_turbine/kernel/wave/expansion.py
harsh-nod: sure let me rename this.

shark_turbine/kernel/wave/expansion.py
```diff
@@ -178,17 +179,22 @@ def compute_index(node: fx.Node) -> bool:
 def expand_graph(
     trace: CapturedTrace,
     constraints_or_scaling: Sequence[Constraint] | dict[IndexSymbol, int],
+    tile_sizes: dict[IndexSymbol, int] = None,
```
Hardcode84, line 182: `tile_sizes` is never checked for `None` before assigning to `dim_tile_size`, and  `dim_tile_size` == `None` will break things.

shark_turbine/kernel/wave/expansion.py
harsh-nod: @raikonenfnu - I believe that's how you define the setter (you can't do it as set_index https://realpython.com/python-getter-setter/)

shark_turbine/kernel/wave/expansion.py
Hardcode84: Adding an assert in corresponding branch will be a good idea. Also, type should be `Optional[dict[IndexSymbol, int]]` in this case

shark_turbine/kernel/wave/expansion.py
harsh-nod: So dim_tile_size is a dictionary from dimension to tile size and looks something like {M: TILE_M, N: TILE_N}.
Our original expression looks something like: index = 2 * TILE_N, 3 * TILE_M
so we want to iterate over the values (TILE_M, TILE_N) and set those to zero to get index = 0 and then update the index.
Does that make sense?

shark_turbine/kernel/wave/expansion.py
raikonenfnu: > My naive attempt at indexing was to just update the index but when I did this I saw overcounting happening which implied that nodes are being visited multiple times

Oof, that looks like we have a bug in the check for whether or not we have expanded the node in that dimension? 


IIRC we have a piece of code that queries the map who have `original_node+Indexing -> expanded_node` to check that we have expanded before. as seen on https://github.com/iree-org/iree-turbine/blob/f0271e8e55fb1d6e51236a621b17a1ff97c1ccf5/shark_turbine/kernel/wave/expansion.py#L221-L225

shark_turbine/kernel/wave/expansion.py
harsh-nod: Thanks, will add that check here!

shark_turbine/kernel/wave/expansion.py
harsh-nod: removed this part of the code.

shark_turbine/kernel/wave/expansion.py
harsh-nod: this is no longer required. I have reworked the code.

shark_turbine/kernel/wave/expansion.py
harsh-nod: I have removed this from the code.

shark_turbine/kernel/wave/expansion.py
harsh-nod: I have modified this so that no the index calculation is part of the expansion. I think that this makes the index calculation easier to understand (with absolute indices being calculated instead of relative indices). With this, I don't think these need to be moved to a separate file.

lit_tests/kernel/wave/codegen.py
```diff
@@ -130,19 +130,19 @@ def test(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
...
         # CHECK: %[[C16_1:.+]] = arith.constant 16 : index
         # CHECK: %[[WG1_OFF:.+]] = arith.muli %[[WG_1]], %[[C16_1]]
         # CHECK: %[[IDX_Y:.+]] = arith.addi %[[WG1_OFF]], %[[T1_OFF]]
-        # CHECK: %[[OFF:.+]] = arith.constant dense<[0, 16, 32, 48, 64, 80, 96, 112, 128, 144, 160, 176, 192, 208, 224, 240]> : vector<16xindex>
+        # CHECK: %[[C16_0:.+]] = arith.constant 16 : index
+        # CHECK: %[[WG0_OFF:.+]] = arith.muli %[[WG_0]], %[[C16_0]]
+        # CHECK: %[[C4:.+]] = arith.constant 4 : index
+        # CHECK: %[[T0_OFF:.+]] = arith.divsi %[[T0]], %[[C4]]
+        # CHECK: %[[IDX_X:.+]] = arith.addi %[[T0_OFF]], %[[WG0_OFF]]
+        # CHECK: %[[OFF:.+]] = arith.constant dense<[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]> : vector<16xindex>
```
Hardcode84, line 142: I don't think this is correct as consecutive indices means we are just doing contiguous load like `vector.load` and they should be strided for transpose. But let's merge it, so I can pick up it.

Hardcode84 (approved):

lit_tests/kernel/wave/codegen.py
harsh-nod: thanks!

PR #70, opened 2024-08-04
Add strided and dynamic offset to memref type

harsh-nod:
When the static shape information is available,
we use a strided memref type with a dynamic offset for integration with iree.


raikonenfnu (approved):
The code looks great! Although I am not completely sure how are we planning to use the  the stride and offset info on our stack. :)

harsh-nod:
> The code looks great! Although I am not completely sure how are we planning to use the the stride and offset info on our stack. :)

So when I was first trying to integrate the tk kernels, I ran into issues with the iree integration. What happens is that one of the stream passes tries to optimize the bindings (use the same buffer for multiple input arguments) and so in order to enable that we need to have a dynamic offset, (not a fixed offset of 0).

raikonenfnu:
Makes sense, thanks for the explanation :)

PR #71, opened 2024-08-06
Adapt to torch 2.4 API

saienduri:
This commit makes the necessary changes to adapt to the new torch 2.4 API. The main change here is regarding the exporting of dynamic shape constraints: https://github.com/pytorch/pytorch/commit/342e7929b804ec56121e82e92d6a199b549c38b1.
Also changed the test for strict export for something that still holds.

### DIFF
--- a/shark_turbine/aot/exporter.py
+++ b/shark_turbine/aot/exporter.py
@@ -252,6 +252,12 @@ def export(
             )
         )
 
+    if not strict_export:
+        warnings.warn(
+            "When strict_export == False, Python interpreter is used and code will execute exactly as it would in eager mode. "
+            "Therefore, graph optimizations and safety are not garaunteed to be at the same level as TorchDynamo."
+        )
+
     TransformedModule: Any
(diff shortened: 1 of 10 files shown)


shark_turbine/aot/exporter.py
```diff
@@ -252,6 +252,12 @@ def export(
             )
         )
 
+    if not strict_export:
+        warnings.warn(
+            "When strict_export == False, Python interpreter is used and code will execute exactly as it would in eager mode. "
+            "Therefore, graph optimizations and safety are not garaunteed to be at the same level as TorchDynamo."
```
stellaraccident, line 258: Please add "This is an experimental feature in PyTorch that the IREE Turbine project is still evaluating. Please report issues or experiences."

stellaraccident (approved):
Thank you.

shark_turbine/aot/exporter.py
saienduri: Done

PR #74, opened 2024-08-08
[TK] Lower `read` with mapping to vector.gather

Hardcode84:


shark_turbine/kernel/wave/codegen.py
```diff
@@ -227,22 +248,63 @@ def handle_read(emitter: WaveEmitter, node: fx.Node):
...
+        subs = [(sym, expr.start) for sym, expr in zip(iters.keys(), index.values())]
+
+        input_index = {
+            key: m.subs(subs) for key, m in zip(mem_index.keys(), input_mapping)
+        }
+
+        strides = strides_from_symbolic_shape(IndexingContext.current(), mem_index)
+        offsets = []
+        subs = [(sym, 0) for sym in iters.keys()]
+        for i in range(elements_per_thread):
```
raikonenfnu, line 282: Can you explain a bit what is going on here? I was reading through vector.gather's doc, this seems to be the `index_vec` but even there it's not really explained very clearly what the `index_vec` does. 

Also any reason why we are using vector.gather as opposed to say `masked_load`?

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -572,6 +572,8 @@ class Read(CustomOp):
 
     @property
     def indexing_dims(self) -> list[IndexSymbol]:
+        if self.mapping is not None:
+            return list(self.mapping.output_shape)
```
raikonenfnu, line 576: I don't quite understand why the mapping.output_shape will be different from `self.memory.type.symbolic_shape`? Doesn't the former also represent "output shape"?

shark_turbine/kernel/lang/wave_types.py
```diff
@@ -209,3 +202,24 @@ def map_output_indices(
         self, symbols: Optional[tuple[IndexSymbol, ...]] = None
     ) -> tuple[IndexExpr, ...]:
         return self._map_indices(self.output_mapping, symbols)
+
+    def is_output_identity(self) -> bool:
+        return _is_identity_mapping(self.iters.keys(), self.output_mapping)
+
+
+def _subs_expr(expr: Any, subs: Iterable[tuple[IndexExpr, IndexExpr]]) -> Any:
```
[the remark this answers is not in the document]
Hardcode84: Moved it back

shark_turbine/kernel/wave/codegen.py
Hardcode84: `vector.gather` is modeled against `llvm.masked.gather` intrinsic https://llvm.org/docs/LangRef.html#llvm-masked-gather-intrinsics

`masked_load` will read a contiguous vector from `start_indices` position, but `gather` will read each individual element using `(start_indices + offsets_vec[i])` formula and construct a vector from them.

`vector.gather` semantics is actually underspecified when you have multidimensional memref as input, as `offsets_vec[i]` elements are always interpreted as 1D but I think using linearized indices should work and it translates straightforwardly to llvm intrinsic.

raikonenfnu: OK that makes sense, so we are using the "offset_vecs[i]" to read/take potentially non-contiguous data that will be contiguous **after** the read. Thanks for the explanation. :)

shark_turbine/kernel/ops/wave_ops.py
raikonenfnu: So you're saying for the case of iGEMM, the `self.memory.type.symbolic_shape` will be (sym.N, sym.C, sym.H, sym.W) as opposed to something like (sym.M, sym.K)?

raikonenfnu: Actually, yeah re-read your reply, that makes sense, thanks :)

shark_turbine/kernel/wave/codegen.py
```diff
@@ -227,22 +248,63 @@ def handle_read(emitter: WaveEmitter, node: fx.Node):
...
+        result = vector_d.load(vector_type, kb_src, start_indices)
+    else:
+        assert (
+            mapping.is_output_identity()
+        ), "non-dentity output mapping is not supported yet"
+        mem_index = memory.index
+        input_mapping = mapping.map_input_indices(mem_index.keys())
+
+        iters = mapping.iters
+        subs = [(sym, expr.start) for sym, expr in zip(iters.keys(), index.values())]
```
raikonenfnu, line 273: We can do this in a follow-on PR, but I think we should move these to the IndexMapping class. IIRC @harsh-nod wanted the codegen.py to be as straightforward as we can be.

Hardcode84: Yeah, makes sense, I need to think more how to better organize this.

raikonenfnu (approved):
Thanks for the work Ivan, PR Overall LGTM! 

Would be nice to have a follow-on PR for moving out some of the logic out of codegen.py into IndexMapping. 

Another follow-on PR for simplifying offset computation would also be nice, seems a little complex for what it is doing now, but I can't think of anything better for now either, so going to approve this PR for now to not block progress. :)

PR #80, opened 2024-08-14
[TK] Fix gather indices

Hardcode84:
Use original `memory` symbolic shape to determine dimensions access order.

### DIFF
--- a/shark_turbine/kernel/wave/codegen.py
+++ b/shark_turbine/kernel/wave/codegen.py
@@ -27,7 +27,7 @@ from shark_turbine.aot.support.ir_utils import _is_float_type, _is_integer_like_
 
 # TK infrastructure imports.
 from shark_turbine.kernel.lang.global_symbols import *
-from ..ops.wave_ops import write, register, mma, read, reduction
+from ..ops.wave_ops import write, register, mma, read, reduction, get_custom
 from ..compiler.base import CodegenError, ValidationError, NDEBUG
 from ..compiler.kernel_codegen import BoundKernelSignature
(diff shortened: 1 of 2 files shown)


Hardcode84:
@raikonenfnu I still need to refactor indices calculation into proper place, but I will do it in separate PR from bugfix

shark_turbine/kernel/wave/codegen.py
```diff
@@ -264,15 +264,27 @@ def handle_read(emitter: WaveEmitter, node: fx.Node):
...
+        # Apply input order to input indices, e.g. if original input_mapping was
+        # {M: iter(0), N: iter(1)} and source mem has shape (N, M), result will
+        # be (iter(1), iter(0))
+        input_mapping = mapping.map_input_indices(mem_shape)
 
         iters = mapping.iters
+
+        # As we only support identity output mapping for now, we can directly
+        # substitute iterators with corresponding expanded index.
         subs = [(sym, expr.start) for sym, expr in zip(iters.keys(), index.values())]
```
harsh-nod, line 281: Is there an assumption here about the ordering of the dimensions in index and iters? Say for example index was {K: ... , M: ... , N: ...} vs {M: .., N: ..., K: ...}, would that create a problem?

Hardcode84: I'm using mapping output shape for indexing if mapping is present, and for now, assert `mapping.is_output_identity()` is enforces output mapping is always `(iter(0), iter(1), ...)`

harsh-nod (approved):
thanks!

PR #81, opened 2024-08-14
Add automatic promotion and hoisting of placeholders

harsh-nod:
This PR adds promotion to shared memory and hoisting of shared memory allocs to wave. If a placeholder is marked with its address space as SHARED_MEMORY, the compiler will introduce the appropriate promotion.

### DIFF
--- a/shark_turbine/kernel/wave/promotion.py
+++ b/shark_turbine/kernel/wave/promotion.py
@@ -1,4 +1,6 @@
+from .._support.tracing import CapturedTrace
 from ...support.logging import get_logger
+from .._support.indexing import IndexingContext
 from ..ops.wave_ops import *
 from ..lang.global_symbols import *
 
@@ -38,3 +40,15 @@ def promote_node(node: Read | Write, address_space: IndexSymbol):
         )
         allocate_node.add_to_graph(node.graph)
         apply_promotion_pattern(node, allocate_node)
+
+
+def promote_placeholders(graph: CapturedTrace):
+    for node in graph.get_root_graph().nodes:
+        custom = get_custom(node)
+        if isinstance(custom, Read) or isinstance(custom, Write):
+            if not custom.type:
+                continue
+            idxc = IndexingContext.current()
+            address_space = custom.type.address_space.subs(idxc.subs)
+            if address_space == SHARED_ADDRESS_SPACE:
+                promote_node(custom, address_space)
--- a/shark_turbine/kernel/wave/wave.py
+++ b/shark_turbine/kernel/wave/wave.py
@@ -15,6 +15,8 @@ from .constraints import (
 )
 from .codegen import WaveEmitter
 from .expansion import expand_graph
+from .promotion import promote_placeholders
+from .hoisting import hoist_allocs
 from ..lang import Grid, IndexMapping
 from ..lang.global_symbols import *
 from ..ops import wave_ops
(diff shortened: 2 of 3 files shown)


shark_turbine/kernel/wave/wave.py
```diff
@@ -170,6 +172,10 @@ def _trace_and_get_kernel_signature(
         idxc = IndexingContext.current()
         idxc.finalize()
 
+        # Promote the placeholders to the appropriate address space.
+        promote_placeholders(graph)
+        hoist_allocs(graph)
```
raikonenfnu, line 177: Can we add a simple codegen test for this?

shark_turbine/kernel/wave/promotion.py
```diff
@@ -38,3 +40,15 @@ def promote_node(node: Read | Write, address_space: IndexSymbol):
...
+
+
+def promote_placeholders(graph: CapturedTrace):
+    for node in graph.get_root_graph().nodes:
+        custom = get_custom(node)
+        if isinstance(custom, Read) or isinstance(custom, Write):
+            if not custom.type:
+                continue
+            idxc = IndexingContext.current()
+            address_space = custom.type.address_space.subs(idxc.subs)
```
raikonenfnu, line 52: For my own education, is type for custom.type.address_space symbolic expression? if so, why can't we just have it be directly `index_symbol("$SHARED_ADDRESS_SPACE")`?

shark_turbine/kernel/wave/promotion.py
```diff
@@ -38,3 +40,15 @@ def promote_node(node: Read | Write, address_space: IndexSymbol):
         )
         allocate_node.add_to_graph(node.graph)
         apply_promotion_pattern(node, allocate_node)
+
+
+def promote_placeholders(graph: CapturedTrace):
+    for node in graph.get_root_graph().nodes:
+        custom = get_custom(node)
+        if isinstance(custom, Read) or isinstance(custom, Write):
```
raikonenfnu, line 48: Seems like `apply_promotion_pattern` that is used by `promote_node` only handles `Read`, are we just adding `Write` since we plan on adding support for it later?

shark_turbine/kernel/wave/promotion.py
raikonenfnu: Looking at the test I see that we are setting it as another symbol `ADDRESS_SPACE_K` where K is argument number. Is this to make the kernels more modification friendly S.T we can use same kernel but change up TestLaunchContext to modify address space?

shark_turbine/kernel/wave/wave.py
harsh-nod: That was my stretch goal for this week :) Will add it after I upload a minimize global load PR.

shark_turbine/kernel/wave/promotion.py
harsh-nod: Yes, in fact this is what Jakub was originally working on. Promoting the writes so that we have nice global write patterns. With Ivan's mapping construct, I think we could implement this.

shark_turbine/kernel/wave/wave.py
raikonenfnu: haha sweet, will leave this for a todo/follow on PR then.

shark_turbine/kernel/wave/promotion.py
raikonenfnu: Makes sense, thanks!

raikonenfnu (approved):

shark_turbine/kernel/wave/promotion.py
raikonenfnu: sounds good!

PR #86, opened 2024-08-16
Add lowerings for mma, register and allocate

harsh-nod:
This PR adds a mma unit test which lowers to
vector.loads/stores and amdgpu.mfmas. Also supports shared memory promotion.


shark_turbine/kernel/wave/wave.py
```diff
@@ -197,10 +202,13 @@ def _trace_and_get_kernel_signature(
         exe = dispatch_codegen.StreamExecutable(mb, name=entrypoint_name)
         dispatch_entrypoint = exe.define_entrypoint(entrypoint_name, kernel_sig, grid)
 
-        emitter = WaveEmitter(dispatch_entrypoint, graph)
+        emitter = WaveEmitter(dispatch_entrypoint, graph, self.constraints)
         emitter.emit(graph.get_root_graph())
         emitter.finish()
 
+        if "canonicalize" in kwargs and kwargs["canonicalize"]:
```
Hardcode84, line 209: `kwargs.get("canonicalize", False)`

lit_tests/kernel/wave/codegen.py
```diff
@@ -247,6 +251,100 @@ def test(
...
+            LOAD_ELEMS_PER_THREAD: 4,
+            STORE_ELEMS_PER_THREAD: 4,
+            ADDRESS_SPACE: SHARED_ADDRESS_SPACE,
+            ADDRESS_SPACE_0: GLOBAL_ADDRESS_SPACE,
+        }
+    ):
+        a = torch.randn(64, 32, dtype=torch.float16)
+        b = torch.randn(128, 32, dtype=torch.float16)
+        c = torch.zeros(64, 128, dtype=torch.float32)
+        print(mma(a, b, c, canonicalize=True).module_op)
```
Hardcode84, line 297: IMO, options like `canonicalize=True` should be part of the launch context and not specific invocation.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -412,9 +459,48 @@ def handle_write(emitter: WaveEmitter, node: fx.Node):
...
+    return result
+
+
 @handle_op(mma)
 def handle_mma(emitter: WaveEmitter, node: fx.Node):
-    raise NotImplementedError("MMA: Currently only stub implementation")
+    try:
+        lhs, rhs, acc = node.args
+        acc = cast_vector(emitter, acc)
+        values = [lhs, rhs]
```
Hardcode84, line 488: `values = [cast_vector(emitter, val) for val in [lhs, rhs]]`

shark_turbine/kernel/wave/codegen.py
```diff
@@ -412,9 +459,48 @@ def handle_write(emitter: WaveEmitter, node: fx.Node):
...
+        for i in range(len(values)):
+            values[i] = cast_vector(emitter, values[i])
+    except ValueError as e:
+        raise ValidationError("Malformed arguments") from e
+
+    vector_type = VectorType(acc.type)
+    result = None
+    for constraint in emitter.constraints:
+        if isinstance(constraint, HardwareConstraint):
+            m, n, k = constraint.mma_matrix_shapes
```
Hardcode84, line 498: mfma can have different sizes depending datatype if I understand correctly. The fact we encoding a single possible matrix shape per kernel may be problematic.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -412,9 +459,48 @@ def handle_write(emitter: WaveEmitter, node: fx.Node):
...
+            result = emit_mfma(m, n, k, vector_type, acc, values)
+            break
+
+    if result:
```
Hardcode84, line 502: Is there any legitimate case when this `if` is not triggered? I think missing `HardwareConstraint` should be a compilation error.

shark_turbine/kernel/wave/constraints.py
```diff
@@ -213,3 +213,19 @@ def apply(self) -> IndexSequence:
...
+
+
+def get_workgroup_distributed_shape(
+    shape: list[IndexExpr], constraints: list[WorkgroupConstraint]
+) -> tuple[IndexExpr]:
+    """
+    Given a shape and workgroup constraints, returns the shape
+    of the tensor after it has been distributed along workgroup dimensions.
+    """
+    distributed_shape = [s for s in shape]
```
Hardcode84, line 225: `distributed_shape = list(shape)`

lit_tests/kernel/wave/codegen.py
harsh-nod: I agree, will change this.

shark_turbine/kernel/wave/codegen.py
harsh-nod: Yes that is correct. Even for the same datatype there are multiple sizes. In the gemm case and for FA as well, we have not found any benefit to having more than one matrix shape per kernel. I think this is a safe assumption for now, but we can revisit it if we run into some issues.

shark_turbine/kernel/wave/codegen.py
harsh-nod: Good point, yes missing hardware constraint should be a compilation error. I will rework this.

shark_turbine/kernel/wave/codegen.py
harsh-nod: thanks

shark_turbine/kernel/wave/constraints.py
harsh-nod: thanks

shark_turbine/kernel/wave/wave.py
```diff
@@ -197,10 +202,13 @@ def _trace_and_get_kernel_signature(
         exe = dispatch_codegen.StreamExecutable(mb, name=entrypoint_name)
         dispatch_entrypoint = exe.define_entrypoint(entrypoint_name, kernel_sig, grid)
 
-        emitter = WaveEmitter(dispatch_entrypoint, graph)
+        emitter = WaveEmitter(dispatch_entrypoint, graph, self.constraints)
         emitter.emit(graph.get_root_graph())
         emitter.finish()
 
+        if kwargs.get("canonicalize", False):
```
raikonenfnu, line 209: When do we not want to canonicalize?

shark_turbine/kernel/wave/utils.py
```diff
@@ -0,0 +1,35 @@
...
+def canonicalize_module(module: Operation):
+    with module.context, Location.unknown():
+        transform_module = builtin_d.Module.create()
+        transform_module_op = module.operation
+        transform_module_op.attributes["transform.with_named_sequence"] = UnitAttr.get()
+        with InsertionPoint(transform_module.body):
+            named_sequence = transform_d.NamedSequenceOp(
+                "__transform_main", [any_op_t()], []
+            )
+            with InsertionPoint(named_sequence.body):
```
raikonenfnu, line 25: Just curious, is this the best way to setup canonicalization? by setting up a TD-like structure to do the canonicalization? I'd assume we'd have better API support from upstream to call canonicalization and or other standard patterns on.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -412,9 +459,52 @@ def handle_write(emitter: WaveEmitter, node: fx.Node):
...
+    hardware_constraints = [
+        constraint
+        for constraint in emitter.constraints
+        if isinstance(constraint, HardwareConstraint)
+    ]
+    if not hardware_constraints:
+        raise CodegenError("No hardware constraints found.")
+
+    result = None
+    for constraint in hardware_constraints:
```
Hardcode84, line 503: We probably should validate `len(hardware_constraints) == 1` and get rid of this loop.

Hardcode84 (approved):
I will need to look at indexing more later, but let's merge it for now so we can make progress.

shark_turbine/kernel/wave/wave.py
harsh-nod: You don't want to canonicalize when you will lose all your IR on canonicalization (because they have no uses). This happens on some of the other tests (you can try it out by setting canonicalize=True on some of the other tests).

shark_turbine/kernel/wave/utils.py
harsh-nod: Good question and I was thinking about this myself. I think we can avoid TD by using other patterns, but there could be some advantages to using TD for now (as it has support for other patterns like LICM etc.). If we discover that we don't need that, we can always rewrite this without using TD.

shark_turbine/kernel/wave/codegen.py
harsh-nod: Sure, will add this to the follow on PR.

PR #88, opened 2024-08-19
Add lowerings for reduction

harsh-nod:
This PR adds lowerings for reduction operators.
With these, we now have a complete GEMM lowering
that can be tested for correctness and benchmarked for performance.

### DIFF
--- a/shark_turbine/kernel/wave/codegen.py
+++ b/shark_turbine/kernel/wave/codegen.py
@@ -3,6 +3,7 @@ import sympy
 from typing import Any, Callable, ClassVar, Optional, List, Type
 from dataclasses import dataclass
 import torch.fx as fx
+import torch.utils._pytree as pytree
 
 from ..compiler.ir import (
     Attribute,
@@ -25,14 +26,25 @@ from ..compiler.ir import (
     gpu_d,
     memref_d,
     stream_d,
+    scf_d,
     vector_d,
 )
 from shark_turbine.aot.support.ir_utils import _is_float_type, _is_integer_like_type
 
 # TK infrastructure imports.
 from shark_turbine.kernel.lang.global_symbols import *
-from ..ops.wave_ops import write, register, mma, read, reduction, get_custom, allocate
-from ..lang.wave_types import IndexMapping
+from ..ops.wave_ops import (
+    write,
+    register,
+    mma,
+    read,
+    reduction,
+    get_custom,
+    get_result,
+    allocate,
+    CustomOp,
+)
+from ..lang.wave_types import IndexMapping, IndexSymbol
 from ..compiler.base import CodegenError, ValidationError, NDEBUG
 from ..compiler.kernel_codegen import BoundKernelSignature
 from .._support.tracing import CapturedTrace
@@ -44,7 +56,7 @@ from ..compiler.vector_codegen import (
     cast_py_value,
     cast_vector,
 )
-from .constraints import Constraint, HardwareConstraint, MMAType
+from .constraints import Constraint, HardwareConstraint, MMAType, TilingConstraint
 
 # Indexing imports.
 from .._support.indexing import IndexingContext, IndexExpr, IndexSequence
@@ -75,6 +87,7 @@ class WaveEmitter:
             gpu_d.thread_id(gpu_d.Dimension.y),
             gpu_d.thread_id(gpu_d.Dimension.z),
         ]
+        self.induction_vars: dict[IndexSymbol, Value] = {}
 
     def emit(self, graph: Optional[fx.Graph] = None):
         with self.ip, Location.unknown():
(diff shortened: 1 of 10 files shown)


shark_turbine/kernel/wave/promotion.py
```diff
@@ -18,9 +18,11 @@ def apply_promotion_pattern(custom_node: Read | Write, allocate_node: Allocate):
             ).add_to_graph(custom_node.graph)
             custom_node.replace_all_uses_with(promoted_read)
             with custom_node.graph.inserting_before(promoted_read):
-                Write(
+                promoted_write = Write(
                     custom_node.fx_node, allocate_node.fx_node, elements_per_thread
                 ).add_to_graph(custom_node.graph)
+                custom_read = get_custom(promoted_read)
+                custom_read.write_dependency = promoted_write
```
raikonenfnu, line 25: I assume this write_dependency will be used for barriers for shared memory read/writes and scheduling, but not sure if I see it currently being used aside from just setting the attribute?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -555,4 +585,74 @@ def handle_sub(emitter: WaveEmitter, node: fx.Node):
...
+    )
+    emitter.induction_vars[axis] = forOp.induction_variable
+    with InsertionPoint(forOp.body):
+        # Add mapping for iter args.
+        subgraph: fx.Graph = emitter.trace.get_subgraph(subgraph)
+        iter_args: list[fx.Node] = get_custom(node).iter_args(subgraph)
+        for i, v in enumerate(forOp.inner_iter_args):
+            emitter.bind_node_proxy(iter_args[i], IRProxyValue(v))
+        captured_vars: list[fx.Node] = get_custom(node).captured_vars(subgraph)
+        for root_v, subgraph_v in zip(implicit_capture, captured_vars):
```
raikonenfnu, line 631: QQ: Are both implicit_capture and captured_vars represented as `Placeholder`? Seems like the implicit_captures are the iter_args, would that mean generally implicit_captures == tensors in shared_memory, captured_vars == registers?

raikonenfnu: Sorry, misread the code, I reckon implicit_capture ~= captured_vars. 

But the original question still stand is 
iter_args == tensor for shared memory, and captured_vars == tensor in registers?

raikonenfnu (commented):
Hey Harsh, thanks for the PR, overall this looks great! Seems like the PR teaches wave compiler how to propagate register analysis on tkw.reduction + lower/generate `scf.for` and it's region + introduce write_dependencies for barriering later on. 

I was wondering if we can move the introduce write_dependencies onto a separate PR that also introduces/lowers to barriers? S.T the commits would be easier to understand.

shark_turbine/kernel/wave/promotion.py
harsh-nod: Yes, that's the PR I am going to submit next. This is required to get the expansion working since the expansion works by following the use-def chains and we have a break in the chain when we write to and then read from shared memory. Think of this as the write to shared memory returns a SSA value that is consumed by the read from shared memory.

shark_turbine/kernel/wave/codegen.py
harsh-nod: For the gemm graph, the implicit captures are [a, b] and the captured_vars are [c_reg]. So the implicit captures are global tensors while the captured vars are the iter args of the for loop.

harsh-nod:
> Hey Harsh, thanks for the PR, overall this looks great! Seems like the PR teaches wave compiler how to propagate register analysis on tkw.reduction + lower/generate `scf.for` and it's region + introduce write_dependencies for barriering later on.
> 
> I was wondering if we can move the introduce write_dependencies onto a separate PR that also introduces/lowers to barriers? S.T the commits would be easier to understand.

Thanks! Unfortunately, if I remove the write_dependency the expansion fails and we can't generate valid IR. Think of this PR as the minimum required to get GEMM IR minus the barriers. Not sure how to separate it out if expansion needs it?

shark_turbine/kernel/wave/promotion.py
raikonenfnu: Ahhh, right right, we have that expansion thingy that backward slice DFS from final-writes. 

OK, so if we do not do this, the write to shared memory wouldn't be expanded?

harsh-nod: yes exactly. since it starts at the write node, it would propagate to read from shared memory but not higher.

raikonenfnu: OK, makes sense, thanks for the explanation. :)

raikonenfnu (approved):
Overall LGTM

PR #89, opened 2024-08-19
[TK] Basic end-to-end infra

Hardcode84:
* Add basic infra to run kernels e2e (`TestLaunchContext` `run=True` argument)
* Add simple copy kernel as test (currently fails, 16 first values are correct and the rest are garbage, will debug it separately)
* To run e2e tests `WAVE_RUN_E2E_TESTS=1 pytest tests/kernel/wave/wave_e2e_test.py`

### DIFF
--- a/shark_turbine/kernel/wave/wave.py
+++ b/shark_turbine/kernel/wave/wave.py
@@ -2,7 +2,7 @@ from typing import Any, Callable, Optional
 import torch.fx as fx
 import inspect
 
-from ..compiler import builder, dispatch_codegen, kernel_codegen
+from ..compiler import builder, dispatch_codegen, kernel_codegen, host_codegen
 from ..compiler.ir import Context, Operation
 from .codegen import WaveEmitter
 from .constraints import (
@@ -31,6 +31,8 @@ from .._support.tracing import (
     KernelRegionGraph,
     Launchable,
 )
+from iree.compiler import compile_str
+import iree.runtime as rt
 
 __all__ = ["wave", "wave_trace_only"]
 
@@ -209,11 +211,43 @@ class LaunchableWave(Launchable):
         if kwargs.get("canonicalize", False):
             canonicalize_module(mb.module_op)
 
-        return mb, graph
+        return mb, graph, exe, kernel_sig, entrypoint_name
 
     def test_execute(self, args, kwargs):
         # For now only tracing
-        mb, graph = self._trace_and_get_kernel_signature(args, kwargs)
+        print("test_execute")
+        mb, graph, exe, kernel_sig, entrypoint_name = self._trace_and_get_kernel_signature(args, kwargs)
+        host_codegen.isolated_test_call(mb, exe, kernel_sig, entrypoint_name)
+        asm = mb.module_op.get_asm()
+        print(asm)
+        target = "rocm"
+        flags = [
+            "--mlir-print-op-on-diagnostic=false",
+            "--iree-llvmcpu-target-cpu-features=host",
+            "--iree-llvmcpu-target-triple=x86_64-linux-gnu",
+            f"--iree-hal-target-backends={target}",
+            "--iree-rocm-target-chip=gfx942",
+            # "--mlir-print-ir-after-all",
+        ]
+        res = compile_str(asm, target_backends=[target], extra_args=flags)
+
+        # TODO: cache
+        device=None
+        config = rt.Config(device)
+        mod = rt.VmModule.from_flatbuffer(config.vm_instance, res)
+
+        vm_modules = [
+            mod,
+            rt.create_hal_module(config.vm_instance, config.device),
+        ]
+        ctx = rt.SystemContext(
+            vm_modules=vm_modules,
+            config=config,
+        )
+
+        CompModule = ctx.modules.compiled_resnet18_model
+        device_arr = [rt.asdevicearray(config.device, inp) for inp in args]
+        output = CompModule["test"](*device_arr)
         return mb
 
     def aot_execute(self, args, kwargs):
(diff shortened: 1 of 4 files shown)


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@require_e2e
+def test_copy():
+    M = tkl.sym.M
+    N = tkl.sym.N
+    K = tkl.sym.K
+    BLOCK_M = tkl.sym.BLOCK_M
+    BLOCK_N = tkl.sym.BLOCK_N
+    BLOCK_K = tkl.sym.BLOCK_K
```
harsh-nod, line 21: Is BLOCK_K and K being used here?

shark_turbine/kernel/wave/wave.py
```diff
@@ -209,11 +218,66 @@ def _trace_and_get_kernel_signature(
...
+            for arg, b in zip(args, kernel_sig.kernel_buffer_bindings):
+                usage = b.kernel_buffer_type.usage
+                if usage == kernel_codegen.KernelBufferUsage.INPUT:
+                    kernel_inputs.append(arg)
+
+                if usage == kernel_codegen.KernelBufferUsage.OUTPUT:
+                    kernel_outputs.append(arg)
+
+            # TODO: Unhardcode
+            target = "gfx942"
```
harsh-nod, line 249: What do you think of a "configuration" class -  which contains things like device(rocm/hip) target arch(gfx942), whether to print mlir after all, etc.? We can instantiate that dict and pass it in as an argument.

shark_turbine/kernel/wave/wave.py
```diff
@@ -209,11 +218,66 @@ def _trace_and_get_kernel_signature(
...
+                "--iree-opt-strip-assertions=true",
+                "--iree-vm-target-truncate-unsupported-floats",
+            ]
+            if kwargs.get("print_ir_after_all", False):
+                flags.append("--mlir-print-ir-after-all")
+
+            res = compile_str(asm, target_backends=["rocm"], extra_args=flags)
+
+            config = rt.Config("hip")
+            mod = rt.VmModule.copy_buffer(config.vm_instance, res)
```
harsh-nod, line 263: Can we use a wrap_buffer here instead of copy buffer?

shark_turbine/kernel/wave/wave.py
```diff
@@ -209,11 +218,66 @@ def _trace_and_get_kernel_signature(
         if kwargs.get("canonicalize", False):
             canonicalize_module(mb.module_op)
 
-        return mb, graph
+        return mb, graph, exe, kernel_sig, entrypoint_name
 
     def test_execute(self, args, kwargs):
         # For now only tracing
```
harsh-nod, line 224: This comment seems outdated now :)

shark_turbine/kernel/wave/wave.py
```diff
@@ -209,11 +218,66 @@ def _trace_and_get_kernel_signature(
...
+                rt.create_hal_module(config.vm_instance, config.device),
+            ]
+            ctx = rt.SystemContext(
+                vm_modules=vm_modules,
+                config=config,
+            )
+
+            CompModule = ctx.modules.module
+            device_arr = [rt.asdevicearray(config.device, inp) for inp in kernel_inputs]
+            output = _to_tuple(CompModule["isolated_benchmark"](*device_arr))
```
harsh-nod, line 276: I believe you can use the vm_context (from the system context) to invoke the function, without accessing the inner modules,
like 
```
config.vm_context().invoke(self._vm_function, arg_list, ret_list)
```
See here: https://github.com/iree-org/iree-turbine/blob/main/shark_turbine/dynamo/executor.py 
and use lookup_function to find the kernel.

shark_turbine/kernel/wave/wave.py
```diff
@@ -209,11 +218,66 @@ def _trace_and_get_kernel_signature(
...
+            for orig, arr in zip(kernel_outputs, output):
+                # Convert to torch tensor without actually importing torch.
+                orig[:] = type(orig)(arr.to_host())
```
harsh-nod, line 279: Are we doing an additional copy here of the outputs? So one from device -> host and then another when casting?

harsh-nod (changes requested):
Thanks for the PR and the e2e test. Mostly my questions revolve around ireert API usage and whether we are doing any redundant copies (which would be expensive).

shark_turbine/kernel/wave/wave.py
Hardcode84: `wrap_buffer` fails with `ValueError: VmModule.from_aligned_memory received an unaligned buffer. Got 0x0x55871ff26120, expected alignment 64`

tests/kernel/wave/wave_e2e_test.py
Hardcode84: removed

shark_turbine/kernel/wave/wave.py
Hardcode84: Good idea, Done.

shark_turbine/kernel/wave/wave.py
Hardcode84: `DeviceArray` doesn't seem to support copying device data into user provided buffer, added a comment. It's ok for testing, I think. For production we probably want a different approach, when user passes device buffers directly, so we will be able to avoid copying altogether.

shark_turbine/kernel/wave/wave.py
harsh-nod: thanks!

shark_turbine/kernel/wave/wave.py
harsh-nod: good to know, thanks for checking.

shark_turbine/kernel/wave/wave.py
harsh-nod: thanks!

harsh-nod:
looks good! thanks!

harsh-nod (approved):
lgtm!

PR #90, opened 2024-08-20
Add shared memory barriers for RAW dependencies

harsh-nod:
This PR introduces a shared memory barrier op that can be used to to introduce barriers between writes and reads to shared memory. We use the write_dependency field of read nodes as well as the node indices to determine whether to insert a barrier. Also, we try to re-use barriers to minimize the total number of barriers.

### DIFF
--- a/lit_tests/kernel/wave/barriers.py
+++ b/lit_tests/kernel/wave/barriers.py
@@ -1,0 +1,272 @@
+# RUN: python %s | FileCheck %s
+
+import logging
+from typing import Callable
+import unittest
+import shark_turbine.kernel as tk
+import shark_turbine.kernel.lang as tkl
+import shark_turbine.kernel.wave as tkw
+from shark_turbine.kernel.wave.promotion import promote_node, promote_placeholders
+from shark_turbine.kernel.wave.barriers import add_shared_memory_barriers
+from shark_turbine.kernel.wave.hoisting import hoist_allocs
+from shark_turbine.kernel.wave.expansion import expand_graph
+from shark_turbine.kernel.lang.global_symbols import *
+from shark_turbine.kernel._support.tracing import CapturedTrace
+from shark_turbine.kernel._support.indexing import IndexingContext
+from shark_turbine.kernel.ops.wave_ops import *
+
+
+def run(func: Callable[[], None]) -> Callable[[], None]:
+    """Run a function as part of the test suite."""
+    if __name__ == "__main__":
+        func()
+        # Print a separator between tests
+        print("-----")
+    return func
+
+
+def get_read_nodes(graph: fx.Graph) -> list[CustomOp]:
+    custom_nodes: list[CustomOp] = [get_custom(node) for node in graph.nodes]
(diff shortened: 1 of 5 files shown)


shark_turbine/kernel/wave/barriers.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    barriers: list[SharedMemoryBarrier] = []
+    for subgraph in trace.region_graph.subgraphs.values():
+        for node in subgraph.nodes:
+            custom = get_custom(node)
+            if isinstance(custom, Read) and custom.write_dependency:
+                if custom.index != custom.write_dependency.index:
+                    if any(
+                        barrier.is_barrier_between(
+                            custom.write_dependency, custom.fx_node
```
Hardcode84, line 22: Isn't `custom.fx_node` is same as just `node` here?

lit_tests/kernel/wave/barriers.py
```diff
@@ -0,0 +1,272 @@
...
+from shark_turbine.kernel.wave.barriers import add_shared_memory_barriers
+from shark_turbine.kernel.wave.hoisting import hoist_allocs
+from shark_turbine.kernel.wave.expansion import expand_graph
+from shark_turbine.kernel.lang.global_symbols import *
+from shark_turbine.kernel._support.tracing import CapturedTrace
+from shark_turbine.kernel._support.indexing import IndexingContext
+from shark_turbine.kernel.ops.wave_ops import *
+
+
+def run(func: Callable[[], None]) -> Callable[[], None]:
```
Hardcode84, line 19: We should probably move this to some common place, as it copypasted in every test.

shark_turbine/kernel/wave/barriers.py
```diff
@@ -0,0 +1,30 @@
...
+                        )
+                        for barrier in barriers
```
Hardcode84, line 24: I'm not entirely sure, but algo complexity here doesn't  looking good (quadratic?). It's probably okay for now, but let's keep it in mind.

Hardcode84 (approved):
LGTM

raikonenfnu (approved):
Thanks for the PR, Overall LGTM :)

shark_turbine/kernel/wave/barriers.py
harsh-nod: yes, will fix.

shark_turbine/kernel/wave/barriers.py
harsh-nod: Yes O(N) to walk through the graph forward , O(N) to walk the graph backward. And we need to repeat this for B barriers, so total complexity should be O(BN). What could make this acceptable is a) if we have few barriers or b) if we have a lot of barriers, but are able to find one early since any(...) short circuits. You could make it better if you could identify likely barrier candidates, but yeah let's leave for now and evaluate in the future.

lit_tests/kernel/wave/barriers.py
harsh-nod: Good point, let me do that cleanup in a subsequent PR.

PR #91, opened 2024-08-20
Refactor run and print_trace methods

harsh-nod:
This PR moves the common run and print_trace
utilities to the utils file so they do not
need to duplicated in every test.


Hardcode84 (approved):
Nice

PR #92, opened 2024-08-21
[TK] Fix e2e copy test

Hardcode84:
With some trial and error found some constraints configuration which gives correct result, but I have no idea what I'm doing. @harsh-nod please take a look.

### DIFF
--- a/tests/kernel/wave/wave_e2e_test.py
+++ b/tests/kernel/wave/wave_e2e_test.py
@@ -24,7 +24,7 @@ def test_copy():
         tkw.HardwareConstraint(
             threads_per_wave=64,
             waves_per_block=(1, 1, 1),
-            # vector_shapes={M: 1, N: ELEMS_PER_THREAD}, # TODO: Symbol desn't work here, but should
+            # vector_shapes={M: 1, N: ELEMS_PER_THREAD}, # TODO: Symbol doesn't work here, but should
             vector_shapes={M: 1, N: 4},
         )
     ]
@@ -43,7 +43,7 @@ def test_copy():
 
     config = {"backend": "rocm", "device": "hip", "target": "gfx942"}
 
-    shape = (16, 16)
+    shape = (256, 128)
     a = torch.randn(shape, dtype=torch.float16)
     b = torch.zeros(shape, dtype=torch.float16)
     with tk.gen.TestLaunchContext(
--- a/tests/kernel/wave/wave_e2e_test.py
+++ b/tests/kernel/wave/wave_e2e_test.py
@@ -17,37 +17,42 @@ def test_copy():
     N = tkl.sym.N
     BLOCK_M = tkl.sym.BLOCK_M
     BLOCK_N = tkl.sym.BLOCK_N
+    ELEMS_PER_THREAD = tkl.sym.ELEMS_PER_THREAD
     ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
 
     constraints: list[tkw.Constraint] = [
         tkw.HardwareConstraint(
-            threads_per_wave=64, waves_per_block=(1, 1, 1), vector_shapes={M: 16, N: 16}
+            threads_per_wave=64,
+            waves_per_block=(1, 1, 1),
+            vector_shapes={M: 1, N: ELEMS_PER_THREAD},
         )
     ]
     constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
     constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
-    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
-    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+    constraints += [tkw.WaveConstraint(M, 64)]
+    constraints += [tkw.WaveConstraint(N, 1)]
 
     @tkw.wave(constraints)
     def test(
         a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
         b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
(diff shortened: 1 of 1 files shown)


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -17,37 +17,43 @@ def test_copy():
...
+            waves_per_block=(1, 1, 1),
+            # vector_shapes={M: 1, N: ELEMS_PER_THREAD}, # TODO: Symbol doesn't work here, but should
+            vector_shapes={M: 1, N: 4},
         )
     ]
     constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
     constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
-    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
-    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+    constraints += [tkw.WaveConstraint(M, 64)]
```
raikonenfnu, line 33: I don't think the WAVE_TILE_SIZE_M > WG_TILE_SIZE_M makes sense. Also to my understanding, `WG_TILE_SIZE_M  / WAVE_TILE_SIZE_M == waves_per_block[0]` .

raikonenfnu: You can revert it, it should still pass. :)

Hardcode84: Yeah, I remember, something wasn't working with `tkw.WaveConstraint(M, 1)`, but current version works.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -17,37 +17,43 @@ def test_copy():
...
+            waves_per_block=(1, 1, 1),
+            # vector_shapes={M: 1, N: ELEMS_PER_THREAD}, # TODO: Symbol doesn't work here, but should
+            vector_shapes={M: 1, N: 4},
         )
     ]
     constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
     constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
-    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
-    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+    constraints += [tkw.WaveConstraint(M, 1)]
```
raikonenfnu, line 33: Can we make it 
```py
constraints += [tkw.WaveConstraint(M, BLOCK_M)]
 constraints += [tkw.WaveConstraint(N, BLOCK_N)]
 ```
 Since that'd actually mean/imply waves_per_block=(1, 1, 1),
 
 waves_per_block IIUC is `waves_per_block ~= (BLOCK_M/WAVE_M, BLOCK_N/WAVE_N, BLOCK_K,WAVE_K)`. (order might change depends on how we specify the shape -> thread_dim).

Comment (Hardcode84) [reply tests/kernel/wave/wave_e2e_test.py:33]: Actually, I've checked, this value doesn't seem to affect result at all, it works even if I just set there arbitrary int value.

Comment (raikonenfnu) [reply tests/kernel/wave/wave_e2e_test.py]: Yeah, I noticed that too, I think there's some missing pieces in expansion and constraints that allow us to do this, we may be doing redundant copies too. I am going to take a look some more at these pieces to see if we can make it do the right thing. :)

**Review by raikonenfnu — APPROVED** [review]

**Comment by raikonenfnu** [comment]
Now just need to figure out how to make the runners with MI300 pick this up haha

File: tests/kernel/wave/wave_e2e_test.py [hunk 3]
```diff
@@ -17,11 +17,15 @@ def test_copy():
...
     BLOCK_N = tkl.sym.BLOCK_N
+    ELEMS_PER_THREAD = tkl.sym.ELEMS_PER_THREAD
     ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
 
     constraints: list[tkw.Constraint] = [
         tkw.HardwareConstraint(
-            threads_per_wave=64, waves_per_block=(1, 1, 1), vector_shapes={M: 16, N: 16}
+            threads_per_wave=64,
+            waves_per_block=(1, 1, 1),
+            # vector_shapes={M: 1, N: ELEMS_PER_THREAD}, # TODO: Symbol doesn't work here, but should
```
harsh-nod, line 27: Yeah I think its missing a subs from the IndexingContext.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -17,11 +17,15 @@ def test_copy():
...
+            vector_shapes={M: 1, N: 4},
```
harsh-nod, line 28: I think 1, 2 and 8 would also work here because they map to the instructions. For 16, not sure what it was doing. We should probably add either an assert or break down larger loads to smaller ones.

harsh-nod (approved):
awesome! the first fully functional e2e kernel! :)

tests/kernel/wave/wave_e2e_test.py
Hardcode84: At least currently, this value must be exactly aligned with `elements_per_thread`, otherwise things will break. And in general `elements_per_thread` currently in some weird position, as never taken in account in expansion or any indexing calculations. Also, big vector sizes works too (I tried 16) as, apparently, underlying llvm pipeline is able to decompose them.

raikonenfnu:
@Hardcode84 Thanks again for the awesome fix and findings. I incorporated some of your findings and further refined this e2e test onto my workgroup distribution/support work + add supports for symbolic vector_shape here https://github.com/iree-org/iree-turbine/pull/95. 

Feel free to land this(#92) PR if you like and I can refine/rebase on #95. :)

PR #93, opened 2024-08-21
[TK] Some basic APIs for benchmarking

Hardcode84:
Example config
```python
    config = {
        "backend": "rocm",
        "device": "hip",
        "target": "gfx942",
        "benchmark": True,
        "benchmark_batch_size": 100,
        "benchmark_repetitions": 3,
        "benchmark_results_file": "out.json",
    }
```
Example output:
```
[
    [
        "BM_isolated_benchmark/process_time/real_time",
        "0.036 ms",
        "0.080 ms",
        "19313",
        "items_per_second=27.6075k/s"
    ],
    [
        "BM_isolated_benchmark/process_time/real_time",
        "0.036 ms",
        "0.081 ms",
        "19313",
        "items_per_second=27.4817k/s"
    ],
    [
        "BM_isolated_benchmark/process_time/real_time",
        "0.036 ms",
        "0.081 ms",
        "19313",
        "items_per_second=27.5867k/s"
    ],
    [
        "BM_isolated_benchmark/process_time/real_time_mean",
        "0.036 ms",
        "0.081 ms",
        "3",
        "items_per_second=27.5586k/s"
    ],
    [
        "BM_isolated_benchmark/process_time/real_time_median",
        "0.036 ms",
        "0.081 ms",
        "3",
```

### DIFF
--- a/shark_turbine/kernel/wave/wave.py
+++ b/shark_turbine/kernel/wave/wave.py
@@ -25,6 +25,7 @@ from ..ops.wave_ops import Reduction, CustomOp, get_custom
 from .register_analysis import determine_register_shape
 from .._support.indexing import IndexingContext, IndexExpr
 import shark_turbine.kernel.lang as tkl
+from ...support.logging import get_logger
 from .._support.tracing import (
     CapturedTrace,
     CompiledContext,
@@ -41,6 +42,9 @@ import numpy
 bench.DTYPE_TO_ABI_TYPE[numpy.dtype(numpy.float16)] = "f16"
 
 __all__ = ["wave", "wave_trace_only"]
+
+
+logger = get_logger("turbine.wave")
 
 
 def wave(constraints: Optional[list[Constraint]] = None):
@@ -78,6 +82,21 @@ def _invoke(vm_context, device, entry_function, inputs, outputs):
 
         # Convert to torch tensor without actually importing torch.
         ret[:] = type(ret)(host_array)
+
+
+def _write_file(name, mode, data):
+    with open(name, mode) as file:
+        file.write(data)
+
+
+def _print_bench_result(result, filename):
+    import json
+
+    res = str(json.dumps(result, sort_keys=True, indent=4))
+    if filename is not None:
+        _write_file(filename, "w", res)
+    else:
+        logger.info(res)
 
 
 class LaunchableWave(Launchable):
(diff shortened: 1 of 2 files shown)


shark_turbine/kernel/wave/wave.py
```diff
@@ -298,8 +319,29 @@ def test_execute(self, args, kwargs):
...
+
+                inputs = [inp.numpy() for inp in kernel_inputs]
+                benchmark_results = bench.benchmark_module(
+                    mod,
+                    entry_function=func_name,
+                    device=device,
+                    inputs=inputs,
+                    **benchmark_flags,
+                )
+                print("Benchmark results:")
```
harsh-nod, line 339: Can we also dump this to a file? Just thinking ahead, when we do tuning or even just a sweep, we would like to know the configuration that produced the benchmark number. Also should we wrap the print with a logger?

shark_turbine/kernel/wave/wave.py
```diff
@@ -298,8 +319,29 @@ def test_execute(self, args, kwargs):
...
+                for result in benchmark_results:
+                    print(result)
+            else:
+                func = mod.lookup_function(func_name)
```
harsh-nod, line 343: The other thing (also based on experience) is that often during tuning, we want to first make sure that the candidate produces correct results and only if it produces correct results do we want to benchmark it and sometimes we might want to just benchmark the function. So I would reorganize this function so that if run or bench are enabled, we do the creation of the vm context / etc. and then if only run is enabled we only run, if only bench is enabled we only bench and if both are enabled we run and bench. The idea is that we can do both run and bench through a single pass through the function which is not possible right now. What do you think?

harsh-nod (changes requested):

shark_turbine/kernel/wave/wave.py
Hardcode84: done

Hardcode84:
Note: iree `benchmark_module` sometimes fails when input data is too big (looks like it tries to pass entire data via cmdline), it probably should be fixed in iree itself (and missing fp16 support).

shark_turbine/kernel/wave/wave.py
harsh-nod: Thanks!

shark_turbine/kernel/wave/wave.py
harsh-nod: Thanks

harsh-nod (approved):
lgtm! thanks!

PR #94, opened 2024-08-21
Add optimization to minimize number of global loads

harsh-nod:
In order to minimize the number of global loads, we modify the expansion and allow more threads to load larger amounts of data. This PR adds this optimization pass.

Also, read nodes can now have multiple write dependencies and the expansion pass has been modified to handle this case. Other utilities have also been added such as delinearization and removing unused nodes.


harsh-nod:
Here is what the graph looks like before: 
before

harsh-nod:
And this is the graph after:

after

harsh-nod:
Here we have a 4x reduction in global loads [8 -> 2]. I will add the code to generate the plots to the test.

harsh-nod:
Also refactored it a bit, so hopefully more clear.

shark_turbine/kernel/wave/visualization.py
```diff
@@ -12,9 +12,9 @@ def visualize_graph(graph: fx.Graph, file_name: str):
         raise ImportError("pygraphviz not installed, cannot visualize graph")
     G = pgv.AGraph(directed=True)
     for node in graph.nodes:
-        G.add_node(node.name)
+        G.add_node(id(node), label=node.name)
```
Hardcode84, line 15: `id(node)` is essentially a pointer, it is not stable between runs, we can potentially get different files each time (even if they look the same).

shark_turbine/kernel/wave/utils.py
```diff
@@ -59,3 +64,43 @@ def print_trace(trace: CapturedTrace, custom_print: bool = True):
...
+    """
+    Delinearizes a 1D index into a multi-dimensional index
+    based on the shapes provided. The returned array contains
+    the multi-dimensional index.
+    """
+    nd_index = []
+    product = 1
+    for i, size in enumerate(reversed(shape)):
+        if i == 0:
+            nd_index.append(sympy.Mod(index, size))
```
Hardcode84, line 102: I think `sympy.Mod(index, size)` is complete equivalent to just `index % size`

shark_turbine/kernel/wave/utils.py
```diff
@@ -59,3 +64,43 @@ def print_trace(trace: CapturedTrace, custom_print: bool = True):
         if custom_print:
             for node in subgraph.nodes:
                 print(get_custom(node))
+
+
+def remove_unused_operators(trace: CapturedTrace):
```
Hardcode84, line 69: nit: Should we just call it `DCE` to follow common terminology?

shark_turbine/kernel/wave/minimize_global_loads.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    Identify sub-optimal global loads that can be removed. A given memory has
+    sub-optimal global loads if
+        num_global_loads > (M * N) / (T * L)
+    where the memory has shape [M, N], there are T threads and each thread can load L elements.
+    """
+    optimizable_loads: dict[fx.Node, tuple[int, Read]] = {}
+    for read_node in global_read_nodes:
+        custom = get_custom(read_node)
+        if custom.memory in optimizable_loads:
```
raikonenfnu, line 101: So we are collecting/indexing on the global memory we are loading from as opposed to tevery instance of readOps right?

shark_turbine/kernel/wave/utils.py
```diff
@@ -59,3 +64,43 @@ def print_trace(trace: CapturedTrace, custom_print: bool = True):
...
+    while removable_nodes := trace.walk(is_removable_operator):
+        for node in removable_nodes:
+            get_custom(node).graph.erase_node(node)
+
+
+def delinearize_index(index: IndexExpr, shape: list[int]) -> list[IndexExpr]:
+    """
+    Delinearizes a 1D index into a multi-dimensional index
+    based on the shapes provided. The returned array contains
+    the multi-dimensional index.
```
raikonenfnu, line 96: NIT: Really NIT/opinion based but would be nice to have an easy sample similar to how affine dialect's delinearize_index doc:

```
%indices:3 = affine.delinearize_index %linear_index into (%c16, %c224, %c224) : index, index, index

%indices:3 conceptually holds the following:

[linear_idx floordiv 50176,  (linear_idx mod 50176) floordiv 224, linear_idx mod 224]
```

shark_turbine/kernel/wave/minimize_global_loads.py
```diff
@@ -0,0 +1,237 @@
...
+            continue
+
+        total_number_of_elements = prod(materialized_shape)
+        expected_number_of_loads = total_number_of_elements // max_elements_per_load
+        actual_number_of_loads = len(
+            [x for x in global_read_nodes if get_custom(x).memory == custom.memory]
+        )
+        if expected_number_of_loads >= actual_number_of_loads:
+            continue
+        optimizable_loads[custom.memory] = (expected_number_of_loads, custom)
```
raikonenfnu, line 117: Should we also check that it's only users are write to shared memory? This will be helpful in the case where we are doing computation without trips to shared memory.

 For example, if we had a simple elementwise add with suboptimal global reads/tile size, we might cause compilation failure or numerical issues.

shark_turbine/kernel/wave/minimize_global_loads.py
```diff
@@ -0,0 +1,237 @@
...
+from math import prod
+import torch.fx as fx
+from collections import defaultdict
+
+
+def has_write_shared_user(node: Read) -> bool:
+    idxc = IndexingContext.current()
+    return any(
+        isinstance(user, Write)
+        and user.type.address_space.subs(idxc.subs) == SHARED_ADDRESS_SPACE
```
raikonenfnu, line 21: Should we also check that it's **only** users are write to shared memory? This will be helpful in not breaking the case where we also have computation that doesn't  shared memory but rely on the same expansion.

EDIT: I think should be OK actually, since we are not actively removing/replacing the original global reads, and have auto-remove of unused ops.

shark_turbine/kernel/wave/expansion.py
```diff
@@ -241,14 +243,23 @@ def expand_graph(
...
-    node: CustomOp,
+    node: CustomOp | list[CustomOp],
     trace: CapturedTrace,
     dim_query: dict[IndexSymbol, int],
     dim_scaling: dict[IndexSymbol, int],
     node_index_setter: Callable[[CustomOp, dict[IndexSymbol, int]], None],
     context: ExpandedNodeMap,
 ) -> CustomOp:
     """Expand a single node in specific dimensions and recursively proceed to its inputs."""
+    if isinstance(node, list):
```
raikonenfnu, line 254: Is this to support expand nodes of `for i, arg in node.node_args.items():`?

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -309,12 +309,18 @@ def name(self) -> str:
...
-        ]
+        custom_args = {}
+        for i, arg in enumerate(self.fx_node.args):
+            if isinstance(arg, fx.Node):
+                custom_args[i] = get_custom(arg)
+            if isinstance(arg, list) and all(isinstance(x, fx.Node) for x in arg):
+                custom_args[i] = [get_custom(x) for x in arg]
+        return custom_args
+
+    def node_arg_index(self, arg: CustomOp) -> Optional[CustomOp | list[CustomOp]]:
```
raikonenfnu, line 322: NIT: I think would be more inutiiative to call it `get_arg_index` or `get_operand_index` or something along that line.

raikonenfnu (commented):
Thanks for this PR Harsh, overall code looks super clen and intuitive! just some couple Qs and NITs.

shark_turbine/kernel/wave/expansion.py
```diff
@@ -241,14 +243,23 @@ def expand_graph(
...
 def _expand_node(
-    node: CustomOp,
+    node: CustomOp | list[CustomOp],
     trace: CapturedTrace,
     dim_query: dict[IndexSymbol, int],
     dim_scaling: dict[IndexSymbol, int],
     node_index_setter: Callable[[CustomOp, dict[IndexSymbol, int]], None],
     context: ExpandedNodeMap,
 ) -> CustomOp:
     """Expand a single node in specific dimensions and recursively proceed to its inputs."""
```
martin-luecke, line 253: Let's modify the docstring to indicate what happens when a list of nodes is passed in

martin-luecke (approved):
LGTM after addressing the existing comments

shark_turbine/kernel/wave/expansion.py
harsh-nod: This is needed because I changed the write_dependency field of read nodes to accept a list of nodes (the read node can have multiple write dependencies). So I had to modify expansion to handle this case.

shark_turbine/kernel/wave/visualization.py
Hardcode84: We can iterate over all nodes, number them starting from 0 and, build a map `{id(node) : int}` and use resulting int as stable id.

Feel free to do it in separate PR to get things going fast.

harsh-nod: Thanks, will do that.

PR #95, opened 2024-08-22
[TKW] Implement workgroup expansion and execution support.

raikonenfnu:
Implements expansion and distribution to workgroup, S.T they will take into account the number of threads in a workgroup/block. Additionally, we also add some support for symbolic vector_shape in HW constraints. Also incorporated some of @Hardcode84 's finding to fix the e2e copy test.

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -289,7 +289,7 @@ def test_mma():
         c = torch.zeros(64, 128, dtype=torch.float32)
         print(mma(a, b, c).module_op)
 
-        # CHECK: func.func @mma(%[[ARG0:.+]]: !stream.binding, %[[ARG1:.+]]: !stream.binding, %[[ARG2:.+]]: !stream.binding) {
+        # CHECK: func.func @mma(%[[ARG0:.+]]: !stream.binding, %[[ARG1:.+]]: !stream.binding, %[[ARG2:.+]]: !stream.binding)
         # CHECK-DAG: %[[C16:.+]] = arith.constant 16 : index
         # CHECK-DAG: %[[C4:.+]] = arith.constant 4 : index
         # CHECK-DAG: %[[C32:.+]] = arith.constant 32 : index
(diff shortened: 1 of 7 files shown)


raikonenfnu:
FYI This is what the sample IR/copy looks like with this change:
```mlir
#translation = #iree_codegen.translation_info<None workgroup_size = [64, 1, 1] subgroup_size = 64>
module attributes {transform.with_named_sequence} {
  stream.executable private @test {
    stream.executable.export public @test workgroups() -> (index, index, index) {
      %c1 = arith.constant 1 : index
      %c256 = arith.constant 256 : index
      stream.return %c1, %c256, %c1 : index, index, index
    }
    builtin.module {
      func.func @test(%arg0: !stream.binding, %arg1: !stream.binding) attributes {translation_info = #translation} {
        %c128 = arith.constant 128 : index
        %c2 = arith.constant 2 : index
        %c0 = arith.constant 0 : index
        %workgroup_id_0 = stream.dispatch.workgroup.id[0] : index
        %workgroup_id_1 = stream.dispatch.workgroup.id[1] : index
        %thread_id_x = gpu.thread_id  x
        %thread_id_y = gpu.thread_id  y
        %0 = stream.binding.subspan %arg0[%c0] : !stream.binding -> memref<256x128xf16, strided<[128, 1], offset: ?>>
        %1 = arith.addi %workgroup_id_1, %thread_id_y : index
        %2 = arith.muli %thread_id_x, %c2 : index
        %3 = arith.muli %workgroup_id_0, %c128 : index
        %4 = arith.addi %3, %2 : index
        %5 = vector.load %0[%1, %4] : memref<256x128xf16, strided<[128, 1], offset: ?>>, vector<2xf16>
        %6 = stream.binding.subspan %arg1[%c0] : !stream.binding -> memref<256x128xf16, strided<[128, 1], offset: ?>>
        vector.store %5, %6[%1, %4] : memref<256x128xf16, strided<[128, 1], offset: ?>>, vector<2xf16>
        return
      }
    }
  }
}
```

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -17,15 +17,18 @@ def test_copy():
...
         tkw.HardwareConstraint(
-            threads_per_wave=64, waves_per_block=(1, 1, 1), vector_shapes={M: 16, N: 16}
+            threads_per_wave=64,
+            waves_per_block=(1, 1, 1),
+            vector_shapes={M: 1, N: ELEMS_PER_THREAD},
         )
     ]
-    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
-    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
```
harsh-nod, line 30: Why are the workgroup ids being switched around?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -430,7 +434,9 @@ def get_dim_scaling(
                 raise ValueError(
                     "Tile size must be divisible by wave count and vector size"
                 )
-            dim_scaling[constraint.dim] = tile_size // wave_count // vector_size
+            dim_scaling[constraint.dim] = (
+                tile_size // wave_count // num_threads // vector_size
```
harsh-nod, line 438: I dont think the expansion needs to depend on the number of threads. Why do you think so?

shark_turbine/kernel/compiler/dispatch_codegen.py
```diff
@@ -163,6 +166,12 @@ def abi_type(binding: BindingDesc):
                 def_func_op = func_d.FuncOp(name, def_ftype)
                 def_func_block = def_func_op.add_entry_block()
                 def_func_args = list(def_func_block.arguments)
+                if workgroup_size is not None and subgroup_size is not None:
+                    def_func_op.attributes["translation_info"] = Attribute.parse(
+                        f"#iree_codegen.translation_info<None "
+                        f"workgroup_size=[{','.join(str(x) for x in workgroup_size)}]"
+                        f"subgroup_size={subgroup_size}>"
+                    )
```
harsh-nod, line 174: Nice, this is definitely needed.

harsh-nod (changes requested):

tests/kernel/wave/wave_e2e_test.py
raikonenfnu: Discussed and resolved offline. 

For documentation purposes, we are switching since by convention we set `thread_per_block` to `(wave_count[0] * subgroup_size, wave_count[1], wave_count[2])`. So setting fastest dim to 0 allow us to distribute the fastest dim to the largest number of threads.

shark_turbine/kernel/wave/expansion.py
raikonenfnu: Also resolved offline. Turns out we do not need this. This was from my misunderstanding that vector_shape == multi-dim threads_per_elem. Where in reality vector_shape is the size of vector owned by a single warp.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -50,9 +49,9 @@ def test(
         {
             M: shape[0],
             N: shape[1],
-            BLOCK_M: 16,
-            BLOCK_N: 16,
-            ELEMS_PER_THREAD: 4,
+            BLOCK_M: 1,
+            BLOCK_N: 128,
+            ELEMS_PER_THREAD: 2,
```
Hardcode84, line 54: As discussed offline, it's equivalent to `min(N / 64, 8)`, right? Will it work if we write it this way?

Or, alternatively, we can try to define `ELEMS_PER_THREAD = `min(N / 64, 8)`` at the beginning.

raikonenfnu: This will work, we do not have to load 8 elements. We can choose how many elements we want to load. But right now since the N dim is small (128) and we have 64 threads at least(size of warp), we can at most use 2 element per thread.

tests/kernel/wave/wave_e2e_test.py
harsh-nod: Thanks, that makes sense. From the programmer's perspective this is not obvious, so I am in favor of not specifying the workgroup dim in the workgroup constraints and have the compiler assign the index to the innermost dimension (we can keep this as a power user argument).

lit_tests/kernel/wave/codegen.py
```diff
@@ -305,37 +305,19 @@ def mma(
...
         # CHECK: %[[ALLOC:.+]] = memref.alloc() : memref<32x16xf16, #gpu.address_space<workgroup>>
-        # CHECK: %[[R5:.+]] = arith.muli %[[WG0]], %[[C32]] : index
-        # CHECK: %[[R6:.+]] = arith.divsi %[[TX]], %[[C4]] : index
-        # CHECK: %[[R7:.+]] = arith.addi %[[R6]], %[[R5]] : index
-        # CHECK: vector.store %4, %[[ALLOC]][%[[R7]], %[[C0]]] : memref<32x16xf16, #gpu.address_space<workgroup>>, vector<4xf16>
-        # CHECK: %[[R8:.+]] = arith.muli %[[WG0]], %[[C32]] : index
-        # CHECK: %[[R9:.+]] = arith.divsi %[[TX]], %[[C4]] : index
-        # CHECK: %[[R10:.+]] = arith.addi %[[R9]], %[[R8]] : index
-        # CHECK: %[[R11:.+]] = vector.load %[[ALLOC]][%[[R10]], %[[C0]]] : memref<32x16xf16, #gpu.address_space<workgroup>>, vector<4xf16>
+        # CHECK: vector.store %4, %[[ALLOC]][%[[R3]], %[[C0]]] : memref<32x16xf16, #gpu.address_space<workgroup>>, vector<4xf16>
```
harsh-nod, line 308: Just wondering, should this %4 be %[[R11]] ?

lit_tests/kernel/wave/codegen.py
```diff
@@ -305,37 +305,19 @@ def mma(
         # CHECK: %[[R3:.+]] = arith.addi %[[R2]], %[[R1]] : index
         # CHECK: %[[R4:.+]] = vector.load %0[%[[R3]], %[[C0]]] : memref<64x16xf16, strided<[16, 1], offset: ?>>, vector<4xf16>
         # CHECK: %[[ALLOC:.+]] = memref.alloc() : memref<32x16xf16, #gpu.address_space<workgroup>>
-        # CHECK: %[[R5:.+]] = arith.muli %[[WG0]], %[[C32]] : index
```
harsh-nod, line 308: Nice, the power of cse in action :)

shark_turbine/kernel/wave/constraints.py
```diff
@@ -79,6 +79,14 @@ def linearized_thread_id(self) -> IndexExpr:
         ]
         return sum([x * y for x, y in zip(thread_ids, threads_per_block)])
 
+    # Inline substitution for vector_size given index map. In the future we can add support for other members.
+    def _subs(self, index_map: dict[IndexSymbol, int]):
+        if not isinstance(self.vector_shapes, dict):
```
harsh-nod, line 84: Isnt the type of vector shapes always dict? So when would we hit this case?

shark_turbine/kernel/wave/constraints.py
```diff
@@ -79,6 +79,14 @@ def linearized_thread_id(self) -> IndexExpr:
         ]
         return sum([x * y for x, y in zip(thread_ids, threads_per_block)])
 
+    # Inline substitution for vector_size given index map. In the future we can add support for other members.
+    def _subs(self, index_map: dict[IndexSymbol, int]):
```
harsh-nod, line 83: Since this is a public facing member function of hardware constraint, we can drop the leading underscore? My understanding was that leading underscore is for functions we don't want to expose to other modules/files.

harsh-nod (approved):
lgtm! minor comment about style, but overall looks good. thanks!

shark_turbine/kernel/wave/constraints.py
raikonenfnu: Some tests hit this case for some reason, that's why I added it there

shark_turbine/kernel/wave/constraints.py
raikonenfnu: I see I see, I was following the pytorch convention of doing `_` meaning it's doing an inline/on variable modification. i.e  `x._add(y) == x += y` while `c = x.add(y) == x+y`. What do you think? I can change it back to subs if you think that's better/more uniform with the overall styling. :)

lit_tests/kernel/wave/codegen.py
raikonenfnu: ahhh, right, good catch, thanks! :)

shark_turbine/kernel/wave/constraints.py
raikonenfnu: btw @harsh-nod, this is because if we specify MMA_type i.e (`mma_type=tkw.MMAType.F32_16x16x16_F16`), in HW constraints, we won't really be using `vector_shapes`. as also shown in the implementation here:
https://github.com/iree-org/iree-turbine/blob/1870973138768bbae4d278ade46782bdde63bcd4/shark_turbine/kernel/wave/expansion.py#L410-L415

shark_turbine/kernel/wave/constraints.py
harsh-nod: Cool, good to know. We haven't been following that convention so far, so I would probably stick with subs.

shark_turbine/kernel/wave/constraints.py
harsh-nod: Makes sense. So if vector_shapes are not defined, are they set to None and can we check for that? 
```
if self.vector_shapes is None:
  return
```

The other thing is that this function is meant to convert the vector_shapes from their symbolic types to static shapes. Could we rename to either subs_vector_shapes (if we want to restrict the scope for now) or finalize (which is what IndexingContext uses, this implies its more general and we don't have to specify what is being substituted)

Hardcode84 (approved):
Lets merge it!

shark_turbine/kernel/wave/constraints.py
raikonenfnu: Yeah that makes sense, in the future we'd probably also let `HardwareConstraint::subs` to call multiple of these `subs_member_variable`. :)

shark_turbine/kernel/wave/constraints.py
raikonenfnu: Renamed to `subs_vector_shapes ` based on the other comment. :)

harsh-amd: thanks!

PR #98, opened 2024-08-23
[TKW] Support arbitrary shapes in e2e tests

Hardcode84:


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -4,27 +4,34 @@
...
     N = tkl.sym.N
-    BLOCK_M = tkl.sym.BLOCK_M
-    BLOCK_N = tkl.sym.BLOCK_N
-    ELEMS_PER_THREAD = tkl.sym.ELEMS_PER_THREAD
     ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
 
+    wave_size = 64
+    BLOCK_M = 1
+    BLOCK_N = sympy.Min(N, 256)
+    ELEMS_PER_THREAD = BLOCK_N / wave_size
```
raikonenfnu, line 28: Since wave_e2e_test.py may be a collection of reference kernels for people to use, would be good to set some comments on why we select some of these.

For example, we can comment that `ELEMS_PER_THREAD = BLOCK_N / wave_size` is selected because we are distributing BLOCK_N to a single subgroup.

Hardcode84: Added comment

raikonenfnu (approved):

harsh-nod (approved):
thanks, this looks great! @erman-gurses - Ivan added the parameterization here so you can just use this e2e test for the GitHub runner.

PR #101, opened 2024-08-26
Get e2e gemm functional

harsh-nod:
This PR adds a generate_iree_ref test utility function that can be used to compare the outputs of tkw with iree. We also update the parameters of the gemm test and use generate_iree_ref to generate iree reference gemm outputs.

### DIFF
--- a/shark_turbine/kernel/ops/wave_ops.py
+++ b/shark_turbine/kernel/ops/wave_ops.py
@@ -18,11 +18,15 @@ from typing import (
 import torch.fx as fx
 
 from ..lang.wave_types import Memory, Register, IndexMapping
+from ..lang.global_symbols import *
 from .._support.indexing import IndexExpr, IndexSymbol, IndexSequence
 from .._support.dtype import DataType
 from .._support.regions import RegionGraph
 from .base import OpDispatcher
 from ..lang.global_symbols import MMA_ACC, MMA_LHS, MMA_RHS
+
+if TYPE_CHECKING:
+    from ..wave.constraints import Constraint
 
 T = TypeVar("T", bound=Type[Any])
 AccT = TypeVar("AccT")
@@ -386,6 +390,13 @@ class CustomOp(ABC):
         else:
             raise ValueError("Index must be a dict")
 
+    def post_expansion(self, constraints: list["Constraint"]) -> None:
+        """
+        Hook for post-expansion operations. This is called after the arguments
+        of the node are expanded.
+        """
+        pass
+
 
 @define_py_op(operator.getitem)
 @define_py_op(operator.add)
@@ -575,7 +586,7 @@ class Allocate(CustomOp):
         return Memory[*self.shape, self.address_space, self.dtype]
 
 
-@define_op("barrier")
+@define_op("shared_memory_barrier")
 @dataclass
 class SharedMemoryBarrier(CustomOp):
     """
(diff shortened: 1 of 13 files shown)


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -29,7 +29,7 @@ def test_copy(shape):
     wave_size = 64
     BLOCK_M = 1
     BLOCK_N = sympy.Min(N, 256)
-    ELEMS_PER_THREAD = BLOCK_N / wave_size
+    ELEMS_PER_THREAD = BLOCK_N
```
raikonenfnu, line 32: Why would we want such large number for elements per thread though?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -261,6 +261,9 @@ def _expand_node(
     new_node.fx_node.name = get_expanded_name(node, restricted_dims)
     node_index_setter(new_node, restricted_dims)
 
+    constraints = node_index_setter.args[0]
+    new_node.pre_expansion(constraints)
```
raikonenfnu, line 265: I don't think we are using pre_expansion for anything right now? Is there a plan to do so or is it just for symmetry? haha

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -374,6 +378,20 @@ def index(self, value: Any):
...
             raise ValueError("Index must be a dict")
 
+    def pre_expansion(self, constraints: list["Constraint"]) -> None:
+        """
+        Hook for pre-expansion operations. This is called before the arguments
+        of the node are expanded.
+        """
+        pass
+
+    def post_expansion(self, constraints: list["Constraint"]) -> None:
```
raikonenfnu, line 388: This is kind of nice, I can it see it being very useful in futureto propagate types and shape information for post and pre-expansion :)

shark_turbine/kernel/wave/promotion.py
```diff
@@ -39,12 +39,10 @@ def promote_node(
 
     assert isinstance(node, Read) or isinstance(node, Write)
     with node.graph.inserting_before(node.fx_node.next):
-        workgroup_distributed_shape = get_workgroup_distributed_shape(
-            node.type.symbolic_shape, constraints
-        )
+        constrained_shape = get_constrained_shape(node.type.symbolic_shape, constraints)
```
raikonenfnu, line 42: Does this mean we anticipate promoting memory that is not part of the workgroup level tiling? I guess that makes sense for case where we are tiling reduction dims i.e workgroup_K is typically full size but we may tile it within a loop and only want to promote the TilingConstraint or IV size . 

Is that correct, or are there other use cases that need this for?

tests/kernel/wave/wave_e2e_test.py
harsh-nod: Yes, you are right, we don't need this. Will update the PR.

shark_turbine/kernel/wave/expansion.py
harsh-nod: Yes, not being used right now. I can remove until we need it.

shark_turbine/kernel/wave/promotion.py
harsh-nod: Usually, we would only promote after workgroup distributing and any sort of tiling because we want to promote the smallest shape possible (since shared memory is a limited resource). This is only for that use case.

harsh-nod:
@raikonenfnu I have removed the iree reference code from the PR.

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -386,6 +390,13 @@ def index(self, value: Any):
         else:
             raise ValueError("Index must be a dict")
 
+    def post_expansion(self, constraints: list["Constraint"]) -> None:
```
raikonenfnu, line 393: This PR used to have pre-expansion mechanisms right? Just curious, are we saving it for later? haha

shark_turbine/kernel/wave/codegen.py
```diff
@@ -744,16 +744,20 @@ def handle_reduction(emitter: WaveEmitter, node: fx.Node):
...
+    end = arith_d.constant(IndexType.get(), int(dim // tile_size))
 
-    step = None
-    for constraint in emitter.constraints:
-        if isinstance(constraint, TilingConstraint) and constraint.dim == axis:
-            tile_size = constraint.tile_size.subs(idxc.subs)
-            step = arith_d.constant(IndexType.get(), int(tile_size))
+    # Since we divide the end by the tile size, we need to make sure that the
+    # step is 1.
+    step = arith_d.constant(IndexType.get(), int(1))
```
raikonenfnu, line 760: Nice fix/refactor! seems like we can do this because we already take tile_size into account for index computation during expansion phase?

raikonenfnu (approved):
Awesome stuff! had some small Qs but LGTM/land as is! :)

shark_turbine/kernel/ops/wave_ops.py
raikonenfnu: NVM, I asked for this before, please ignore!

shark_turbine/kernel/wave/codegen.py
harsh-nod: yes, we compute the index as induction_var * tile_size so by doing this, we don't double multiply.

PR #103, opened 2024-08-26
Partition writes based on access patterns

harsh-nod:
This PR adds a pattern to partition writes if it
determines that writes are strided (based on the index sequence). In this case, it partitions the write into a series of extract and writes for each individual element of the register being written.


shark_turbine/kernel/_support/tracing.py
```diff
@@ -121,6 +121,10 @@ class CapturedTrace:
     def __init__(self, region_graph: RegionGraph, root_graph: str):
         self.region_graph = region_graph
         self.root_graph = root_graph
+        self.region_graph.subgraphs[root_graph].subgraphs = {}
+        for name, subgraph in self.region_graph.subgraphs.items():
+            if name != root_graph:
+                self.region_graph.subgraphs[root_graph].subgraphs[name] = subgraph
```
raikonenfnu, line 127: is it not already structured this way? otherwise why would we be able to do `for name, subgraph in self.region_graph.subgraphs.items():` ?

raikonenfnu: ah nvm I see you are moving from `self.region_graph.subgraphs` into `self.region_graph.subgraphs[root_graph].subgraphs`. Seems useful for making the other subgraphs under the root_graph in terms of hierarchy, but not 100% sure where it's used

shark_turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    def has_strided_access(node: fx.Node) -> bool:
+        """
+        Checks for writes on 2d tensors with strided access on a single dimension.
+        """
+        custom = get_custom(node)
+        if isinstance(custom, Write) and len(custom.type.symbolic_shape) == 2:
+            strides = [simplify_index(custom.index[dim]).stride for dim in custom.index]
+            return sum(1 for stride in strides if stride > 1) == 1
```
raikonenfnu, line 39: Can we use `any(stride > 1 for stride in strides)` ?

shark_turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -0,0 +1,68 @@
...
+        return False
+
+    strided_operators = trace.walk(has_strided_access)
+    for operator in strided_operators:
+        custom = get_custom(operator)
+        simplified_index = {
+            dim: simplify_index(custom.index[dim]) for dim in custom.index
+        }
+        max_stride = int(max(simplified_index[dim].stride for dim in simplified_index))
+        hw_constraint = [c for c in constraints if isinstance(c, HardwareConstraint)][0]
```
raikonenfnu, line 49: Should hoist this out of the loop for cleanliness

shark_turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -0,0 +1,68 @@
...
+        idxc = IndexingContext.current()
+        elements_per_thread = custom.elements_per_thread
+        if isinstance(elements_per_thread, IndexSymbol):
+            elements_per_thread = elements_per_thread.subs(idxc.subs)
+        with custom.graph.inserting_before(operator):
+            for i in range(elements_per_thread):
+                extract = ExtractSlice(custom.register_, [i], [1], [1]).add_to_graph(
+                    custom.graph
+                )
+                offset = np.unravel_index(i * max_stride, shape)
```
raikonenfnu, line 60: What is the intuition behind delinearizing i * max_stride by the shape? not super obvious to me here

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -9,13 +9,20 @@ def set_register_shape(trace: CapturedTrace, custom: CustomOp) -> None:
     for custom_user in custom.users:
         if isinstance(custom_user, MMA):
             arg_index = custom_user.fx_node.args.index(custom.fx_node)
+            get_thread_shape = lambda index: max(x.size for x in index.values())
             match arg_index:
                 case 0:
-                    custom.fx_node.thread_shape = custom_user.lhs_index[0].size
+                    custom.fx_node.thread_shape = get_thread_shape(
```
raikonenfnu, line 15: For the case of lhs_index specifically(we can generalize this to rhs and acc_index), is the logic here since we set MMA_RHS and MMA_ACC to 0s and only set MMA_LHS to 1, only the LHS related dimension will be non zero here?

shark_turbine/kernel/wave/utils.py
```diff
@@ -108,3 +109,60 @@ def delinearize_index(index: IndexExpr, shape: list[int]) -> list[IndexExpr]:
...
+    mma_nodes = trace.walk(is_mma)
+    mapping: dict[IndexSymbol, int] = {}
+    for node in mma_nodes:
+        custom: MMA = get_custom(node)
+        m, n = custom.acc_type.symbolic_shape[-2:]
+        mapping[m] = 0
+        mapping[n] = 1
+        k = [
+            dim
+            for dim in custom.lhs_type.symbolic_shape
```
raikonenfnu, line 146: Maybe NIT:  Personal preference to express it as sets operations
```
k = list(set(custom.lhs_type.symbolic_shape).intersection(custom.rhs_type.symbolic_shape) - set(custom.acc_type.symbolic_shape))[0]
```

but this works too :)

raikonenfnu: > if dim in custom.rhs_type.symbolic_shape

Is custom.rhs_type.symbolic_shape already a set? if it is then probably even better to use the first comment set operations.  :)

harsh-nod: I dont think it is a set, but I think we can write this more concisely using sets. I will change.

shark_turbine/kernel/wave/index_sequence_analysis.py
raikonenfnu: I think we'd need to add a Not Implemented error for cases where multiple dims are strided since we do not handle it yet.

shark_turbine/kernel/wave/index_sequence_analysis.py
raikonenfnu: Documentation from offline chat: it is i * max_stride because for example if we are writing 4 elements to a transposed matrix with fastest dim len of 16, then we'd need to write it at index 0[0 * stride], 16 [1 * stride], 32[2 * stride] , 48[3 * stride]

raikonenfnu (approved):
Overall looks great just need to add NYI error for multiple strides then we are good to go :)

shark_turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: just added this in my next PR.

PR #104, opened 2024-08-27
Use more stable ordering of nodes during visualization

harsh-nod:
Previously, nodes were identified by their id(). This is unstable and would result in different files being generated for the same input. We fix this by using our manually specified ordering for the nodes.


Hardcode84 (approved):
LGTM, thanks

PR #105, opened 2024-08-27
[TKW] Add reduction operators and their lowerings

raikonenfnu:


lit_tests/kernel/wave/codegen.py
```diff
@@ -577,6 +577,82 @@ def test(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
...
+    @tkw.wave(constraints)
+    def test(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+        b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+        c: tkl.Memory[M, ADDRESS_SPACE, tkl.f16],
+    ):
+        lhs = tkw.read(a, elements_per_thread=ELEMS_PER_THREAD)
+        rhs = tkw.read(b, elements_per_thread=ELEMS_PER_THREAD)
+        res = lhs * rhs
+        res = tkw.sum(res)
```
Hardcode84, line 610: So, what the actual semantics of the `tkw.sum(res)`? I would expect it to reduce  input array into single scalar/0d-array. Shouldn't it be `tkw.sum(res, dim=N)` instead?

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -863,3 +875,54 @@ class ExtractSlice(CustomOp):
...
         return get_custom(self.register_).type
+
+
+@define_op("sum")
+@dataclass
+class ReduceOp(CustomOp, ABC):
+    """
+    Represents a Reduce computation.
+
+    arg: Soure tensor/value to reduce
```
Hardcode84, line 886: nit: Soure -> Source

lit_tests/kernel/wave/codegen.py
raikonenfnu: Ah that makes sense, done! :)

shark_turbine/kernel/ops/wave_ops.py
raikonenfnu: Good catch, thanks!

shark_turbine/kernel/wave/decompose_reduce_ops.py
```diff
@@ -0,0 +1,112 @@
...
+        if isinstance(c, TilingConstraint) or isinstance(c, WorkgroupConstraint)
+    }
+    subgroup_size = hardware_constraint.threads_per_wave
+    for node in reduce_nodes:
+        custom = get_custom(node)
+        with custom.graph.inserting_before(custom.fx_node):
+            reduction_src, reduction_acc, reduction_dim = node.args
+            binary_fn = TKW_COMBINER[custom.tkw_op_name]
+            if reduction_dim is None:
+                reduction_dim = custom.type.symbolic_shape[-1]
```
Hardcode84, line 83: As I mentioned in other comment, when dimension is not specified, user would expect reduction over all dimensions instead of fastest (following numpy rules). For now, let's just bail out and ask user to explicitly specify dimension.

Hardcode84 (approved):

shark_turbine/kernel/wave/decompose_reduce_ops.py
raikonenfnu: Makes sense, done, thanks! :)

shark_turbine/kernel/wave/codegen.py
```diff
@@ -541,6 +544,64 @@ def handle_mma(emitter: WaveEmitter, node: fx.Node):
...
+    element = vector_d.extract(src, static_position=static_pos, dynamic_position=[])
+    element_original_type = element.type
+
+    # Pad to 32 bit if needed.
+    if not _is_float_type(element.type):
+        raise NotImplementedError("Currently only support shuffle for floats.")
+    if element.type.width > 32:
+        raise ValueError("Cannot shuffle more than 32 bit.")
+    elif element.type.width < 32:
+        element = arith_d.extf(F32Type.get(), element)
```
harsh-nod, line 591: I will add more comments as I review, but this stuck out to me as the one that will hurt performance the most. If we have f16, rather than extending to f32, we can pack to f16s together and halve the total number of shuffles.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -541,6 +544,64 @@ def handle_mma(emitter: WaveEmitter, node: fx.Node):
...
+    if math.prod(src.type.shape) != 1:
+        raise NotImplementedError("Currently only support unit vector for shuffleOp.")
+
+    # Scalarize (vector<FLOAT_TYPE> -> FLOAT_TYPE).
+    static_pos = [0 for i in range(src.type.rank)]
+    element = vector_d.extract(src, static_position=static_pos, dynamic_position=[])
+    element_original_type = element.type
+
+    # Pad to 32 bit if needed.
+    if not _is_float_type(element.type):
```
harsh-nod, line 586: If you look at the iree code, everything is always cast to i32 so you dont have to worry about support ints and floats. They both go down the same path.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -541,6 +544,64 @@ def handle_mma(emitter: WaveEmitter, node: fx.Node):
...
+
+    TODO: Handle non-unit vector types such as vector<4xF8> (useful for resolving layouts).
+    """
+    try:
+        src, offset, width = node.args
+    except ValueError as e:
+        raise ValidationError("Malformed arguments") from e
+    if not isinstance(offset, int) or not isinstance(width, int):
+        raise NotImplementedError(
+            "Non-const wdith or offset is not yet implemented for shuffleOp."
```
harsh-nod, line 568: typo: width

shark_turbine/kernel/wave/codegen.py
raikonenfnu: Agreed! this is indeed the next action item/optimizations that I am going to do after I get max in. The reason I am not doing it sooner is because we won't need it yet for the classic F16-FA + F32-Softmax.

shark_turbine/kernel/wave/codegen.py
raikonenfnu: Yeap! In IREE we need to cast it to i32 for shuffle because Vulkan-SPIRV do not handle F32 shuffles. Although for iree-turbine, I'd prefer to have a separate path for int and float since it is relatively simple(just another if condition with exti) and it saves 2 bitcastings per shuffle! :)

shark_turbine/kernel/wave/codegen.py
raikonenfnu: Good catch, I'll push up a fix for this :)

shark_turbine/kernel/wave/decompose_reduce_ops.py
```diff
@@ -0,0 +1,114 @@
...
+            # Global Reduce
+            global_reduction = emit_global_reduction(
+                binary_fn, local_reduction, custom.graph, subgroup_size
+            )
+
+            # Local Accumulator Reduce
+            final_reduction = global_reduction
+            if reduction_acc is not None:
+                final_reduction = get_graph_node(
+                    binary_fn(reduction_acc, global_reduction), graph
```
harsh-nod, line 108: graph here is not defined in this function, should it be custom.graph?

shark_turbine/kernel/wave/codegen.py
harsh-nod: Sure, maybe add a todo ?

raikonenfnu: I have a TODO on the description of the handle_shuffle function,
"TODO: Handle non-unit vector types such as vector<4xF8> (useful for resolving layouts)."

But I can add one on top of the NotImplementedError for better visibility. :)

shark_turbine/kernel/wave/utils.py
```diff
@@ -77,10 +78,12 @@ def DCE(trace: CapturedTrace):
     def is_removable_operator(node: fx.Node) -> bool:
         custom = get_custom(node)
         idxc = IndexingContext.current()
-        is_global_write = (
-            isinstance(custom, Write)
-            and custom.type.address_space.subs(idxc.subs) == GLOBAL_ADDRESS_SPACE
+        is_global_write = isinstance(custom, Write) and (
+            custom.type.address_space.subs(idxc.subs) == GLOBAL_ADDRESS_SPACE
+            or custom.type.address_space.subs(idxc.subs)
+            == tkl.AddressSpace.GLOBAL_MEMORY.value
```
harsh-nod, line 84: Why do we need this comparison test for tkl.AddressSpace.GLOBAL_MEMORY.value ?

raikonenfnu: this is for DCE

raikonenfnu: Currently, the wave e2e tests are using `tkl.AddressSpace.GLOBAL_MEMORY.value` for ADDRESS_SPACE  as convention.

shark_turbine/kernel/wave/decompose_reduce_ops.py
raikonenfnu: Good catch, I'll push up a fix for this :)

PR #107, opened 2024-08-28
[TKW] Im2col e2e test

Hardcode84:


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -180,3 +181,104 @@ def test(
...
+        tkw.HardwareConstraint(
+            threads_per_wave=wave_size,
+            waves_per_block=(1, 1, 1),
+            vector_shapes={M: BLOCK_M, K: ELEMS_PER_THREAD},
+        )
+    ]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(K, BLOCK_K, 1)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
+    constraints += [tkw.WaveConstraint(K, BLOCK_K)]
```
raikonenfnu, line 242: does WaveConstraint do unexpected behavior if undefined as opposed to erroring out?

Hardcode84: Without any K constraints example is still compiles successfully and K access is still unrolled according to `vector_shapes`, but it gets indices weird: K index is always 0.

Hardcode84:
@harsh-nod I've tried to rebase on top of https://github.com/iree-org/iree-turbine/pull/108, and im2col actually breaks, looking into it

harsh-nod:
Let me know if it still breaks with the commit.

Hardcode84:
@harsh-nod I'm still getting wrong results, current read indexing looks like
```
{M: 64*$WG0 + 64*floor($T0/64), K: 12*$T1 + 8*$WG1 : 4 : 1}
{M: 64*$WG0 + 64*floor($T0/64), K: 12*$T1 + 8*$WG1 + 4 : 4 : 1}
```
While `M` should be just `64*$WG0 + $T0` for kernel to work

harsh-nod:
> @harsh-nod I'm still getting wrong results, current read indexing looks like
> 
> ```
> {M: 64*$WG0 + 64*floor($T0/64), K: 12*$T1 + 8*$WG1 : 4 : 1}
> {M: 64*$WG0 + 64*floor($T0/64), K: 12*$T1 + 8*$WG1 + 4 : 4 : 1}
> ```
> 
> While `M` should be just `64*$WG0 + $T0` for kernel to work

So why is 64*$WG0 + $T0 the right index? If you are trying to read elements along the M dimension, then shouldn't there be a stride to the reads? So something like 64*$WG0 + $T0 + stride?

harsh-nod:
The current code that I pushed in sets up an access pattern where we only load from the contiguous dimension. You can see that in the indices above which show that 4 elements are being loaded along the K dimension with stride 1.

Hardcode84:
The idea was for for each thread to read/write individual row, but with indexing like this all the threads will just access row 0 (WG size is 64).

Hardcode84:
rebased, everything works now

shark_turbine/kernel/wave/codegen.py
```diff
@@ -368,6 +368,9 @@ def _construct_gather_scatter_indices(
         ), "non-identity input mapping is not supported yet"
         index_mapping = mapping.map_output_indices(symbolc_shape)
 
+    idxc = IndexingContext.current()
+    index_mapping = tuple(i.subs(idxc.subs) for i in index_mapping)
```
raikonenfnu, line 372: For my own education, do you have example on how index mapping used to be and how it is with this change?

raikonenfnu: as in what gets subbed during this index mapping, the tile sizes? haha

Hardcode84:
```
Before: (floor($index0/((H - HF + 1)*(W - WF + 1))), floor($index1/(HF*WF)), floor((Mod($index1, HF*WF))/WF) + floor((Mod($index0, (H - HF + 1)*(W - WF + 1)))/(W - WF + 1)), Mod($index0, W - WF + 1) + Mod($index1, WF))
After:  (floor($index0/64), floor($index1/4), floor((Mod($index0, 64))/8) + floor((Mod($index1, 4))/2), Mod($index0, 8) + Mod($index1, 2))
```
Mapping can contain arbitrary symbols and we need to substitute them. Previous transpose example worked without it, because it was only using iterators (`$indexN`) and not any other symbols.

raikonenfnu: Makes sense yeah! :)

raikonenfnu (approved):

PR #108, opened 2024-08-28
Specify thread-level access pattern when using vector_shapes

harsh-nod:


Hardcode84:
So, here is how im2col IR and indices looks like before and after this change.
https://gist.github.com/Hardcode84/b935527ccb011c31ec849c3613701ed7

I'm not quite understand what's going on with indices after this change

harsh-nod:
> So, here is how im2col IR and indices looks like before and after this change. https://gist.github.com/Hardcode84/b935527ccb011c31ec849c3613701ed7
> 
> I'm not quite understand what's going on with indices after this change

Can you try with the latest PR? I might have fixed some of the issues?

shark_turbine/kernel/wave/constraints.py
```diff
@@ -92,7 +103,27 @@ def subs_vector_shapes(self, index_map: dict[IndexSymbol, int]):
...
+        elements_per_thread: int | IndexSymbol,
+    ) -> IndexSequence:
+        if dim not in self.vector_shapes:
+            raise ValueError(f"No vector shape specified for dimension {dim}")
+        thread_id = self.get_thread_id_from_workgroup_dim(workgroup_dim)
+        return IndexSequence(thread_id * elements_per_thread, elements_per_thread, 1)
+
+    def apply(
+        self,
+        mma_index: int,
```
raikonenfnu, line 119: NIT: Should we call it constraint_index now?

shark_turbine/kernel/wave/constraints.py
```diff
@@ -92,7 +103,27 @@ def subs_vector_shapes(self, index_map: dict[IndexSymbol, int]):
...
+        dim: IndexSymbol,
+        elements_per_thread: int | IndexSymbol,
+    ) -> IndexSequence:
+        if self.vector_shapes is not None:
+            return self.compute_access_pattern_using_vector_shapes(
```
raikonenfnu, line 124: For matmul applications, would it ever go to the Piecewise function part? This looks like we would always return early, since I reckon even for MMA cases we'd have `self.vector_shapes` ?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -129,15 +147,42 @@ def set_node_index(
...
+                # The semantics of elements_per_thread are that it represents the number of
+                # elements that are loaded contiguously from memory.
+                elements_per_thread = getattr(custom, "elements_per_thread", None)
+                constraint_index, elements_per_thread = (
+                    (
+                        workgroup_constraints[dim].workgroup_dim,
+                        (
+                            0
+                            if not is_contiguous_dim(
+                                dim,
```
raikonenfnu, line 174: Do we expect the indexExpr to be `0, 0, 1` after this or should we just raise error?

shark_turbine/kernel/wave/constraints.py
harsh-nod: sure I can change.

shark_turbine/kernel/wave/expansion.py
harsh-nod: So this is why I added the comment about the semantics of elements_per_thread. What it specifies is the number of elements that can be read contiguously by a single thread. But when you specify that on a read, the read could have a type of [M, N]. So what this function does is specify the access pattern of N to be consistent with elements_per_thread and sets the access_pattern of M to be 0 elements. So we should not raise an error in this case, it is the expected access pattern along the M dimension.

shark_turbine/kernel/wave/constraints.py
```diff
@@ -92,7 +103,27 @@ def subs_vector_shapes(self, index_map: dict[IndexSymbol, int]):
...
+    def compute_access_pattern_using_vector_shapes(
+        self,
+        dim: IndexSymbol,
+        workgroup_dim: int,
+        elements_per_thread: int | IndexSymbol,
+    ) -> IndexSequence:
+        if dim not in self.vector_shapes:
+            raise ValueError(f"No vector shape specified for dimension {dim}")
+        thread_id = self.get_thread_id_from_workgroup_dim(workgroup_dim)
+        return IndexSequence(thread_id * elements_per_thread, elements_per_thread, 1)
```
raikonenfnu, line 115: Nice! now we have the quintessential elementwise index setting that can take `WorkgroupConstraint(M, BLOCK_M, 0) + element_per_thread` to `tid.x * elem_per_thread : tid.x * elem_per_thread + elem_per_thread ` indexing in M dim! :)

harsh-nod: You also need to specify vector_shape, but yes! :)

shark_turbine/kernel/wave/expansion.py
raikonenfnu: Understood, thanks for the explanation! :)

May or may not mean anything but shouldn't the generated IndexSequence here be 0,1,1? since there is 1 element there?

shark_turbine/kernel/wave/constraints.py
raikonenfnu: But in this PR seems like we are only using vector_shapes to ensure that we are accessing the contiguous/fastest dimension right?

shark_turbine/kernel/wave/expansion.py
harsh-nod: So that's an interesting question. When you are loading a 1d vector from dimension N of a [M, N] tensor then are you loading [0, 4] elements or [1, 4] elements ? Basically we can represent a 1d vector as vector<4xf32> (which corresponds to 0 elements along the M dimension) or as vector<1x4xf32> (which corresponds to 1 element along the M dimension). They both represent the same quantity, so makes sense to go with the most concise representation which corresponds to M = 0 (vector<4xf32>).

raikonenfnu: Ah OK, that makes sense, thanks! :)

raikonenfnu (approved):
LGTM! just 1 nit left but then should be good to land after! :)

shark_turbine/kernel/wave/constraints.py
harsh-nod: When using vector_shapes we need the operators to define elements_per_thread which specifies the linear access pattern and elements_per_thread requires that the data being loaded be contiguous. So yes, use of vector_shapes implies contiguous accesses.

PR #109, opened 2024-08-28
Add yaml file for perf

erman-gurses:
This PR and future related PRs will be transformed into performance-CI, benchmarking different operators with various input parameters. For now, we are starting with GEMMs. In the future, we will add convolution, flash attention, and other operators as needed. The ultimate goal is to create performance plots for those operators so we can see their performance daily.

The `perf.yaml` file is added to the repo to distinguish between `Performance` and `CI `related operations.


saienduri:
Wait what's the motivation here? Looks like it's running the same commands to me. A PR description would help

ScottTodd:
> The motivation is that we want a separate performance-ci that is only run nightly and can be triggered with a button. (as opposed to the ci that is triggered on every commit). For now, they are identical but we will change the performance-ci to be more compute heavy in the future.

That general direction sounds good to me, but can someone write out more of a plan in the PR description and/or an issue? I'm curious which tests /  benchmarks will be running. I want us to be able to connect what the workflows are running with the project overview in the README, for example (AOT export, eager execution, custom ops). I think some organization around the project roadmap will go a long way.

erman-gurses:
> > The motivation is that we want a separate performance-ci that is only run nightly and can be triggered with a button. (as opposed to the ci that is triggered on every commit). For now, they are identical but we will change the performance-ci to be more compute heavy in the future.
> 
> That general direction sounds good to me, but can someone write out more of a plan in the PR description and/or an issue? I'm curious which tests / benchmarks will be running. I want us to be able to connect what the workflows are running with the project overview in the README, for example (AOT export, eager execution, custom ops). I think some organization around the project roadmap will go a long way.

This PR and future related PRs will be transformed into performance-CI, benchmarking different operators with various input parameters. For now, we are starting with GEMMs. In the future, we will add `convolution `and `flash attention` as well as other operators as needed. The ultimate goal is to create performance plots for those operators so we can see their performance daily.

saienduri:
Cool, that sounds good. Can we push adding this CI to when the workflows are actually different. As long as they are the same, you can just use the CI trigger/nightly that is happening

PR #110, opened 2024-08-28
[TKW] Implement define interface op and tk.max

raikonenfnu:
In this PR, we add 3 things:

    1. Define Interface Op
    2. Reduction max op
    3. elementwise maximum op

Define interface op is created to generate new subclass op that can
derive from base interface class. this is useful for custom TK ops that
has very similar properties such as "sum" and "max" S.T we can reuse
code as much as possible.

### DIFF
--- a/shark_turbine/kernel/ops/wave_ops.py
+++ b/shark_turbine/kernel/ops/wave_ops.py
@@ -92,7 +92,19 @@ def exp2(src: "Register") -> "Register":
     ...
 
 
+def maximum(lhs: "Register", rhs: "Register") -> "Register":
+    ...
+
+
 def sum(
+    src: "Register",
+    acc: Optional["Register"] = None,
+    dim: Optional[IndexExpr | int] = None,
+) -> "Register":
+    ...
+
+
+def max(
     src: "Register",
     acc: Optional["Register"] = None,
     dim: Optional[IndexExpr | int] = None,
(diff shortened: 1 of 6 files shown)


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -234,4 +234,4 @@ def test(
         run_config=config,
     ):
         test(a, b, c)
-        assert_allclose(ref, c, atol=0.1)
+        assert_allclose(ref, c, atol=0.15)
```
harsh-nod, line 237: Why did the tolerance go up? This is doing a sum so it should have remained the same? Also can you add a max e2e test (for which I am guessing you should match exactly?) And not for this PR, but would be nice to compare against iree to see whether there are any differences (there should be none?)

harsh-nod (changes requested):

tests/kernel/wave/wave_e2e_test.py
raikonenfnu: great Q, yeap, it is unrelated to this PR. But there was one instance the error gets up to 0.125, since test generate random values. it was for a really large number along the magnitude of 84.872 to 84.747. So I set 0.15 just to make sure it doesn't break ever. I can also set manual seed instead. :)

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -1,6 +1,7 @@
 from .._support.tracing import CapturedTrace
 from ...support.logging import get_logger
-from ..ops.wave_ops import *
+from ..ops.wave_ops import get_custom, NewRegister, CustomOp, MMA, Reduction
+import torch.fx as fx
```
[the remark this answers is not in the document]
raikonenfnu: register_analysis uses "max", but since we import *, it imported tkw.max as "max" hahaha

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -180,6 +192,53 @@ def new_function(*args: Any, **kwargs: dict[str, Any]):
...
+def define_interface_op(op_name: str) -> Callable[[T], T]:
+    """
+    Generate new subclass for op handling, deriving from the base interface class.
+    Generated subclass can be used for emitting the op from compiler/python side,
+    by calling the generated subclass name.
+
+    The generated subclass name would be pascal case of the TKW op name. For example:
+    "tkw.op_name" -> "OpName"
+    "tkw.exp2" -> "Exp2"
+    """
```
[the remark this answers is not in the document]
raikonenfnu: thanks! :)

shark_turbine/kernel/wave/register_analysis.py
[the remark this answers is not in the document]
harsh-nod: ah makes sense, thanks!

tests/kernel/wave/wave_e2e_test.py
harsh-nod: yeah let's set a manual seed so the tests results are reproducible.

raikonenfnu: SG! let me push this up in a separate PR for better visibilityy and cleaner git commit history  :)

harsh-nod: Sounds good. Have you validated that the max reduction works e2e?

raikonenfnu: yeap! I have another repo where I collect and test TK-Kernels, LMK push it up here as well

raikonenfnu: Added test, and this one doesn't even need the atol modified. 😁

harsh-nod: nice! I think you can probably set to exact equality and it should still work because there is no loss in precision during max.

harsh-nod (approved):
thanks!

tests/kernel/wave/wave_e2e_test.py
raikonenfnu: ah shoot, missed this comment, I'll put it up in another PR fixing the reduce_sum test with manual seed. :)

PR #111, opened 2024-08-29
Add iree_ref for gemm functionality test

harsh-nod:
This PR adds utils to generate an iree reference
mmt kernel that can be compiled and run to
obtain a golden output for comparison. Compared
to the torch output, the advantage of this output
is that it can be matched exactly and makes for
an easier comparison target.

### DIFF
--- a/shark_turbine/kernel/wave/iree_utils.py
+++ b/shark_turbine/kernel/wave/iree_utils.py
@@ -1,0 +1,54 @@
+import torch
+from .utils import compile_and_invoke
+
+
+def get_mmt_asm(lhs_type: str, rhs_type: str, acc_type: str) -> str:
+    acc_dtype = acc_type.split("x")[-1]
+    matmul_function = (
+        f"func.func @mmt(%lhs: tensor<{lhs_type}>, %rhs: tensor<{rhs_type}>) -> tensor<{acc_type}> {{\n"
+        f"  %c0 = arith.constant 0.0 : {acc_dtype}\n"
+        f"  %init = tensor.empty() : tensor<{acc_type}>\n"
+        f"  %inital_result = linalg.fill ins(%c0 : {acc_dtype}) outs(%init : tensor<{acc_type}>) -> tensor<{acc_type}>\n"
+        f"  %result = linalg.matmul_transpose_b ins(%lhs, %rhs: tensor<{lhs_type}>, tensor<{rhs_type}>)\n"
+        f"             outs(%inital_result: tensor<{acc_type}>) -> tensor<{acc_type}>\n"
+        f"  return %result : tensor<{acc_type}>\n"
+        f"}}\n"
+    )
+    return matmul_function
+
+
+def dtype_str(dtype: torch.dtype) -> str:
+    if dtype == torch.float32:
+        return "f32"
+    elif dtype == torch.float16:
+        return "f16"
+    else:
+        raise ValueError(f"Unsupported dtype: {dtype}")
+
+
+def get_type_str(shape: tuple[int], dtype: torch.dtype) -> str:
+    return "x".join([str(x) for x in shape] + [dtype_str(dtype)])
+
+
+def generate_iree_ref(
+    kernel_type: str,
+    kernel_inputs: list[torch.Tensor],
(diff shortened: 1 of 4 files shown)


harsh-nod:
This test will fail until https://github.com/iree-org/iree-turbine/pull/101 lands.

shark_turbine/kernel/wave/iree_utils.py
```diff
@@ -0,0 +1,54 @@
+import torch
+from .utils import compile_and_invoke
+
+
+def get_mmt_asm(lhs_type: str, rhs_type: str, acc_type: str) -> str:
+    acc_dtype = acc_type.split("x")[-1]
```
raikonenfnu, line 6: [No action for now]: I think in the future we can also update this to use jinja templates S.T we can create separate files for MLIR templates.

shark_turbine/kernel/wave/iree_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    return matmul_function
+
+
+def dtype_str(dtype: torch.dtype) -> str:
+    if dtype == torch.float32:
+        return "f32"
+    elif dtype == torch.float16:
+        return "f16"
+    else:
+        raise ValueError(f"Unsupported dtype: {dtype}")
```
raikonenfnu, line 26: NIT: Can we use https://github.com/iree-org/iree-turbine/blob/36c99a686d5790add276bae3a88aa33dfb23c939/shark_turbine/support/conversions.py#L66-L86 instead?

shark_turbine/kernel/wave/iree_utils.py
Hardcode84: I think we can already do
```python
matmul_function = f"""
func.func @mmt(%lhs: tensor<{lhs_type}>, %rhs: tensor<{rhs_type}>) 
...
"""

File: shark_turbine/kernel/wave/utils.py [hunk 3]
```diff
@@ -181,3 +190,118 @@ def remove_global_indexing(
...
+    func_name: str,
+    config: dict[str, str],
+    kernel_inputs: list[torch.Tensor],
+    kernel_outputs: list[torch.Tensor],
+    run: bool = False,
+    run_bench: bool = False,
+):
+    backend = config["backend"]
+    device = config["device"]
+    flags = [
```
raikonenfnu, line 243: Just curious do we not need to specify a different flag when we start compilation from distributed/lower level  vector dialect VS linalg? :O

shark_turbine/kernel/wave/iree_utils.py
harsh-nod: Good point on the jinja templates. Thanks for the pointer of the f-triple string, will change to use that.

shark_turbine/kernel/wave/iree_utils.py
harsh-nod: sure, good find!

shark_turbine/kernel/wave/iree_utils.py
raikonenfnu: Ooo, if that is possible, that is also very nice haha

raikonenfnu:
Overall looks awesome! just some NITs and Qs then we should be good to go! :)

shark_turbine/kernel/wave/utils.py
harsh-nod: Surprisingly not :) You can call it good design on the iree side :)

raikonenfnu: Nice! :D

shark_turbine/kernel/wave/iree_utils.py
Hardcode84: + you don't need `\n` in multiline literals

raikonenfnu (approved):
Looks great! feel free to land after Ivan's comment on `\n` for mult-line literal! :)

shark_turbine/kernel/wave/iree_utils.py
harsh-nod: thanks you learn something new everyday :)

PR #113, opened 2024-08-29
[TKW] Fix thread offset on non-fastest dimension

Hardcode84:
* Thread offset for non-fastest dimension must be 1 instead of 0 so indexing actually takes thread_id into account
* Change `type.symbolic_shape` to `indexing_dims`, they should be the same for normal read/writes, but with presence of mapping dims must be extracted from mapping and `indexing_dims` is already doing this.

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -95,12 +95,13 @@ def test_read():
         # CHECK:            %[[D2:.+]] = arith.divsi %[[THREAD_ID_X]], %[[C64]] : index
         # CHECK:            %[[D3:.+]] = arith.muli %[[D2]], %[[C16_0]] : index
         # CHECK:            %[[D4:.+]] = arith.addi %[[D3]], %[[D1]] : index
+        # CHECK:            %[[D4_1:.+]] = arith.addi %[[D4]], %[[THREAD_ID_X]] : index
         # CHECK-DAG:        %[[C16_1:.+]] = arith.constant 16 : index
         # CHECK:            %[[D5:.+]] = arith.muli %[[WORKGROUP_ID_1]], %[[C16_1]] : index
         # CHECK-DAG:        %[[C32:.+]] = arith.constant 32 : index
         # CHECK:            %[[D6:.+]] = arith.muli %[[THREAD_ID_Y]], %[[C32]] : index
         # CHECK:            %[[D7:.+]] = arith.addi %[[D6]], %[[D5]] : index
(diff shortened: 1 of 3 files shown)


lit_tests/kernel/wave/codegen.py
```diff
@@ -199,18 +201,19 @@ def test(
         # CHECK-DAG:        %[[THREAD_ID_Y:.+]] = gpu.thread_id  y
         # CHECK:            %[[D0:.+]] = stream.binding.subspan %[[ARG0]][%[[C0]]] : !stream.binding -> memref<16x16xf16,
         # CHECK-SAME:         strided<[16, 1], offset: ?>>
-        # CHECK:            %[[D1:.+]] = arith.muli %[[WORKGROUP_ID_0]], %[[C16]] : index
+        # CHECK:            %[[D1:.+]] = arith.muli %workgroup_id_0, %[[C16]] : index
```
raikonenfnu, line 204: why do we not use the LIT variable?

harsh-nod (approved):
Thanks! Good to know that indexing_dims takes into account the mapping. Can you see a reason why we should use the symbolic_shape API or should we all move to indexing_dims API?

lit_tests/kernel/wave/codegen.py
Hardcode84: copypaste error )

shark_turbine/kernel/wave/expansion.py
```diff
@@ -169,10 +169,10 @@ def set_node_index(
                     (
                         workgroup_constraints[dim].workgroup_dim,
                         (
-                            0
+                            1
                             if not is_contiguous_dim(
                                 dim,
-                                custom.type.symbolic_shape,
+                                custom.indexing_dims,
```
raikonenfnu, line 175: Seems correct, but I would like to know how was it breaking before?

Hardcode84: Indexing dims for the read currently implemented like this:
```
    @property
    def indexing_dims(self) -> list[IndexSymbol]:
        if self.mapping is not None:
            return list(self.mapping.output_shape)
        # TODO: This could contain ints.
        return list(self.type.symbolic_shape)
```
If mapping is present we are using shape from mapping for indexing (output shape for read, input shape for write).
Without this fix `im2col` read will use `N, C, H, W` for indexing instead of expected `M, K`

raikonenfnu: Ahhh, makes sense, thanks! :)

raikonenfnu (approved):
Very awesome work! Thanks Ivan :)

PR #116, opened 2024-09-04
Add graph utils for modulo scheduling

harsh-nod:
This PR adds utils to find strongly connected components and simple cycles in directed graphs. These are required for implementing modulo scheduling.

### DIFF
--- a/shark_turbine/kernel/wave/scheduling/graph_utils.py
+++ b/shark_turbine/kernel/wave/scheduling/graph_utils.py
@@ -1,0 +1,132 @@
+import torch.fx as fx
+from random import shuffle, seed
+from collections import defaultdict
+
+
+def find_strongly_connected_components(
+    graph: fx.Graph, random_seed: int
+) -> dict[fx.Node, list[fx.Node]]:
+    """
+    Find the strongly connected components in the graph.
+    Returns a list of strongly connected components.
+    """
+    seed(random_seed)
+    initial_times = list(range(len(graph.nodes)))
+    shuffle(initial_times)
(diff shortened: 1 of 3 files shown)


shark_turbine/kernel/wave/scheduling/graph_utils.py
```diff
@@ -0,0 +1,132 @@
...
+            t, s = run_dfs(node, t, s, reverse, visited_nodes)
+
+
+def run_dfs(
+    node: fx.Node,
+    t: int,
+    s: fx.Node,
+    reverse: bool,
+    visited_nodes: set[fx.Node],
+) -> None:
```
Hardcode84, line 47: return type should be `Tuple[int, fx.Node]` I believe.

shark_turbine/kernel/wave/scheduling/graph_utils.py
```diff
@@ -0,0 +1,132 @@
...
+from random import shuffle, seed
+from collections import defaultdict
+
+
+def find_strongly_connected_components(
+    graph: fx.Graph, random_seed: int
+) -> dict[fx.Node, list[fx.Node]]:
+    """
+    Find the strongly connected components in the graph.
+    Returns a list of strongly connected components.
```
Hardcode84, line 11: Some links to the actual algorithm will help, I assume you are doing https://en.wikipedia.org/wiki/Kosaraju%27s_algorithm, sorting by `node.f` instead of maintaining explicit nodes stack.

Hardcode84 (approved):

PR #118, opened 2024-09-04
[TKW] Read type fixes

Hardcode84:
* Change `Read` return type from from `Memory` to `Register` and make it respect mapping shape. 
* Add `memory_type` property to get the original mem type
* Update related passes
* Some smaller fixes:
  * `_contains` func to properly handle case when input is `None`
  * `safe_subs` to properly handle case when input is not a sympy expr (e.g. when `address_space` has explicit value set instead of sympy placeholder)
  * `custom.axis.name` -> `str(custom.axis)` to handle cases when axis is expr and not just a symbol

### DIFF
--- a/shark_turbine/kernel/ops/wave_ops.py
+++ b/shark_turbine/kernel/ops/wave_ops.py
@@ -784,12 +784,16 @@ class Read(CustomOp):
         if self.mapping is not None:
             return list(self.mapping.output_shape)
         # TODO: This could contain ints.
-        return list(get_custom(self.memory).type.symbolic_shape)
+        return list(self.memory_type.symbolic_shape)
 
     @property
     def type(self) -> "Register":
-        dtype = get_custom(self.memory).type.dtype
+        dtype = self.memory_type.dtype
         return Register[*self.indexing_dims, dtype]
+
+    @property
+    def memory_type(self) -> "Memory":
+        return get_custom(self.memory).type
 
     @property
     def write_dependency(self) -> fx.Node:
@@ -911,6 +915,10 @@ class Write(CustomOp):
 
     @property
     def type(self) -> "Memory":
+        return get_custom(self.memory).type
+
+    @property
+    def memory_type(self) -> "Memory":
         return get_custom(self.memory).type
 
     @property
--- a/shark_turbine/kernel/ops/wave_ops.py
+++ b/shark_turbine/kernel/ops/wave_ops.py
@@ -791,12 +791,7 @@ class Read(CustomOp):
     @property
     def type(self) -> "Register":
         dtype = self.memory_type.dtype
-        res = Register[*self.indexing_dims, dtype]
-        # res = self.memory_type
-        # print(self)
-        # import sys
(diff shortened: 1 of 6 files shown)


shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -765,9 +765,11 @@ def post_expansion(self, constraints: list["Constraint"]) -> None:
         self.rhs.index = self.rhs_index
         self.acc.index = self.acc_index
 
-        if self.lhs_type.address_space == SHARED_ADDRESS_SPACE:
+        # TODO: this is really wrong place for it, it relies on specific kernel structure,
+        # generated, if mma input is not come from load, things will breaks.
```
harsh-nod, line 769: Yes good point. With the separation into type and memory_type, not all ops will have a memory_type and so this won't work. Can you move this into the index setter for Read and Write Ops which I think will be a better place for this? So the idea is that we set the lhs, rhs and acc indices here. But within the Read and Write index setters, we can check if the memory type has a shared address space and if so we remove the global indexing.

shark_turbine/kernel/wave/utils.py
```diff
@@ -305,3 +305,14 @@ def compile_and_invoke(
             **benchmark_flags,
         )
         _print_bench_result(benchmark_results, bench_file)
+
+
+def safe_subs(input: Any, subs: List[Tuple[IndexExpr, IndexExpr]]) -> Any:
+    """
+    Substitute input using provided `subs` list if input is sympy object.
+    Otherwise return input unchanged.
+    """
```
harsh-nod, line 314: Nice! Should we instantiate the indexing context within this function so you don't need to pass subs or you can make subs optional? That will avoid having to instantiate an indexing context multiple times in the code.

harsh-nod (changes requested):
Nice! Thanks for the cleanup. Just a comment about where to put the code to remove global indexing.

raikonenfnu (commented):
Aside from Harsh's Qs, looks good to me!

shark_turbine/kernel/wave/utils.py
Hardcode84: Done

shark_turbine/kernel/ops/wave_ops.py
Hardcode84: I tried to make index setter smart, but it didn't wok., `remove_global_indexing` need access to constraints, which are not available in setter.

Hardcode84:
+ MI300 CI seems to be broken, working fine locally on my snapshot (IREE version update?)

shark_turbine/kernel/ops/wave_ops.py
harsh-nod: Thanks for trying, I can take a look at this.

harsh-nod (approved):
lgtm! thanks!

harsh-nod:
Usually re-triggering the job works but we should take a closer look at this. @saienduri

PR #119, opened 2024-09-04
Add e2e parameterization using file

erman-gurses:
This PR and future related PRs will be transformed into performance-CI, benchmarking different operators with various input parameters. For now, we are starting with GEMMs. In the future, we will add convolution, flash attention, and other operators as needed. The ultimate goal is to create performance plots for those operators so we can see their performance daily.

The perf.yaml file is added to the repo to distinguish between Performance and CI related operations.


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -7,17 +7,21 @@
...
 import torch
+import json
 
 _run_e2e = int(os.environ.get("WAVE_RUN_E2E_TESTS", 0))
-
 require_e2e = pytest.mark.skipif(not _run_e2e, reason="e2e tests are disabled")
 
-
-_test_shapes = [(1, 128), (256, 64), (256, 128), (256, 256), (256, 1024)]
+test_copy_shapes = [(1, 128), (256, 64), (256, 128), (256, 256), (256, 1024)]
```
harsh-nod, line 15:
```
default_test_shapes = [...]
user_specified_test_shapes = None
test_params_path = os.environ.get("TEST_PARAMS_PATH", "")
# Load JSON file only once
if os.path.isfile(test_params_path):
    with open(test_params_path, "r") as file:
        user_specified_test_shapes = json.load(file)

def get_test_shapes(test_name: str) -> list[tuple[int]]:
     if test_name in user_specified_test_shapes:
        return user_specified_test_shapes[test_name]
    return default_test_shapes

@pytest.mark.parameterize("shape", get_test_shapes("test_copy"))
```

erman-gurses: Thanks, Harsh, updated my implementation

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -7,17 +7,29 @@
...
 _run_e2e = int(os.environ.get("WAVE_RUN_E2E_TESTS", 0))
-
 require_e2e = pytest.mark.skipif(not _run_e2e, reason="e2e tests are disabled")
+default_test_shapes = [(1, 128), (256, 64), (256, 128), (256, 256), (256, 1024)]
+
+user_specified_test_shapes = ""
+
+test_params_path = os.environ.get("TEST_PARAMS_PATH", "")
+
+if os.path.isfile(test_params_path):
```
Hardcode84, line 20: Instead of checking `os.path.isfile(test_params_path)` we can do `os.environ.get("TEST_PARAMS_PATH", None)` instead and then test for `None`

erman-gurses: You mean something like this?
```
test_params_path = os.environ.get("TEST_PARAMS_PATH", None)

if test_params_path:
    with open(test_params_path, "r") as file:
        user_specified_test_shapes = json.load(file)
```

Hardcode84: yes

Hardcode84 (approved):
LGTM

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1,25 +1,35 @@
...
+import json
 
 _run_e2e = int(os.environ.get("WAVE_RUN_E2E_TESTS", 0))
-
 require_e2e = pytest.mark.skipif(not _run_e2e, reason="e2e tests are disabled")
+default_test_shapes = [(1, 128), (256, 64), (256, 128), (256, 256), (256, 1024)]
+
+user_specified_test_shapes = ""
+
+test_params_path = os.environ.get("TEST_PARAMS_PATH", None)
```
ScottTodd, line 18: (Saw this after a comment on https://github.com/iree-org/iree-turbine/pull/163#discussion_r1772383338)

Did you consider using pytest marks instead of environment variables and json files?

If all you want to enable is running different sets of tests in different workflows, pytest marks may be a better tool for the job than this configuration coming from files and environment variables.
* https://docs.pytest.org/en/stable/how-to/mark.html
* https://docs.pytest.org/en/stable/example/markers.html#mark-examples

With marks, you can define all the tests in the source file and then developers and automated workflows can choose to skip tests with a given mark (e.g. "large", "large_shapes", "benchmark", etc.).

This environment variable approach gives you more flexibility to customize the shapes without modifying the test source, which can help when running the tests from outside of this repository, but it doesn't look like that is a case you want to support at this moment. I think the ergonomics of setting an environment variable to a specific file path are quite poor for local development, especially if you plan on adding more of these files (one per test? each with its own environment variable?)

PR #122, opened 2024-09-04
Add documentation for GEMM example

harsh-nod:
This PR adds documentation describing
how to implement GEMMs in tkw and in the
process explains the different primitives
and concepts in tkw.

### DIFF
--- a/shark_turbine/kernel/wave/docs/gemm_example.md
+++ b/shark_turbine/kernel/wave/docs/gemm_example.md
@@ -1,7 +1,9 @@
 # Matrix Multiplication
 
 Let's assume we want to compute the following matrix multiplication.
-$$ C = A \times B^{T} $$
+ ```math
+ C = A \times B^{T}
+ ```
 where $A$ is a $M \times K$ matrix of type `fp16`, $B$ is a $N \times K$ matrix of type `fp16` and $C$ is a $M \times N$ matrix of type `fp32`. This is a mixed-precision transposed-B matrix multiplication operation that is common in machine learning applications.
 
 In `tkw`, the kernels are written using a wave/warp programming model. This means that when we author the kernel we write it from the perspective of the work that a single wave would do. This also means that we need to know what operations a single wave can do.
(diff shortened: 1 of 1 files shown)


shark_turbine/kernel/wave/docs/gemm_example.md
```diff
@@ -0,0 +1,153 @@
+# Matrix Multiplication
+
+Let's assume we want to compute the following matrix multiplication.
+$$ C = A \times B^{T} $$
```
Hardcode84, line 4: Formatting is broken here: https://github.com/harsh-nod/iree-turbine/blob/dox/shark_turbine/kernel/wave/docs/gemm_example.md

shark_turbine/kernel/wave/docs/gemm_example.md
```diff
@@ -0,0 +1,153 @@
...
+
+```python
+constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+```
+
+This reads as "distribute dimension `M` with a tile size of `BLOCK_M` along workgroup dimension 0". So if we have a tensor of shape `[M, N]`, then with this constraint, each workgroup would be operating on a tile of size `[BLOCK_M, N]`.
+
+Once we have a workgroup constraint in place, we add further wave constraints on the same dimension.
+
+```
```
Hardcode84, line 128: nit: missing ```python

shark_turbine/kernel/wave/docs/gemm_example.md
harsh-nod: thanks, will fix.

shark_turbine/kernel/wave/docs/gemm_example.md
harsh-nod: thanks, will add.

Hardcode84 (approved):
Overall looking good. We probably need a separate section on our `TestLaunchContext` harness and symbol binding.

PR #124, opened 2024-09-05
Enable disabling strictness

rsuderman:
For large model strictness requires loading the contents of each tensor. Disabling allows extremely large models to not require massive amounts of ram.

### DIFF
--- a/shark_turbine/aot/fx_programs.py
+++ b/shark_turbine/aot/fx_programs.py
@@ -167,7 +167,7 @@ class FxProgramsBuilder(FxPrograms):
         args=None,
         kwargs=None,
         dynamic_shapes=None,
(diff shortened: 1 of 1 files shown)


shark_turbine/aot/fx_programs.py
```diff
@@ -167,13 +167,15 @@ def export_program(
         args=None,
         kwargs=None,
```
stellaraccident, line 168: Might as well type it: `strict: bool = True`

stellaraccident (approved):

shark_turbine/aot/fx_programs.py
rsuderman: Done.

PR #127, opened 2024-09-05
Use absolute stride when computing index sequence for vector shapes

harsh-nod:
This PR modifies the linear index sequence when using vector shapes to use the absolute stride instead of the relative stride.

### DIFF
--- a/shark_turbine/kernel/wave/expansion.py
+++ b/shark_turbine/kernel/wave/expansion.py
@@ -124,6 +124,11 @@ def compute_stride(
             break
         assert dim in vector_shapes, f"Dimension {dim} not found in vector shapes"
         stride *= vector_shapes[dim]
+
+    try:
+        stride = int(stride)
+    except Exception as e:
+        logger.error(e)
     return stride
 
 
--- a/shark_turbine/kernel/wave/expansion.py
+++ b/shark_turbine/kernel/wave/expansion.py
@@ -108,6 +108,25 @@ def is_contiguous_dim(
     return is_innermost_dim or all_unit_dims
 
 
+def compute_stride(
+    symbolic_shape: tuple[IndexSymbol, ...],
+    vector_shapes: dict[IndexSymbol, int],
+    target_dim: IndexSymbol,
+) -> int:
+    """
+    Compute the stride for a given dimension based on the vector shapes.
+    The stride is the product of the vector shapes of all dimensions that are
+    not the given dimension.
+    """
+    stride = 1
+    for dim in reversed(symbolic_shape):
+        if dim == target_dim:
+            break
+        assert dim in vector_shapes, f"Dimension {dim} not found in vector shapes"
+        stride *= vector_shapes[dim]
+    return stride
+
+
 def set_node_index(
     constraints: Sequence[Constraint],
     mma_index: dict[IndexSymbol, int],
@@ -165,7 +184,7 @@ def set_node_index(
                 # The semantics of elements_per_thread are that it represents the number of
                 # elements that are loaded contiguously from memory.
                 elements_per_thread = getattr(custom, "elements_per_thread", None)
-                constraint_index, elements_per_thread = (
+                constraint_index, elements_per_thread, stride = (
                     (
                         workgroup_constraints[dim].workgroup_dim,
                         (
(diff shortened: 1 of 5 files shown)


shark_turbine/kernel/wave/expansion.py
```diff
@@ -108,6 +108,25 @@ def is_contiguous_dim(
     return is_innermost_dim or all_unit_dims
 
 
+def compute_stride(
+    symbolic_shape: tuple[IndexSymbol, ...],
+    vector_shapes: dict[IndexSymbol, int],
+    target_dim: IndexSymbol,
+) -> int:
```
Hardcode84, line 115: Is return type always intended to be resolved to `int` at this point?

shark_turbine/kernel/wave/utils.py
```diff
@@ -312,7 +312,7 @@ def safe_subs(input: Any, subs: List[Tuple[IndexExpr, IndexExpr]]) -> Any:
     Substitute input using provided `subs` list if input is sympy object.
     Otherwise return input unchanged.
     """
-    if isinstance(input, sympy.Basic):
+    if isinstance(input, sympy.Basic) or isinstance(input, IndexSequence):
```
Hardcode84, line 315: nit: you can do `isinstance(input, (sympy.Basic, IndexSequence))`

shark_turbine/kernel/wave/expansion.py
harsh-nod: Good question. Basically what you are asking is whether vector_shapes is always going to be specified as an int? Since it is used to represent the minimum granularity of the operations, I think it should always be specified as an int.

shark_turbine/kernel/wave/utils.py
harsh-nod: sure will change.

harsh-nod: thats pretty cool, didnt know you could do that :)

shark_turbine/kernel/wave/expansion.py
Hardcode84: We can do `return int(stride)` to force it (or get error).

raikonenfnu: I think error is better since because forcing things to be int may not be unexpected.

shark_turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -31,11 +35,16 @@ def partition_strided_operators(trace: CapturedTrace, constraints: list[Constrai
...
+        Checks for writes on 2d tensors with strided access on a single dimension that
+        read more than a single element.
         """
         custom = get_custom(node)
         if isinstance(custom, Write) and len(custom.type.symbolic_shape) == 2:
             strides = [simplify_index(custom.index[dim]).stride for dim in custom.index]
+            elements_per_thread = [
+                simplify_index(custom.index[dim]).size for dim in custom.index
+            ]
+            strides = [x for x, y in zip(strides, elements_per_thread) if y > 1]
```
raikonenfnu, line 47: I thought we only care for fastest dimensions no?

raikonenfnu: I guess there wouldn't/shouldn't be a case where fastest dim's elem_per_thread == 1, while the a non fastest > 1. => shouldn't matter!

shark_turbine/kernel/wave/expansion.py
harsh-nod: Okay will add a check here and throw an error if not an int.

shark_turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: So with this patch for a tensor of [M, N] the stride will be [N, 1]. Now lets say we are reading 4 elements along the N dimension. So our access patterns for N will be [offset, read 4 elements, stride 1] and our access pattern for M will be [offset, read 1 element, stride N]. Without this check, this pattern would get applied to this tensor which we don't want. So that's why its not enough to just check for strided access, we also need to know how many elements are being loaded. In cases where we are reading 4 elements from N, our access pattern for N will be [offset, read 1 element, stride 1] and for M will be [offset, read 4 elements, stride N] and then we do want this pattern to kick in.

raikonenfnu: makes sense, thanks!

raikonenfnu (approved):
LGTM!

PR #128, opened 2024-09-05
Add modulo scheduling implementation

harsh-nod:
This PR adds an implementation of modulo scheduling. This is a good baseline algorithm for scheduling
and can be used for scheduling operations in tkw.


shark_turbine/kernel/wave/scheduling/modulo_scheduling.py
```diff
@@ -19,27 +26,185 @@ class ModuloScheduler:
...
+        """
+        RT_prime = np.array(RT)
+        for i in range(node.rrt.shape[0]):
+            RT_prime[(s + i) % T] += node.rrt[i]
+        if np.all(RT_prime <= self.resources):
+            logger.debug(f"Scheduled node: {node} at time slot: {s}.")
+            self.schedule[node] = s
+            RT[:] = np.array(RT_prime)
+            return True
+
```
Hardcode84, line 207: Missing return?

shark_turbine/kernel/wave/scheduling/modulo_scheduling.py
```diff
@@ -19,27 +26,185 @@ class ModuloScheduler:
...
         self.seed = 2024
 
+    def get_edge(self, from_node: fx.Node, to_node: fx.Node) -> Edge:
+        """
+        Returns the edge between two nodes.
+        """
+        for edge in self.edges:
+            if edge._from == from_node and edge._to == to_node:
+                return edge
+        return Edge(None, None, EdgeWeight(0, 0))
```
Hardcode84, line 44: Maybe better to return `None` instead and check on the caller side?

shark_turbine/kernel/wave/scheduling/modulo_scheduling.py
harsh-nod: good catch! Thanks will add a return here.

shark_turbine/kernel/wave/scheduling/modulo_scheduling.py
harsh-nod: sure will change.

Hardcode84 (approved):
LGTM. I didn't looked too closely on the algorithm implementation, but let's merge it to move forward.

PR #129, opened 2024-09-06
Add separate function to apply shared memory indexing

harsh-nod:
This PR introduces a new function to apply shared
memory indexing corrections. Previously, this was being done as part of post expansion. With the new PR, we look at the entire graph holistically and apply indexing changes to reads from shared memory.

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -348,7 +348,7 @@ def test_mma():
         # CHECK:            %[[D5:.+]] = vector.load %[[D0]][%[[D4]], %[[C0]]] : memref<64x16xf16, strided<[16, 1], offset: ?>>,
         # CHECK-SAME:         vector<4xf16>
         # CHECK:            %[[ALLOC:.+]] = memref.alloc() : memref<32x16xf16, #[[GPU:.+]].address_space<workgroup>>
-        # CHECK:            vector.store %[[D5]], %[[ALLOC]][%[[D4]], %[[C0]]] : memref<32x16xf16,
+        # CHECK:            vector.store %[[D5]], %[[ALLOC]][%[[D2]], %[[C0]]] : memref<32x16xf16,
         # CHECK-SAME:         #[[GPU]].address_space<workgroup>>, vector<4xf16>
         # CHECK:            amdgpu.lds_barrier
         # CHECK:            %[[D6:.+]] = arith.remsi %[[THREAD_ID_X]], %[[C16]] : index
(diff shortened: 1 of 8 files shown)


lit_tests/kernel/wave/codegen.py
```diff
@@ -348,7 +348,7 @@ def mma(
         # CHECK:            %[[D5:.+]] = vector.load %[[D0]][%[[D4]], %[[C0]]] : memref<64x16xf16, strided<[16, 1], offset: ?>>,
         # CHECK-SAME:         vector<4xf16>
         # CHECK:            %[[ALLOC:.+]] = memref.alloc() : memref<32x16xf16, #[[GPU:.+]].address_space<workgroup>>
-        # CHECK:            vector.store %[[D5]], %[[ALLOC]][%[[D4]], %[[C0]]] : memref<32x16xf16,
+        # CHECK:            vector.store %[[D5]], %[[ALLOC]][%[[D2]], %[[C0]]] : memref<32x16xf16,
```
raikonenfnu, line 351: Just curious, why would there be a change in LIT test? I feel like the refactoring just cleans it up/change the location of this transformation?

raikonenfnu (commented):
Overall looks good/ really nice cleanup! just 1 Q.

lit_tests/kernel/wave/codegen.py
harsh-nod: So this was a bug. The shared memory store indices should not have any workgroup dependence but they did.

shark_turbine/kernel/wave/minimize_global_loads.py
```diff
@@ -140,9 +141,7 @@ def add_optimized_nodes(
                         write = Write(
                             read, custom_user.memory, load_elems_per_thread
                         ).add_to_graph(custom.graph)
-                        write.index = remove_global_indexing(
-                            read.index, tilingConstraints
-                        )
+                        write.index = deepcopy(read.index)
```
Hardcode84, line 144: Why do we need `deepcopy` now?

lit_tests/kernel/wave/codegen.py
raikonenfnu: Makes sense :)

shark_turbine/kernel/wave/minimize_global_loads.py
harsh-nod: So if we just assign read.index to write.index, then when we modify write.index in the corrections pass it will also affect read.index. But did you have a better way to handle this?

Hardcode84: But `remove_global_indexing` is already creating a copy, it shouldn't affect original index object.

harsh-nod: Yes, that's right. `remove_global_indexing` is creating a copy, but that copy will be assigned to write.index and read.index which is not what we want. If you remove the `deepcopy`, two of the lit tests fail.

harsh-nod: okay I was able to figure this out. I have removed the deep copy. @Hardcode84 - can you take a look when you get a chance? thanks!

Hardcode84 (approved):
LGTM!

PR #131, opened 2024-09-10
Various fixes and cleanups, needed for igemm

Hardcode84:
* Use proper `memory_type`/`register_type` in various places as they may be different for reads/writes with mapping
* Support for `x + (a/b)` sympy expr in lowering.
* Rework gather/scatter offsets generation, handle the case when we need a mix of dynamic and static offsets, generate static offsets vector first and then `insertelement` remaining dynamic offsets into it.
* Generate plain `vector.store` instead of `vector.scatter` if `elements_per_thread` == 1 as 0-th scatter offset should always be 0.
* Propagate mapping, when generating new Read/Write nodes
* Use `subs_idxc`


shark_turbine/kernel/wave/codegen.py
```diff
@@ -213,14 +214,24 @@ def gen_sympy_index(emitter: WaveEmitter, expr: sympy.Expr) -> OpResult:
...
-                        operation = arg(operation)
+                        operation = arg(operation, True)
                 stack.append(operation)
             case sympy.Add():
-                summand = stack.pop()
-                add = summand
-                for _ in range(1, len(term.args)):
-                    add = arith_d.AddIOp(add, stack.pop())
-                stack.append(add)
+                args = []
```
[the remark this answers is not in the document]
harsh-nod: Also, I think the code would become simpler if you moved the callable(s) to the end of args (in the first loop over term.args) so you don't need the second for loop over args.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -231,6 +242,8 @@ def gen_sympy_index(emitter: WaveEmitter, expr: sympy.Expr) -> OpResult:
                 # But check whether floordivsi is needed.
                 stack.append(stack.pop())
             case sympy.Rational():
+                # `x * (a/b)` transformed into `(x * a) / b`
+                # `x + (a/b)` transformed into `(x*b + a) / b`
```
harsh-nod, line 246: Why do we need to transform x + (a/b) - > (x*b + a) / b? Can't we do (a/b) and then add x? Just trying to avoid an extra multiplication.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -375,31 +399,51 @@ def _construct_gather_scatter_indices(
...
+    dynamic_offsets = []
     for i in range(elements_per_thread):
         # Update most-minor dim, i.e. in case of identity mapping it will
         # be equivalent to just vector.load
-        subs[-1] = (subs[-1][0], i)
-        indices = [int(i.subs(subs)) for i in index_mapping]
-        offsets.append(
-            IntegerAttr.get(IndexType.get(), _compute_offset(indices, strides))
+        subs = [(sym, idx) for sym, idx in zip(iters.keys(), start_indices_orig)]
+        subs[-1] = (subs[-1][0], start_indices_orig[-1] + i)
```
harsh-nod, line 420: Why are the start indices being added to i? Can you provide a comment explaining this?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -375,31 +399,51 @@ def _construct_gather_scatter_indices(
...
+        indices = [i.subs(subs) for i in index_mapping]
```
harsh-nod, line 421: Why do we not need to cast to int anymore?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -375,31 +399,51 @@ def _construct_gather_scatter_indices(
...
+        offset = _compute_offset(indices, strides) - _compute_offset(
+            start_indices, strides
         )
+        offset = subs_idxc(offset)
+
+        try:
```
harsh-nod, line 427: Instead of the try-except, can we add specific checks that we are looking for? I'm guessing if it is a sympy expression then we dont want to cast to int only when its a sympy integer?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -495,7 +539,10 @@ def handle_write(emitter: WaveEmitter, node: fx.Node):
             is_read=False,
         )
 
-        vector_d.scatter(kb_dest, start_indices, offsets_vec, mask, insert_vector)
+        if elements_per_thread == 1:
```
harsh-nod, line 542: Does this mean that we cannot codegen contiguous writes of more than 1 element? If so, seems very restrictive.

shark_turbine/kernel/wave/promotion.py
```diff
@@ -42,7 +42,7 @@ def promote_node(
     with node.graph.inserting_before(node.fx_node.next):
         constrained_shape = get_constrained_shape(node.type.symbolic_shape, constraints)
         allocate_node = Allocate(
-            node.memory_type.symbolic_shape,
+            node.type.symbolic_shape,
```
harsh-nod, line 45: Shouldn't this be memory type?

harsh-nod (changes requested):

shark_turbine/kernel/wave/codegen.py
Hardcode84: After all the sympy expansions I ended with expressions like `floor(x + 3/4)` which will be rounded to `x` if we do division first. I'm still not sure if any of this correct, but I followed the existing `mul` approach.

shark_turbine/kernel/wave/codegen.py
Hardcode84: Contiguous writes are covered in separate branch above, added this case because `partition_strided_operators` tends to generate a bunch of single-element writes and we don't need `scatter` for 1-element writes a 0th scatter index is always 0.

shark_turbine/kernel/wave/promotion.py
Hardcode84: For reads with mapping op sequence will look like this:
`global_mem -> gather -> write -> shared_mem alloc -> read -> mma`

`memory_type.symbolic_shape` will return shape of the gather input and `type.symbolic_shape` will be shape of gather output, what we really want here.

shark_turbine/kernel/wave/codegen.py
Hardcode84: Yeah, I will add some comments explaining the algorithm

shark_turbine/kernel/wave/codegen.py
Hardcode84: Is there any 'canonical' way of checking if sympy expression convertible to int constant?

shark_turbine/kernel/wave/codegen.py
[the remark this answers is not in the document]
Hardcode84: done

shark_turbine/kernel/wave/codegen.py
Hardcode84: added some comments

shark_turbine/kernel/wave/codegen.py
```diff
@@ -158,6 +159,31 @@ def get_type_or_element_type(operand_type: IrType):
 def gen_sympy_index(emitter: WaveEmitter, expr: sympy.Expr) -> OpResult:
     stack: list[OpResult] = []
 
+    def _process_mul_add_ops(term, is_mul):
```
harsh-nod, line 162: nice! looks good!

shark_turbine/kernel/wave/codegen.py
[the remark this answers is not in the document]
harsh-nod: thanks!

shark_turbine/kernel/wave/codegen.py
harsh-nod: I see. So a and b are integers and / corresponds to signed integer division, so the only way to ensure the correct result is to do the multiplication first (and addition) and then the division. I think this is a reasonable approach to take considering the alternative which would involve casting a, b to floats and doing all the arithmetic on floats.

shark_turbine/kernel/wave/codegen.py
harsh-nod: thanks, helps clarify whats going on.

shark_turbine/kernel/wave/codegen.py
harsh-nod: I believe you can use is_number, https://docs.sympy.org/latest/modules/core.html#sympy.core.expr.Expr.is_number

shark_turbine/kernel/wave/codegen.py
harsh-nod: makes sense, thanks for clarifying.

shark_turbine/kernel/wave/promotion.py
harsh-nod: What about without mapping op? Then the sequence would be
`global_mem -> write -> shared_mem alloc -> read -> mma`

What would `memory_type.symbolic_shape` and `type.symbolic_shape` return in this case?

shark_turbine/kernel/wave/codegen.py
Hardcode84: done

shark_turbine/kernel/wave/codegen.py
Hardcode84: Moved cast to the later time, after all indices calculations are applied.

shark_turbine/kernel/wave/promotion.py
harsh-nod: sounds good, thanks.

PR #132, opened 2024-09-10
Use `assert_allclose` in gemm test

Hardcode84:
`assert_allclose` will actually show you how many elements mismatched and the relative/absolute difference.

### DIFF
--- a/tests/kernel/wave/wave_gemm_test.py
+++ b/tests/kernel/wave/wave_gemm_test.py
@@ -8,6 +8,7 @@ import shark_turbine.kernel.wave as tkw
 from shark_turbine.kernel.lang.global_symbols import *
 from shark_turbine.kernel.wave.iree_utils import generate_iree_ref
 import os
+from numpy.testing import assert_allclose
 
 _run_e2e = int(os.environ.get("WAVE_RUN_E2E_TESTS", 0))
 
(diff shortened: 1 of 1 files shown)


tests/kernel/wave/wave_gemm_test.py
```diff
@@ -93,7 +94,7 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
             gemm(a, b, c)
             iree_ref = torch.zeros(2048, 10240, dtype=torch.float32)
             generate_iree_ref("mmt", [a, b], [iree_ref], config)
-            assert torch.equal(c, iree_ref)
+            assert_allclose(c.numpy(), iree_ref.numpy())
```
harsh-nod, line 97: Can you check to see if torch.assert_allclose works the same? Then we dont have to convert to numpy?

harsh-nod (approved):
lgtm!

tests/kernel/wave/wave_gemm_test.py
Hardcode84: `torch.assert_allclose` is deprecated, switched to `torch.assert_close`, error message now looks like:
```
E           AssertionError: Tensor-likes are not close!
E
E           Mismatched elements: 20971517 / 20971520 (100.0%)
E           Greatest absolute difference: 1413.0994873046875 at index (1486, 268) (up to 1e-05 allowed)
E           Greatest relative difference: 63904244.0 at index (243, 8383) (up to 1.3e-06 allowed)
```

PR #136, opened 2024-09-11
Add another mapping e2e test

Hardcode84:
This is essentially an igemm conv, but without final col2im transform, it reads input and weights with mapping, but writes result without any mapping.


Hardcode84:
Closing in favor of https://github.com/iree-org/iree-turbine/pull/137

PR #144, opened 2024-09-16
[TKW] Masking

Hardcode84:
Add support for cases when input sizes are not divisible by WG sizes, use masked vector instructions int his case, to only access elements within arrays bounds.

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -197,6 +197,88 @@ def test_read_write():
 
 
 @run_test
+def test_read_write_masked():
+    constraints: list[tkw.Constraint] = [
+        tkw.HardwareConstraint(
+            threads_per_wave=64, waves_per_block=(1, 1, 1), vector_shapes={M: 4, N: 4}
+        )
+    ]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
+    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+
+    @tkw.wave(constraints)
+    def test(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+        b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+    ):
+        res = tkw.read(a, elements_per_thread=4)
+        tkw.write(res, b, elements_per_thread=4)
+
+    with tk.gen.TestLaunchContext(
+        {
+            M: 1,
+            N: 3,
+            BLOCK_M: 4,
+            BLOCK_N: 4,
+            ADDRESS_SPACE: tkl.AddressSpace.SHARED_MEMORY.value,
+        },
+        canonicalize=True,
+    ):
+        a = torch.randn(4, 4, dtype=torch.float16)
+        b = torch.zeros(4, 4, dtype=torch.float16)
+        print(test(a, b).module_op)
+
+        # CHECK:          func.func @test(%[[ARG0:[a-zA-Z0-9_]+]]: !stream.binding, %[[ARG1:[a-zA-Z0-9_]+]]: !stream.binding)
+        # CHECK-DAG:        %[[CST:.+]] = arith.constant dense<0.000000e+00> : vector<4xf16>
+        # CHECK-DAG:        %[[C0:.+]] = arith.constant 0 : index
+        # CHECK-DAG:        %[[C1:.+]] = arith.constant 1 : index
+        # CHECK-DAG:        %[[C2:.+]] = arith.constant 2 : index
+        # CHECK-DAG:        %[[C3:.+]] = arith.constant 3 : index
+        # CHECK-DAG:        %[[C4:.+]] = arith.constant 4 : index
+        # CHECK-DAG:        %[[C8:.+]] = arith.constant 8 : index
+        # CHECK-DAG:        %[[C64:.+]] = arith.constant 64 : index
+        # CHECK:            %[[WORKGROUP_ID_0:.+]] = stream.dispatch.workgroup.id[0] : index
+        # CHECK:            %[[WORKGROUP_ID_1:.+]] = stream.dispatch.workgroup.id[1] : index
+        # CHECK-DAG:        %[[THREAD_ID_X:.+]] = gpu.thread_id  x
+        # CHECK-DAG:        %[[THREAD_ID_Y:.+]] = gpu.thread_id  y
+        # CHECK:            %[[D0:.+]] = stream.binding.subspan %[[ARG0]][%[[C0]]] : !stream.binding ->
+        # CHECK-SAME:         memref<1x3xf16, strided<[3, 1], offset: ?>>
+        # CHECK:            %[[D1:.+]] = arith.muli %[[WORKGROUP_ID_0]], %[[C4]] : index
+        # CHECK:            %[[D2:.+]] = arith.divsi %[[THREAD_ID_X]], %[[C64]] : index
+        # CHECK:            %[[D3:.+]] = arith.muli %[[D2]], %[[C4]] : index
+        # CHECK:            %[[D4:.+]] = arith.addi %[[D3]], %[[D1]] : index
+        # CHECK:            %[[D5:.+]] = arith.addi %[[D4]], %[[THREAD_ID_X]] : index
+        # CHECK:            %[[D6:.+]] = arith.muli %[[WORKGROUP_ID_1]], %[[C4]] : index
+        # CHECK:            %[[D7:.+]] = arith.muli %[[THREAD_ID_Y]], %[[C8]] : index
+        # CHECK:            %[[D8:.+]] = arith.addi %[[D7]], %[[D6]] : index
+        # CHECK:            %[[D9:.+]] = vector.constant_mask [4] : vector<4xi1>
+        # CHECK:            %[[D10:.+]] = arith.cmpi slt, %[[D5]], %[[C1]] : index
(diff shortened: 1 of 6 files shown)


Hardcode84:
e2e tests now all pass locally with weird sizes, so it's ready for review. Bu I'll still wait until e2e fixed in main before merging.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -543,11 +551,14 @@ def repeat(acc: tkl.Register[M, NF, tkl.f32]) -> tkl.Register[M, NF, tkl.f32]:
...
-def test_igemm_conv():
-    n, c, h, w = 1, 4, 5, 5  # Image.
-    nf, cf, hf, wf = 16, c, 2, 2  # Filters.
+@pytest.mark.parametrize("n", [1, 2, 4])
+@pytest.mark.parametrize("c", [1, 3, 4, 10])
+@pytest.mark.parametrize("nf", [1, 2, 16])
+@pytest.mark.parametrize("stride", [1, 2, 3])
+def test_igemm_conv(n, c, nf, stride):
+    h, w = 5, 5  # Image.
+    cf, hf, wf = c, 2, 2  # Filters.
```
harsh-nod, line 560: Not for this patch, but would it work if you parameterized h, w, hf and wf?

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -13,7 +13,15 @@
 
 _run_e2e = int(os.environ.get("WAVE_RUN_E2E_TESTS", 0))
 require_e2e = pytest.mark.skipif(not _run_e2e, reason="e2e tests are disabled")
-default_test_shapes = [(1, 128), (256, 64), (256, 128), (256, 256), (256, 1024)]
+default_test_shapes = [
```
harsh-nod, line 16: nice!

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -43,7 +51,7 @@ def test_copy(shape):
     # elements.
     wave_size = 64
     BLOCK_M = 1
-    BLOCK_N = sympy.Min(N, 256)
+    BLOCK_N = sympy.Max(sympy.Min(N, 256), 64)
```
harsh-nod, line 54: What happens if you remove the max and leave it to min(N, 256)? Just wondering why the max is necessary.

shark_turbine/kernel/wave/constraints.py
```diff
@@ -170,6 +170,13 @@ class WorkgroupConstraint(Constraint):
     tile_size: IndexExpr
     workgroup_dim: int
 
+    @property
+    def iterations(self) -> IndexExpr:
```
harsh-nod, line 174: since this is the total number of workgroups, should we just call it that? num_workgroups() ? (iterations is a little misleading since usually it is used in the context of a for-loop which in our case would be the tkw.reduction / TilingConstraint)

shark_turbine/kernel/wave/codegen.py
```diff
@@ -830,17 +907,17 @@ def handle_reduction(emitter: WaveEmitter, node: fx.Node):
...
+            iterations = subs_idxc(constraint.iterations)
+    assert (
+        iterations is not None
+    ), "Could not find tiling constraint for reduction axis."
 
     # For now, we assume that dimensions that have tiling constraints on them,
     # do not have any other constraints.
-    dim = axis.subs(idxc.subs)
-    end = arith_d.constant(IndexType.get(), int(dim // tile_size))
+    end = arith_d.constant(IndexType.get(), int(iterations))
```
harsh-nod, line 920: Not for this PR, but wondering whether we should have subs_idxc return an int when we know the return type can be converted to an int? This way subs_idxc and safe_subs are guaranteed to give either floats/ints or symbolic expressions.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -372,6 +383,51 @@ def _is_identity_mapping(
     return True
 
 
+def _build_mask(
```
harsh-nod, line 386: Can you add a lit test showing what the IR looks like?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -372,6 +383,51 @@ def _is_identity_mapping(
...
+
+        dim = constraint.dim
+        if dim not in index:
+            continue
+
+        work_size = constraint.iterations * constraint.tile_size
+        if subs_idxc(work_size) == subs_idxc(dim):
+            continue
+
+        bounds.append((dim, gen_sympy_index(emitter, dim)))
```
harsh-nod, line 402: Doesn't get_sympy_index(dim) = dim ? So why do we need this?

shark_turbine/kernel/wave/codegen.py
harsh-nod: Also helps for me to debug and step through locally without having access to the cluster.

harsh-nod (changes requested):
Could you add a lit test that shows the IR? I will be able to follow better if I can step through the code and understand what's happening. Thanks!

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -671,4 +682,4 @@ def repeat(acc: tkl.Register[M, NF, tkl.f32]) -> tkl.Register[M, NF, tkl.f32]:
         run_config=config,
     ):
         conv(x, we, out)
-        assert_allclose(out, out_ref, rtol=1e-05, atol=1e-05)
+        assert_allclose(out, out_ref, rtol=1e-03, atol=1e-03)
```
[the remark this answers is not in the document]
Hardcode84: As I'm testing 3*4*3*3 total permutations now, it's infeasible to find good seed for each one and I'm getting occasional mismatches on `1e-05`.

tests/kernel/wave/wave_e2e_test.py
Hardcode84: We should, but I need to find a better way of deciding which combination of parameters we are going to test, as currently amount of test permutations grow multiplicative.

shark_turbine/kernel/wave/constraints.py
Hardcode84: I choose `iterations` specifically, because it's used interchangeably between `WorkgroupConstraint` and `TilingConstraint` in bounds check so I don't need to check for them separately. Maybe any ideas for better common name?

tests/kernel/wave/wave_e2e_test.py
Hardcode84: I want to keep block size no smaller than `wave_size`, but I should probably change it  `sympy.Max(sympy.Min(N, 256), wave_size)` to make it clear

Hardcode84: updated

shark_turbine/kernel/wave/codegen.py
Hardcode84: done

tests/kernel/wave/wave_e2e_test.py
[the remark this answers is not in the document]
harsh-nod: Okay makes sense. Thanks!

tests/kernel/wave/wave_e2e_test.py
harsh-nod: Makes sense. Just wanted to know if it was working and sounds like it is which is great. We can parameterize it for the nightly test.

tests/kernel/wave/wave_e2e_test.py
harsh-nod: thanks

shark_turbine/kernel/wave/constraints.py
harsh-nod: How about counts? Specifies number of loop counts or workgroup counts?

shark_turbine/kernel/wave/codegen.py
harsh-nod: thanks!

shark_turbine/kernel/wave/codegen.py
harsh-nod: that explains it, yeah agree a better name would help.

lit_tests/kernel/wave/codegen.py
```diff
@@ -196,6 +196,88 @@ def test(
...
+        # CHECK-SAME:         memref<1x3xf16, strided<[3, 1], offset: ?>>
+        # CHECK:            %[[D1:.+]] = arith.muli %[[WORKGROUP_ID_0]], %[[C4]] : index
+        # CHECK:            %[[D2:.+]] = arith.divsi %[[THREAD_ID_X]], %[[C64]] : index
+        # CHECK:            %[[D3:.+]] = arith.muli %[[D2]], %[[C4]] : index
+        # CHECK:            %[[D4:.+]] = arith.addi %[[D3]], %[[D1]] : index
+        # CHECK:            %[[D5:.+]] = arith.addi %[[D4]], %[[THREAD_ID_X]] : index
+        # CHECK:            %[[D6:.+]] = arith.muli %[[WORKGROUP_ID_1]], %[[C4]] : index
+        # CHECK:            %[[D7:.+]] = arith.muli %[[THREAD_ID_Y]], %[[C8]] : index
+        # CHECK:            %[[D8:.+]] = arith.addi %[[D7]], %[[D6]] : index
+        # CHECK:            %[[D9:.+]] = vector.constant_mask [4] : vector<4xi1>
```
harsh-nod, line 256: This is exactly what I was looking for - thank you. So this brings me to my main question, why do we need to do all this arithmetic to compute the mask? Since we have the information from sympy in the python layer, can we figure out what the mask should be before codegen? The advantage of that is we avoid all these extra instructions.

harsh-nod: If the shape was dynamic, then I think we would have to do this, but if it was static I think we could avoid this?

harsh-nod (approved):
lgtm, modulo changes we discussed!

PR #150, opened 2024-09-17
[TKW] Add register analysis support for ReduceOp

raikonenfnu:
Implement pieces to enable partial reductions/tiling of ReduceOps(max/sum) in a for loop/tkw.reduction loop. 

1.  Add register analysis support for reduce op.
2. Updated implementation of `replace_all_uses_with` with FX's implementation of it. 

num.2 is required to support updating the use of result from `ReduceOp` on `Output` from `Reduction`. In this case, with existing implementation of `update_arg(user.get_node_arg_index(` where `user` is `Output` causes issues because `output.fx_node.arg` is a `tuple[list[fx.node]]`, we can update it to not use fx's implementation but may get quite invasive and ad-hoc.


harsh-nod:
@raikonenfnu - I cherry-picked your change and ran it on the MI300.
```
git fetch origin pull/150/head:stan
git checkout stan
```
With my top commit being, 
```
commit cedf2d081f9d76a146a0eb62b61bd29870bb4efa (HEAD -> stan)
Author: Stanley Winata <stanley.winata@amd.com>
Date:   Tue Sep 17 23:02:03 2024 -0700

    Add test path

    Signed-off-by: Stanley Winata <stanley.winata@amd.com>
```
But when I run the e2e test,
```
pytest -s tests/kernel/wave/wave_e2e_test.py
```
it fails with
```
FAILED tests/kernel/wave/wave_e2e_test.py::test_tiled_reduce_max[shape1] - ValueError: Tile size must be divisible by wave count and vector size
```
Are you seeing this on your end? Not sure how the builder is passing, but if it also works for you then must be something wrong with my setup.

lit_tests/kernel/wave/codegen.py
```diff
@@ -691,7 +691,8 @@ def test_reduce_max():
         )
     ]
     constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
-    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, N, 0)]
+    constraints += [tkw.TilingConstraint(N, BLOCK_N)]
```
harsh-nod, line 695: Does this work if you remove 
```
constraints += [tkw.WorkgroupConstraint(N, N, 0)]
```
? Or is this required to get the workgroup_id.x ?

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -51,5 +87,10 @@ def determine_register_shape(trace: CapturedTrace | fx.Graph) -> None:
     register_nodes = trace.walk(lambda node: isinstance(get_custom(node), NewRegister))
     if not register_nodes:
         return
+    vector_map = {}
+    for c in constraints:
```
harsh-nod, line 91: This seems like a useful piece of code. Do you think you can put it as a function in utils?

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -33,16 +42,43 @@ def set_register_shape(trace: CapturedTrace, custom: CustomOp) -> None:
...
-            set_register_shape(trace, iter_arg)
+            set_register_shape(trace, iter_arg, vector_map)
             custom.fx_node.thread_shape = iter_arg.fx_node.thread_shape
             break
+        elif isinstance(custom_user, ReduceOp):
+            get_thread_shape = lambda reg_shape: max(reg_shape)
+            # Check that dim is non-reduction && in hw_constraint.vector_shape.
+            is_valid_dim = lambda dim: dim != custom_user.dim and dim in vector_map
+            # TODO: Modify num_reduction_dims once we add support for multi-dim reduction.
+            num_reduction_dims = 1
```
harsh-nod, line 53: Can you query this from the ReduceOp and assert if the number of reduction dims is greater than 1?

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -33,16 +42,43 @@ def set_register_shape(trace: CapturedTrace, custom: CustomOp) -> None:
...
                 ]
             )
-            set_register_shape(trace, iter_arg)
+            set_register_shape(trace, iter_arg, vector_map)
             custom.fx_node.thread_shape = iter_arg.fx_node.thread_shape
             break
+        elif isinstance(custom_user, ReduceOp):
+            get_thread_shape = lambda reg_shape: max(reg_shape)
+            # Check that dim is non-reduction && in hw_constraint.vector_shape.
+            is_valid_dim = lambda dim: dim != custom_user.dim and dim in vector_map
```
harsh-nod, line 51: If vector_map is None (if no vector shapes are specified), will this fail?

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -33,16 +42,43 @@ def set_register_shape(trace: CapturedTrace, custom: CustomOp) -> None:
                     idx
                 ]
             )
-            set_register_shape(trace, iter_arg)
+            set_register_shape(trace, iter_arg, vector_map)
             custom.fx_node.thread_shape = iter_arg.fx_node.thread_shape
             break
+        elif isinstance(custom_user, ReduceOp):
+            get_thread_shape = lambda reg_shape: max(reg_shape)
```
harsh-nod, line 49: Do you need this lambda? Seems like it is just computing the max?

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -33,16 +42,43 @@ def set_register_shape(trace: CapturedTrace, custom: CustomOp) -> None:
...
+            # Check that dim is non-reduction && in hw_constraint.vector_shape.
```
harsh-nod, line 50: Also, could you change this to is_parallel_dim which is what I believe you are looking for? (parallel dims)

shark_turbine/kernel/wave/register_analysis.py
```diff
@@ -33,16 +42,43 @@ def set_register_shape(trace: CapturedTrace, custom: CustomOp) -> None:
...
+            ]
+            expected_result_rank = (
+                len(custom_user.type.symbolic_shape) - num_reduction_dims
+            )
+            # If rank do not match => some dims not found in hw_constraint.vector_shape.
+            if len(register_shape) != expected_result_rank:
+                raise NotImplementedError(
+                    "NYI: Handling of dim not in vector_shapes during register analysis."
+                )
+            num_thread_dims = sum(1 for dim in register_shape if dim > 1)
```
harsh-nod, line 67: Is this counting the number of non unit dims? If so can we call it that?

harsh-nod (changes requested):
Mostly looks good, just some minor comments.

lit_tests/kernel/wave/codegen.py
raikonenfnu: we need it to get workgroup_id.x yes

shark_turbine/kernel/wave/register_analysis.py
raikonenfnu: Done! Initially, I wrapped it around lambda to be more verbose on what we are trying to do with the max

shark_turbine/kernel/wave/register_analysis.py
raikonenfnu: Done! Initially, I wrapped it around lambda to be more verbose on what we are trying to do with the max

shark_turbine/kernel/wave/register_analysis.py
raikonenfnu: Done!

shark_turbine/kernel/wave/register_analysis.py
raikonenfnu: Good idea, done!

shark_turbine/kernel/wave/register_analysis.py
raikonenfnu: Done!

shark_turbine/kernel/wave/register_analysis.py
raikonenfnu: SG! Done!

.github/workflows/ci.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
       if: "contains(matrix.os, 'mi300') && !cancelled()"
       run: |
         export WAVE_RUN_E2E_TESTS=1
+        export TEST_PARAMS_PATH=./tests/kernel/wave/test_param.json
```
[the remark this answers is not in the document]
raikonenfnu: np! :)

shark_turbine/kernel/wave/register_analysis.py
raikonenfnu: If vector_map is None, it will fail in the assertion on below checking for that rank(thread_shape) = rank(symbolic_shape) - num_reduction_dims.

raikonenfnu: do you want a earlier/more explicit assert for vector_map != None?

shark_turbine/kernel/wave/register_analysis.py
harsh-nod: thanks!

shark_turbine/kernel/wave/register_analysis.py
harsh-nod: No just wanted to make sure we would catch the failure case which sounds like it does, thanks!

harsh-nod (approved):
lgtm!

PR #156, opened 2024-09-20
Reorder the nodes in the graph to match the dim expansion order

harsh-nod:
This PR modifies the order in which nodes are inserted into the graph to follow the canonical Cartesian product. This is especially important for the iter args inside a for loop that (without this PR) are not in the same order as the init args and outputs. By reordering the nodes, we ensure that the init args, iter args and outputs are all in the same order and map 1-1.


harsh-nod:
No longer needed.

PR #159, opened 2024-09-21
Simplify barrier insertion strategy

harsh-nod:
We move to an extremely simple barrier insertion
strategy as a baseline. While suboptimal, this
should provide a good reference when transitioning to more advance barrier insertion approaches.

### DIFF
--- a/shark_turbine/kernel/wave/barriers.py
+++ b/shark_turbine/kernel/wave/barriers.py
@@ -5,48 +5,41 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from .._support.tracing import CapturedTrace
-from ..ops.wave_ops import get_custom, Read, SharedMemoryBarrier
+from ..ops.wave_ops import get_custom, Read, SharedMemoryBarrier, Write, Reduction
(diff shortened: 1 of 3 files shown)


shark_turbine/kernel/wave/barriers.py
```diff
@@ -5,48 +5,41 @@
...
+            and custom.memory_type.address_space == SHARED_ADDRESS_SPACE
+        ):
+            if last_node is None:
+                last_node = custom
+                continue
+            if type(custom) != type(last_node):
+                with graph.inserting_before(node):
+                    SharedMemoryBarrier().add_to_graph(graph)
+            last_node = custom
+        if isinstance(custom, Reduction):
```
Hardcode84, line 42: So, if we have read as the last op in reduction and write immediately after the reduction (or vice-versa), should we insert barrier after the reduction? current implementation won't detect such case.

harsh-nod: Good catch! So I think that case will be handled if I return the value of last_node from the reduction? I will add that to the PR.

Hardcode84 (approved):

PR #161, opened 2024-09-23
Reorder iter args to match ordering of init args and outputs

harsh-nod:
This PR modifies the insertion point for iter args to ensure that the iter args are in the same order as the init args and outputs. This simplifies the mapping between init args, iter args and outputs.


raikonenfnu (approved):
Code looks great, and verified locally on my branch that it unblocks me! :)

PR #163, opened 2024-09-24
[TKW] Add xfail decorator for unaligned shape

raikonenfnu:


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -30,8 +30,10 @@
 ]
 
 user_specified_test_shapes = ""
-
-test_params_path = os.environ.get("TEST_PARAMS_PATH", None)
+default_filename = "test_param.json"
```
Hardcode84, line 33: Please, don't. We need to make all tests work with default shapes.

The core issue that `test_tiled_reduce_max` doesn't work with weird shapes, instead of skipping this shape entirely, we must fix test or mark it xfail.

Hardcode84 (changes requested):

tests/kernel/wave/wave_e2e_test.py
raikonenfnu: Interesting, so what was the motivation for adding test_param.json shapes to begin with?

Hardcode84: The motivation was to run perf test with different set of shapes from CI tests.

raikonenfnu: ah, that makes sense, thanks :)

ScottTodd: I left a comment on https://github.com/iree-org/iree-turbine/pull/119#discussion_r1773607105 . I think pytest marks (for test type / size, as well as xfail) are a better fit here than environment variables and separate json files.

raikonenfnu: @Hardcode84 updated the test ,re-review please :)

Hardcode84 (approved):

PR #164, opened 2024-09-24
Enable benchmarking in performance ci

harsh-nod:
This PR creates a benchmark script for wave sdpa. It takes in a json file and benchmarks it using rpdtrace.

The command is 
```
python iree/turbine/kernel/wave/perf/benchmark.py --config iree/turbine/kernel/wave/perf/benchmark_configs.json --num_warmup 0 --num_iterations 1
```

### DIFF
--- a/iree/turbine/kernel/wave/perf/benchmark.py
+++ b/iree/turbine/kernel/wave/perf/benchmark.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python3
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import argparse
+import json
+import time
+import torch
+import iree.turbine.kernel.wave.nn as wave_nn
+
+try:
+    from rpdTracerControl import rpdTracerControl
+    import sqlite3
+    import pandas as pd
+except ImportError:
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/wave/perf/benchmark.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python3
```
xintin, line 1: Do we need to tell the python path explicitly?

iree/turbine/kernel/wave/perf/benchmark.py
```diff
@@ -0,0 +1,137 @@
...
+    import sqlite3
+    import pandas as pd
+except ImportError:
+    print("rpdTraceControl not found, skipping profiling")
+    exit(1)
+
+DEFAULT_OUTPUT_FILENAME = "trace.rpd"
+
+
+def benchmark_wave_sdpa(
```
xintin, line 25: Shall we add type hints for args?

xintin:
LGTM! Left minor comments.

iree/turbine/kernel/wave/perf/benchmark.py
harsh-nod: I can remove this.

iree/turbine/kernel/wave/perf/benchmark.py
harsh-nod: Yes, I can add that.

xintin (approved):

PR #166, opened 2024-09-25
 [TKW] Implement support for multiple iter args on Reduction

raikonenfnu:
The main motivation behind this PR is to enable multiple induction variable/iterArg on the same tiled "Reduction" loop. To enable above we did a couple things:

1. Enable lowering/expansion on `operator.getitem` (the op that extract multiple results in python i.e `res0, res1 = fn`) by templating it on`GetResult(CustomOp)` since they have the same args and interface and can reuse most of the indexing/expansion helper.

2. Introduce `res_idx`, a variable to represent which result index of an op we are referring to, during expansion and context map. This is useful for ops that has more than one results / variables as outputs.

3. bug fix in expand_reduction, where we hoist out iterating and expanding of `reduction.init_args` out of the loop that iterates and expands over the `yield`/`return_val` of the reduction loop. It is expected that the size of `init_args`  is the same as size of `yield`/`return_val`. Hence if we had N iter_args/yields, we ended up expanding the `init_args` N x N time instead of N times. We haven't seen it thus far because we have been only playing with 1 init_arg/iterArg, and 1x1 == 1. 

4. Introduce a canonicalization pattern to fold chains of GetResult. this is because GetResult by semantic/design is only expected to extract and have one result. Hence a chain of GetResult should just be replaced by itself. This help clean up the IR. 

num.4 also helps circumvent issue where Reduction and GetResult is expanded completely by itself not following the DFS structure per dimension like the rest of the expansion code. This becomes especially problematic for multiple IterArg since Getitem is not expecting its' source value to be expanded without it.

### DIFF
--- a/shark_turbine/kernel/wave/expansion.py
+++ b/shark_turbine/kernel/wave/expansion.py
@@ -19,7 +19,7 @@ from ..ops.wave_ops import *
 from .._support.indexing import IndexingContext, IndexSequence
 from ...support.logging import get_logger
 from .._support.tracing import CapturedTrace
-from .utils import get_mma_dimensional_mapping, specialize_index_sequence
+from .utils import get_mma_dimensional_mapping, specialize_index_sequence, remove_chained_getresult
 from ..lang.global_symbols import *
 
 logger = get_logger("turbine.wave.expansion")
@@ -293,6 +293,7 @@ def expand_graph(
                 node_index_setter,
                 expansion_context,
             )
+    remove_chained_getresult(trace)
 
 
 def _expand_node(
@@ -321,7 +322,7 @@ def _expand_node(
         return _expand_reduction(
             node, trace, dim_query, dim_scaling, node_index_setter, context
         )
-    elif isinstance(node, GetResult):
+    elif isinstance(node, GetResult) and not isinstance(node, Getitem):
         # The presence of a GetResult node indicates that the reduction has already
         # been expanded. Simply return the corresponding node.
         reduction = get_custom(node.value)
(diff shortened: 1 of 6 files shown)


shark_turbine/kernel/wave/utils.py
```diff
@@ -123,6 +123,20 @@ def is_removable_operator(node: fx.Node) -> bool:
             get_custom(node).graph.erase_node(node)
 
 
+def remove_chained_getresult(trace: CapturedTrace):
+    def is_chained_getresult(node: fx.Node) -> bool:
+        custom = get_custom(node)
+        idxc = IndexingContext.current()
```
harsh-nod, line 129: Is idxc used anywhere?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -19,7 +19,11 @@
...
 from .._support.tracing import CapturedTrace
-from .utils import get_mma_dimensional_mapping, specialize_index_sequence
+from .utils import (
+    get_mma_dimensional_mapping,
+    specialize_index_sequence,
+    remove_chained_getresult,
+)
 from ..lang.global_symbols import *
 
 logger = get_logger("turbine.wave.expansion")
```
harsh-nod, line 29: Could you update the comment below now that the type of expandednodemap has changed?

shark_turbine/kernel/wave/expansion.py
```diff
@@ -293,6 +297,7 @@ def expand_graph(
                 node_index_setter,
                 expansion_context,
             )
+    remove_chained_getresult(trace)
```
harsh-nod, line 300: Should we move this out of this function to after expansion? You can move it to wave.py. This way we can isolate what expansion is doing vs the others.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -343,7 +351,7 @@ def repeat(
         # Assert equal does cast to boolean on torch.Tensor
         # which causes issues, hence we cast to numpy before
         # checking.
-        assert_equal(c, ref.values.numpy())
+        assert_allclose(ref, c, atol=0.07)
```
harsh-nod, line 354: That tolerance seems quite high for fp32?

harsh-nod (approved):
lgtm! modulo some comments

shark_turbine/kernel/wave/utils.py
raikonenfnu: good catch, removed, thanks!

shark_turbine/kernel/wave/expansion.py
raikonenfnu: done!

shark_turbine/kernel/wave/expansion.py
raikonenfnu: makes sense, done!

tests/kernel/wave/wave_e2e_test.py
raikonenfnu: lowered to 0.015 :)

PR #167, opened 2024-09-25
Get GEMMs working without minimize_global_loads

harsh-nod:
This PR removes the need for propagating indices using
post expansion. The new approach propagates the MMA
indices to the MMA dimensions of all tensors (rather
than just MMA nodes) and then specializes them depending
on whether they lie within the backward slices of the
LHS and RHS or forward slices of the ACC.

### DIFF
--- a/shark_turbine/kernel/wave/utils.py
+++ b/shark_turbine/kernel/wave/utils.py
@@ -1,5 +1,4 @@
 # Copyright 2024 The IREE Authors
-#
 # Licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -16,7 +15,16 @@ from typing import Callable, Any, List, Tuple
 from .._support.tracing import CapturedTrace
 from .._support.indexing import IndexExpr, IndexingContext, IndexSymbol, IndexSequence
 from ..lang.global_symbols import *
-from ..ops.wave_ops import get_custom, Output, Write, MMA
+from ..ops.wave_ops import (
+    get_custom,
+    Output,
+    Write,
+    MMA,
+    CustomOp,
+    Reduction,
+    GetResult,
+    IterArg,
+)
 from .constraints import Constraint, HardwareConstraint, TilingConstraint
 import torch.fx as fx
 import shark_turbine.kernel.lang as tkl
@@ -145,7 +153,9 @@ def simplify_index(index: IndexExpr) -> IndexExpr:
     return subs_idxc(index.subs(mapping))
 
 
-def get_mma_dimensional_mapping(trace: CapturedTrace) -> dict[IndexSymbol, int]:
+def get_mma_dimensional_mapping(
+    trace: CapturedTrace,
+) -> tuple[dict[IndexSymbol, int], dict[IndexSymbol, list[fx.Node]]]:
     """
     Given a trace, determine the MMA dimensional mapping for all the
     MMA operations in the graph. For example, if we have
@@ -159,7 +169,8 @@ def get_mma_dimensional_mapping(trace: CapturedTrace) -> dict[IndexSymbol, int]:
         return isinstance(get_custom(node), MMA)
 
     mapping: dict[IndexSymbol, int] = {}
-    for node in trace.walk(is_mma):
+    mma_nodes = trace.walk(is_mma)
+    for node in mma_nodes:
         custom: MMA = get_custom(node)
         m, n = custom.acc_type.symbolic_shape[-2:]
         lhs_shape = custom.lhs_type.symbolic_shape
@@ -170,7 +181,7 @@ def get_mma_dimensional_mapping(trace: CapturedTrace) -> dict[IndexSymbol, int]:
         mapping[n] = 1
         mapping[k] = 2
 
-    return mapping
+    return mapping, capture_mma_slices([get_custom(x) for x in mma_nodes])
 
 
 def get_hardware_vector_size(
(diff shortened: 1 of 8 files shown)


shark_turbine/kernel/wave/utils.py
```diff
@@ -378,3 +389,132 @@ def erase_graph(graph: fx.Graph):
...
+            iter_arg_idx = reduction.iter_args.index(node)
+            inputs.append(reduction.init_args[iter_arg_idx])
+            continue
+        inputs.append(input)
+    return inputs, reduction
+
+
+def bfs(
+    node: fx.Node,
+    visited: list[fx.Node],
```
raikonenfnu, line 454: I don't think we need the `visited`/`queue` right? especially since it's iterative search?

shark_turbine/kernel/wave/utils.py
```diff
@@ -378,3 +389,132 @@ def erase_graph(graph: fx.Graph):
...
+    based on whether the node is used as the LHS, RHS or ACC in the MMA node.
+    If the node is not used as any of the operands, return the original index sequence
+    with all the MMA symbols zeroed out.
+    """
+    if isinstance(custom, MMA):
+        return index_seq
+    operand_map = {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 0}
+    for key in mma_slices:
+        if custom.fx_node in mma_slices[key]:
+            operand_map[key] = 1
```
raikonenfnu, line 518: IIUC, we want for every for key in mma_slices to be 
{MMA_LHS: 1, MMA_RHS: 0, MMA_ACC: 0} then
{MMA_LHS: 0, MMA_RHS: 1, MMA_ACC: 0} and then
{MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1},

But in current state wouldn't this be 
{MMA_LHS: 1, MMA_RHS: 0, MMA_ACC: 0}
{MMA_LHS: 1, MMA_RHS: 1, MMA_ACC: 0}
{MMA_LHS: 1, MMA_RHS: 1, MMA_ACC: 1}

Although if it that is indeed what we are going for, can you explaine intuition behind it? :)

raikonenfnu (commented):
Code looks good overall, left some comments. 

Also quick conceptual Q, so in this PR are we just obtaining the slices to collect all ops that will be impacted by indexing of MMA, and then adjusting their index_seq appropriately S.T it will work with the MMA indexing?

raikonenfnu:
> Code looks good overall, left some comments.
> 
> Also quick conceptual Q, so in this PR are we just obtaining the slices to collect all ops that will be impacted by indexing of MMA, and then adjusting their index_seq appropriately S.T it will work with the MMA indexing?

Seems like the "specialization"/"adjusting" of index is only related with subbing in the MMA_ACC/MMA_LHS/MMA_RHS to respective op's index_seq, if this is not happening before, wouldn't we still have a symbolic indexing and the program won't run at all?

raikonenfnu:
> > Code looks good overall, left some comments.
> > Also quick conceptual Q, so in this PR are we just obtaining the slices to collect all ops that will be impacted by indexing of MMA, and then adjusting their index_seq appropriately S.T it will work with the MMA indexing?
> 
> Seems like the "specialization"/"adjusting" of index is only related with subbing in the MMA_ACC/MMA_LHS/MMA_RHS to respective op's index_seq, if this is not happening before, wouldn't we still have a symbolic indexing and the program won't run at all?

Ohhh actually were we circumventing around that issue by using `self.lhs.index = self.lhs_index` (similarly for rhs and out)?

shark_turbine/kernel/wave/utils.py
harsh-nod: So the idea here is that we want to find all the transitive dependencies of the lhs, rhs (backward slice) and acc (forward slice). So find all the nodes that could affect lhs and rhs and all the nodes that are affected by acc. visited keeps track of all these nodes. queue just helps with the bfs. The reason we do that is so that now we are assigning global per dimension indices. So we need to "specialize" these indices by identifying whether they lie in the backward slices of the lhs, rhs or forward slice of the acc. When we have multiple mma ops, then we will have some of these slices intersecting. Specifically, the forward acc slice will intersect with the backward slices of the lhs and rhs. This would be the equivalent of a conflict.

shark_turbine/kernel/wave/utils.py
harsh-nod: So if a node is determined to be in the backward slice of the LHS, then we want to specialize it by substituting {MMA_LHS = 1, all else 0}. For RHS, we want {MMA_RHS = 1, all else 0}. For ACC, {MMA_ACC = 1, all else 0}. And if its not in the backward slices of the LHS and RHS or forward slice of the ACC, then {all = 0}. You can think of this as an alternative to propagation. Because we set the entire indices everywhere, we need to specialize them depending on some constraints, and for that we use the forward/backward slices of the MMA operands.

harsh-nod:
> > > Code looks good overall, left some comments.
> > > Also quick conceptual Q, so in this PR are we just obtaining the slices to collect all ops that will be impacted by indexing of MMA, and then adjusting their index_seq appropriately S.T it will work with the MMA indexing?
> > 
> > 
> > Seems like the "specialization"/"adjusting" of index is only related with subbing in the MMA_ACC/MMA_LHS/MMA_RHS to respective op's index_seq, if this is not happening before, wouldn't we still have a symbolic indexing and the program won't run at all?
> 
> Ohhh actually were we circumventing around that issue by using `self.lhs.index = self.lhs_index` (similarly for rhs and acc)?

So what was happening before is that we were setting the MMA indices only to the MMA op and then propagating this to the operands during post-expansion. The problem with this was that it required us to do even more propagation to get the shared memory IGEMM case working (since we had to propagate the indices all the way to the global read). As an alternative to propagation, we are now setting globally per dimension indices that include the effects of all the constraints. But the problem with this is you end up with a whole bunch of Piecewise functions that you can't reason about. So we use the slices to determine how to convert these Piecewise functions to indices. As I said in another comment, this will be useful when we deal with multiple MMAs because we will now have overlapping slices and can do some sort of "equality saturation" to determine how to resolve multiple indices per node.

shark_turbine/kernel/wave/utils.py
raikonenfnu: makes sense, in that case I think we need to move the `operand_map = {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 0}` above the `if custom.fx_node in mma_slices[key]:`. Otherwise the previous state carry over. i.e we will get:
```
{MMA_LHS: 1, MMA_RHS: 0, MMA_ACC: 0} then
{MMA_LHS: 0, MMA_RHS: 1, MMA_ACC: 0} and then
{MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1},
```

raikonenfnu:
> > > > Code looks good overall, left some comments.
> > > > Also quick conceptual Q, so in this PR are we just obtaining the slices to collect all ops that will be impacted by indexing of MMA, and then adjusting their index_seq appropriately S.T it will work with the MMA indexing?
> > > 
> > > 
> > > Seems like the "specialization"/"adjusting" of index is only related with subbing in the MMA_ACC/MMA_LHS/MMA_RHS to respective op's index_seq, if this is not happening before, wouldn't we still have a symbolic indexing and the program won't run at all?
> > 
> > 
> > Ohhh actually were we circumventing around that issue by using `self.lhs.index = self.lhs_index` (similarly for rhs and acc)?
> 
> So what was happening before is that we were setting the MMA indices only to the MMA op and then propagating this to the operands during post-expansion. The problem with this was that it required us to do even more propagation to get the shared memory IGEMM case working (since we had to propagate the indices all the way to the global read). As an alternative to propagation, we are now setting globally per dimension indices that include the effects of all the constraints. But the problem with this is you end up with a whole bunch of Piecewise functions that you can't reason about. So we use the slices to determine how to convert these Piecewise functions to indices. As I said in another comment, this will be useful when we deal with multiple MMAs because we will now have overlapping slices and can do some sort of "equality saturation" to determine how to resolve multiple indices per node.
> 
> This was not a problem before because we restricted these MMA_{LHS/RHS/ACC} variables only to MMA nodes and their neighbors.

Makes sense, thanks! :)

shark_turbine/kernel/wave/utils.py
raikonenfnu: That makes sense, although what I meant was couldn't we initialize `visited`/`queue` inside the function? Since we always send empty list for both visited and queue anyways?

shark_turbine/kernel/wave/utils.py
```diff
@@ -378,3 +389,132 @@ def erase_graph(graph: fx.Graph):
...
+                visited.append(neighbor)
+                queue.append(neighbor)
+    return visited
+
+
+def capture_forward_slice(node: fx.Node) -> set[fx.Node]:
+    """
+    Run BFS on the graph to capture the forward slice of a node.
+    """
+    return bfs(node, [], [], lambda x, y: get_users(x, y))
```
raikonenfnu, line 478: RE:
> When we have multiple mma ops, then we will have some of these slices intersecting. Specifically, the forward acc slice will intersect with the backward slices of the lhs and rhs. This would be the equivalent of a conflict

Can we make the make `visited` a set then for easier index conflict later on?

shark_turbine/kernel/wave/utils.py
harsh-nod: Ah yes, that makes sense. Will change.

shark_turbine/kernel/wave/utils.py
harsh-nod: Yes, makes sense. Will change.

shark_turbine/kernel/wave/utils.py
raikonenfnu: Realized I put in the wrong state on the previous comment, updated it to make it make more sense haha

harsh-nod: Ah yes, we would get carry over, except for the fact that we return as soon as we get a match. So that guarantees that we our dictionary's values will always only have one non-zero entry (= 1) .

raikonenfnu: OK that makes sense that's why it's implicitly functionally equivalent. Can we still bring it down though for better clarity/straightforward-ness? :)

raikonenfnu (approved):
LGTM!

PR #173, opened 2024-09-27
Fixes/updates to get tests passing

IanNod:


shark_turbine/aot/passes/functorch.py
```diff
@@ -42,15 +42,13 @@
 # destructuring rewrites that torch.export does.
 def functorch_functionalize(gm_callable: Any, *args) -> GraphModule:
     functionalized_callable = _functionalize_callabale(gm_callable)
-    # TODO: There is more of a dance needed if the user has entered with a fake_mode.
-    with proxy_tensor.maybe_disable_fake_tensor_mode():
```
saienduri, line 46: Btw, we can change this to conform with the latest pytorch API
```
from torch._subclasses.fake_tensor import unset_fake_temporarily
with unset_fake_temporarily():
```
https://github.com/pytorch/pytorch/pull/132770/files

IanNod: I thought I saw a patch that looked like this became unnecessary but not seeing it now. If this is still needed I can unset_fake_temporarily. Little surprised we don't have any unit tests that look to be using fake_tensors

PR #176, opened 2024-09-27
[TKW] Implement broadcastOp class, lowering and insertion

raikonenfnu:
Motivation of this PR is to be able to codegen/lower broadcast properly. With that in mind, we implemented these things:

1. BroadcastOp class, op and lowering, to represent and store broadcasting information. Mostly S.T we can query target shape information and the source operand of broadcast.
2. Treat broadcast-add as an index conflict and handle it by emitting broadcastOp.


shark_turbine/kernel/wave/codegen.py
```diff
@@ -995,6 +996,79 @@ def handle_extract_slice(emitter: WaveEmitter, node: fx.Node):
...
+    hw_constraint = hw_constraints[0]
+    if not hasattr(hw_constraint, "vector_shapes"):
+        raise NotImplementedError(
+            "Only support broadcast for kernel with non-MMA constraints."
+        )
+
+    # As of writing of this pass, TKW expects only 1D vectors,
+    # or only 1 non-unit dim vector_shape. So we are not handling
+    # broadcast of N-d case.
+    broadcast_dims = set(target_shape) - set(src_shape)
```
harsh-nod, line 1031: So this actually made me think about this broadcast op as a whole. I'm not sure you need to store [M, N] at all. Since we are forcing all vectors to be 1-d, you will always end up with vector<Qxf16> + vector<1xf16> (everything else will be disallowed). If that's the case, then we only need to keep track of what Q is in the broadcast op and I think you get that from elements_per_thread + propagation of those shapes. Let me know if I am missing something.

harsh-nod (changes requested):
I didn't completely understand why we need to keep track of the original shapes in the broadcast op. Since our vectors are always 1D I think that should simplify a lot of this.

shark_turbine/kernel/wave/codegen.py
raikonenfnu: That is a fantatstic comment/Q! In general, this is true but only implicitly, there is nothing in our language (from user perspective) that forces the vector_shapes to only have a single unit-dim, and there is no guarantees that we won't be supporting say 2D vectors in the future. 

TL;DR: Analysis is set to explicitly error out/safeguard unwanted broadcasts from happening, i.e if we do not check original shape say (M -> K,N) and if we implement 2D in the future, it will tell us what is missing/needs to be done to fix that.

lit_tests/kernel/wave/codegen.py
```diff
@@ -1051,6 +1051,78 @@ def repeat(
...
+            LOAD_ELEMS_PER_THREAD: 2,
+            STORE_ELEMS_PER_THREAD: 2,
+            ADDRESS_SPACE: tkl.AddressSpace.GLOBAL_MEMORY.value,
+        },
+        canonicalize=True,
+        run=False,
+        run_config=config,
+    ):
+        print(test(a, b, c).module_op)
+        # CHECK: func.func @test(%[[ARG0:.+]]: !stream.binding, %[[ARG1:.+]]: !stream.binding, %{{.+}}: !stream.binding)
```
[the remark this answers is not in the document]
raikonenfnu: thanks! :)

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -890,7 +896,10 @@ def captured_vars(self, graph: fx.Graph) -> list[fx.Node]:
 
     @property
     def type(self) -> list[Memory | Register]:
```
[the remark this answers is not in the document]
raikonenfnu: yeap, changed!

shark_turbine/kernel/wave/codegen.py
harsh-nod: That makes sense and I think I see why you need to store [M, N]. Let me try and crystallize my thoughts.

raikonenfnu: WRT just the original shape, we need the original shape accessible for expansion, register analysis, index_sequence_analysis as well

harsh-nod: So say we have a 3x4xi8 tensor with the following access patterns

1 1 2 2
3 3 4 4
5 5 6 6

And we want to add it to a 3xi8 tensor with the following access patterns

3
4
1

We can do it, but extract + splat won't give the right result.

On the other hand, if we had a 3xi8 with this access pattern

1
3
5

Now we could do an extract + splat which gives

1 1
3 3
5 5

This still doesn't give the right result. It only does if we the 3xi8 is
the result of a butterfly shuffle in which case threads 2, 4 and 6
have the same values as 1, 3 and 5 and so doing extract + splat ends up with

1 1 2 2
3 3 4 4
5 5 6 6

So this brings me to my first point which is we do have to maintain the tensor
dims so that we can compare the index sequences of the common dims to make
sure they are the same and either throw an error if not or emit shuffles.
We also have to keep note of whether the 1d tensor is the result of a butterfly
shuffle reduction and if not then throw an error or emit shuffles.

Now if we all the above conditions hold, then we need to do extract + splat.
For that, I think it would be easier if we just keep track of elements per
thread rather than recalculate it.

raikonenfnu: I don't quite understand 3 4 1 access pattern very well, not sure if I understand the second case fully, why wouldn't the 2nd thread expand it's own data which => it doesn't need to shuffle data. Maybe we should chat over VC haha

harsh-nod: So 3 4 1 is just a sample random access pattern. What I was getting at is that in general the access patterns of the [N, M] tensor and [M] tensor might not be the same and so we need to add a check for this. Yes, sounds good :)

raikonenfnu: Just double checking, if I understand your most recent comments correctly, we essentially need to do two things: 
1. we need to add more checks for the N,M case
2.refactor to use elements per thread instead of recomputing?

WRT 1, we can chat more over VC on Monday so we can draw out shapes and examples better. WRT 2, totally agree, although that means we'd need to start propagating elements_per_thread/ index sequence sizes. IIRC, currently we are only setting indexSequence size for reads/writes and MMA.

harsh-nod (approved):
lgtm! thanks!

shark_turbine/kernel/ops/wave_ops.py
```diff
@@ -522,9 +528,16 @@ def type(self) -> Memory:
...
-            raise ValueError("Expected lhs and rhs to have same type post-expansion")
-        return lhs_type
+        if has_same_type:
+            return lhs_type
+        lhs_dim_set = set(lhs_type.symbolic_shape)
+        rhs_dim_set = set(rhs_type.symbolic_shape)
+        if lhs_dim_set.isdisjoint(rhs_dim_set):
+            raise ValueError(
+                "BinaryPyOp requires lhs and rhs shape to be at least broadcastable."
+            )
```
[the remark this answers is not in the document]
raikonenfnu: done!

PR #177, opened 2024-09-27
[TKW] Fix types, shapes and propagate resolved indexing 

raikonenfnu:
This PR add supports for doing accumulate on non-induction variables. For most part our stack already supports this, but we'd need to fix reduction symbolic shape which used to be just the input shape to input shape - reduction dim. Without this our thread shape analysis wouldn't be able to handle broadcasting of reduced op properly.

To support the above, we also introduce `Extract` op in addition to the existing `ExtractSlice` to fix the shape types. Specifically, ExtractSlice follow upstream's semantic where we just slice but do not reduce any dimensions. For the ReduceOp case, specifically the local reduction, we'd want it to have a reducing semantic on the fastest dimension.

Additionally, we also add a propagation of our resolution for thread shape. This is helpful in the test we added which is a broadcast-sub followed by an exp2.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -45,6 +45,13 @@ def allocate(
     ...
 
 
+def extract(
+    register: "Register",
+    offsets: tuple[IndexExpr],
+) -> "Register":
+    ...
+
+
 def extract_slice(
     register: "Register",
     offsets: tuple[IndexExpr],
@@ -1057,6 +1064,52 @@ class GetResult(CustomOp):
         CustomOp.index.fset(self, value)
 
 
+@define_op("extract")
+@dataclass
+class Extract(CustomOp):
+    """
+    Op Rationale:
+
+    Extract is an op used to represent extracting of
+    a scalar from TKW's 1-D vector on the specified index.
+
+    This can also be viewed as indexing/slicing on the fastest
+    dimension. Hence, the semantic of this op is designed to
+    see itself as a reduction on the indexed/fastest dimension.
+    """
+
+    register_: fx.Proxy
+    offset: IndexExpr | int
+
+    @property
+    def type(self) -> "Register":
+        # Intuition here is we are trying to extract an element
+        # from fastest dim => we reduce the fastest dim.
+        src_type = get_custom(self.register_).type
+        # Return itself if just 0-D/1-D symbolic.
+        if len(src_type.symbolic_shape) <= 1:
+            return src_type
+
+        # Typically only fastest dim has non-unit dim,
+        # but if all unit-dim get fastest/last one.
+        all_unit_dims = lambda index: all(x.size == 1 for x in index.values())
+        if all_unit_dims(self.register_.index):
+            dst_shape = src_type.symbolic_shape[:-1]
+            dst_type = Register[*dst_shape, src_type.dtype]
+            return dst_type
+
+        # If non-unit dim exists => non-unit dim is fastest dim.
+        non_unit_dim = [k for k, v in self.register_.index.items() if v.size != 1]
+        if len(non_unit_dim) != 1:
+            raise NotImplementedError(
+                f"NYI: Extract only support exactly 1 non-unit dim, but found: {len(non_unit_dim)}"
+            )
+        dst_shape = list(src_type.symbolic_shape)
+        dst_shape.remove(non_unit_dim[0])
+        dst_type = Register[*dst_shape, src_type.dtype]
+        return dst_type
+
+
 @define_op("extract_slice")
 @dataclass
 class ExtractSlice(CustomOp):
@@ -1116,12 +1169,16 @@ class ReduceOp(CustomOp, ABC):
 
     @property
     def indexing_dims(self) -> list[IndexSymbol]:
-        return get_custom(self.arg).indexing_dims
+        src_indexing = get_custom(self.arg).indexing_dims
+        dst_indexing = [dim for dim in src_indexing if dim != self.dim]
+        return dst_indexing
 
     @property
     def type(self) -> Memory:
         src_type = get_custom(self.arg).type
-        return src_type
+        reduced_dims = [dims for dims in src_type.symbolic_shape if dims != self.dim]
+        dst_type = Register[*reduced_dims, src_type.dtype]
+        return dst_type
 
     @property
     def num_reduction_dims(self) -> int:
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -1073,9 +1073,10 @@ class Extract(CustomOp):
     Extract is an op used to represent extracting of
     a scalar from TKW's 1-D vector on the specified index.
 
-    This can also be viewed as indexing/slicing on the fastest
-    dimension. Hence, the semantic of this op is designed to
-    see itself as a reduction on the indexed/fastest dimension.
+    This can also be viewed as indexing/slicing on the fastest dim/
+    SIMT/thread world, as opposed to the SIMD/wave world.
+    Hence, from the SIMD/wave world POV, the indexed dim
+    will be unit-size/"reduced".
     """
 
     register_: fx.Proxy
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -1073,10 +1073,9 @@ class Extract(CustomOp):
     Extract is an op used to represent extracting of
     a scalar from TKW's 1-D vector on the specified index.
 
-    This can also be viewed as indexing/slicing on the fastest dim/
-    SIMT/thread world, as opposed to the SIMD/wave world.
-    Hence, from the SIMD/wave world POV, the indexed dim
-    will be unit-size/"reduced".
+    This can also be viewed as indexing/slicing on the fastest
+    dimension. Hence, the semantic of this op is designed to
+    see itself as a reduction on the indexed/fastest dimension.
     """
 
     register_: fx.Proxy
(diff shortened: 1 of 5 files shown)


raikonenfnu:
Currently broken on igemm because ExtractSlice used in iGEMM semantic needs to follow upstream. We need to implement an extract/indexing op for semantic required in local thread reduction

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1030,6 +1037,52 @@ def index(self, value: dict[IndexSymbol, IndexSequence]):
...
+class Extract(CustomOp):
+    """
+    Op Rationale:
+
+    Extract is an op used to represent extracting of
+    a scalar from TKW's 1-D vector on the specified index.
+
+    This can also be viewed as indexing/slicing on the fastest
+    dimension. Hence, the semantic of this op is designed to
+    see itself as a reduction on the indexed/fastest dimension.
```
harsh-nod, line 1051: I am a little confused about how an extract is a reduction on the indexed/fastest dimension? Say you have a <8xf32> and you extract [0] you will get a f32, but how is this a reduction?

harsh-nod: Or maybe we should clarify - in the wave paradigm, when you do an extract of [M, N][0, :] -> [N] dimensional tensor, whereas in the thread paradigm, a <8xf32> -> f32 ? Is this what you mean?

raikonenfnu: That is correct, say we have a tensor named `input` of shape [M, N], by indexing on the fastest dimension i.e `extracted = input[:, 3]` where 3 can be arbitary number between 0 and N. `extracted` will be of shape `Mx1` which is where the "reduction" came from.

harsh-nod: If so, I still would not call this a reduction.

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1030,6 +1037,52 @@ def index(self, value: dict[IndexSymbol, IndexSequence]):
...
+    """
+
+    register_: fx.Proxy
+    offset: IndexExpr | int
```
harsh-nod, line 1055: In numpy semantics, if a = [4x8], then a[0] gives you a [8] tensor. This makes sense. What if a is more than 2d? If we don't want to handle that, we should just assert.

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: okay that makes sense. but I still wouldn't call it a reduction because that implies combining values in some way, whereas this is more of slicing.

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: okay let me reread, with the new information I have. Ignore this.

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1030,6 +1037,52 @@ def index(self, value: dict[IndexSymbol, IndexSequence]):
...
+
+        # Typically only fastest dim has non-unit dim,
+        # but if all unit-dim get fastest/last one.
+        all_unit_dims = lambda index: all(x.size == 1 for x in index.values())
+        if all_unit_dims(self.register_.index):
+            dst_shape = src_type.symbolic_shape[:-1]
+            dst_type = Register[*dst_shape, src_type.dtype]
+            return dst_type
+
+        # If non-unit dim exists => non-unit dim is fastest dim.
```
harsh-nod, line 1074: okay this is a little non-intuitive from the semantics of the op. Basically, this extract op is a slice along the axis with the most elements rather than the fastest dimension.

harsh-nod: Alternatively, you can call it a slice in the SIMT/thread world, as opposed to the SIMD/wave world?

iree/turbine/kernel/wave/codegen.py
```diff
@@ -1076,6 +1077,26 @@ def handle_shared_memory_barrier(emitter: WaveEmitter, node: fx.Node):
...
+@handle_op(extract)
+def handle_extract(emitter: WaveEmitter, node: fx.Node):
+    try:
+        register, offset = node.args
+    except ValueError as e:
+        raise ValidationError("Malformed arguments") from e
+    assert isinstance(offset, list) and len(offset) == 1
+    extract_vector = cast_vector(emitter, register)
+    result_type = VectorType.get([1], extract_vector.type.element_type)
+    element = vector_d.extract_strided_slice(
```
harsh-nod, line 1089: Just wondering why an extract() wouldn't work here instead?

iree/turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -13,6 +13,10 @@
 
 logger = get_logger("turbine.wave.thread_shape_analysis")
 
+################################################################
```
harsh-nod, line 16: Remnants from old code?

iree/turbine/kernel/ops/wave_ops.py
raikonenfnu: Yeah exactly! it's a SIMT slice, it is expected to be used as a lower level op. We extract in the vector/post-distributed/SIMT world, but it essentially retains symbolic shape of the whole tensor S.T we can be compliant in higher level analysis.

harsh-nod: make sense, can you update the documentation to describe that?

iree/turbine/kernel/wave/codegen.py
raikonenfnu: it does and that's what we should do, it was just easier to use the previous implementation haha

raikonenfnu: Tried moving it to vector_d.extract, but actually it makes our shuffleOp lowering a little more complicated since not all inputs are vectorType. Additionally, it may be good to keep extract_strided_slice for future upgrade to multi-row extract.

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: Going to resolve all these old comments for now.

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1057,6 +1064,53 @@ def index(self, value: dict[IndexSymbol, IndexSequence]):
...
+    Op Rationale:
+
+    Extract is an op used to represent extracting of
+    a scalar from TKW's 1-D vector on the specified index.
+
+    This can also be viewed as indexing/slicing on the fastest dim/
+    SIMT/thread world, as opposed to the SIMD/wave world.
+    Hence, from the SIMD/wave world POV, the indexed dim
+    will be unit-size/"reduced".
+    """
```
harsh-nod, line 1080: thanks, this looks good.

iree/turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -59,21 +80,32 @@ def handle_binaryop_conflict(custom_node: CustomOp):
     broadcast = Broadcast(broadcast_src.fx_node, dst_op.type)
     with custom_node.graph.inserting_before(custom_node.fx_node):
         broadcast.add_to_graph(custom_node.graph)
-    setattr(broadcast.fx_node, "index", dst_op.index)
-    custom_node.index = dst_op.index
     custom_node.update_arg(broadcast_idx, broadcast.fx_node)
-    return True
+    propagated_resolutions = capture_forward_slice(broadcast.fx_node, propagatable_op)
+    for node in propagated_resolutions:
+        setattr(node, "index", dst_op.index)
```
harsh-nod, line 86: can we use the custom interface to set the index instead?

iree/turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -59,21 +80,32 @@ def handle_binaryop_conflict(custom_node: CustomOp):
     broadcast = Broadcast(broadcast_src.fx_node, dst_op.type)
     with custom_node.graph.inserting_before(custom_node.fx_node):
         broadcast.add_to_graph(custom_node.graph)
-    setattr(broadcast.fx_node, "index", dst_op.index)
-    custom_node.index = dst_op.index
     custom_node.update_arg(broadcast_idx, broadcast.fx_node)
-    return True
+    propagated_resolutions = capture_forward_slice(broadcast.fx_node, propagatable_op)
```
harsh-nod, line 84: Do we need to capture the forward slice? Isn't it enough to set the index on the conflicted_ops ?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1057,6 +1064,53 @@ def index(self, value: dict[IndexSymbol, IndexSequence]):
...
+        # Intuition here is we are trying to extract an element
+        # from fastest dim => we reduce the fastest dim.
+        src_type = get_custom(self.register_).type
+        # Return itself if just 0-D/1-D symbolic.
+        if len(src_type.symbolic_shape) <= 1:
+            return src_type
+
+        # Typically only fastest dim has non-unit dim,
+        # but if all unit-dim get fastest/last one.
+        all_unit_dims = lambda index: all(x.size == 1 for x in index.values())
```
harsh-nod, line 1096: Do we need this check? Can we merge this with the non_unit_dim check? if len(non_unit_dim) == 0, use fastest dim.

harsh-nod (changes requested):
This looks pretty good. Just a few comments

iree/turbine/kernel/wave/thread_shape_analysis.py
harsh-nod: Based on our conversation, this makes sense as we need to propagate the resolved index to downstream ops.

iree/turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -59,21 +80,32 @@ def handle_binaryop_conflict(custom_node: CustomOp):
...
-            handle_binaryop_conflict(custom)
+            resolved_ops = handle_binaryop_conflict(custom)
+            cummulative_resolved = cummulative_resolved.union(resolved_ops)
         else:
-            return False
-    return True
+            continue
+    # Superset because path/cumulative resolved includes resolution helper ops
+    # such as broadcast.
+    all_conflicts_resolved = cummulative_resolved.issuperset(conflicted_ops)
```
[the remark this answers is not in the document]
harsh-nod: didn't realize this takes the set of ops as input, makes sense.

harsh-nod (approved):
lgtm! just some minor nits

iree/turbine/kernel/wave/thread_shape_analysis.py
raikonenfnu: done!

iree/turbine/kernel/ops/wave_ops.py
raikonenfnu: done! I think :)

PR #178, opened 2024-09-28
Add support for dynamic dims

harsh-nod:
This PR adds support for dynamic dimensions in the
kernels. The user specifies the dynamic dimensions
by
- Not adding them to the hyperparameter dictionary
- Explicitly specifying them with the dynamic_symbols kwarg
  and the dynamic_symbols_mapping kwarg to specify which
  values to use for the dynamic dims at runtime

This PR does not modify the codegen and so incorrect or
unsupported values for the dynamic dims will result
in incorrect results. (garbage in -> garbage out)

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -355,7 +355,7 @@ def test_dynamic_copy():
         a = torch.randn(16, 16, dtype=torch.float16)
         print(test(a).module_op)
 
-    # CHECK:        stream.executable.export public @test workgroups(%[[ARG0:[a-zA-Z0-9_]+]]: index, %[[ARG1:[a-zA-Z0-9_]+]]:
+    # CHECK:        stream.executable.export public @test workgroups(%[[ARG0:.*]]: index, %[[ARG1:.*]]:
     # CHECK-SAME:       index) -> (index, index, index) {
     # CHECK-DAG:        %[[C1:.+]] = arith.constant 1 : index
     # CHECK-DAG:        %[[C16:.+]] = arith.constant 16 : index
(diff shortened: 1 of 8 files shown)


lit_tests/kernel/wave/codegen.py
```diff
@@ -328,6 +334,72 @@ def test(
...
+
+    # CHECK:        stream.executable.export public @test workgroups(%[[ARG0:[a-zA-Z0-9_]+]]: index, %[[ARG1:[a-zA-Z0-9_]+]]:
+    # CHECK-SAME:       index) -> (index, index, index) {
+    # CHECK-DAG:        %[[C1:.+]] = arith.constant 1 : index
+    # CHECK-DAG:        %[[C16:.+]] = arith.constant 16 : index
+    # CHECK:            %[[D0:.+]] = arith.ceildivsi %[[ARG0]], %[[C16]] : index
+    # CHECK:            %[[D1:.+]] = arith.ceildivsi %[[ARG1]], %[[C16]] : index
+    # CHECK:            stream.return %[[D0]], %[[D1]], %[[C1]] : index, index, index
+    # CHECK:          }
+    # CHECK:          func.func @test(%[[ARG0]]: !stream.binding, %[[ARG1]]: index, %[[ARG2:[a-zA-Z0-9_]+]]: index)
```
Hardcode84, line 366: nit: `[[ARG0:.*]]`?. And as general comment, MLIR doesn't really guarantee SSA names format, but `[[ARG0:.*]]` should work in all cases, don't need `[a-zA-Z0-9_]+`

Hardcode84 (approved):

lit_tests/kernel/wave/codegen.py
harsh-nod: sure will change.

PR #182, opened 2024-10-01
[TKW] Update IR interpreter

Hardcode84:
* Add `arith.andi`, `arith.cmpi`, `vector.maskedload`, `vector.gather`, `vector.contant_mask`, `vector.insertelement`, `vectot.splat`, support non-splatted contants.
* Add `interpret_ndrange` helper


harsh-nod (approved):
Thanks! Eventually we will want to refactor this code, but since it is just a debugging tool for now, this is fine.

PR #183, opened 2024-10-02
Add benchmark support for e2e tests

erman-gurses:


shark_turbine/kernel/wave/utils.py
```diff
@@ -366,7 +369,29 @@ def compile_and_invoke(
         _invoke(ctx.vm_context, device, func, kernel_inputs, kernel_outputs)
 
     if run_bench:
-        inputs = [inp.numpy() for inp in kernel_inputs]
+        bench_with_constant_weights = config.get("backend", False)
+        tempfiles = []
+        inputs = []
+        if bench_with_constant_weights:
```
Hardcode84, line 375: Why do we need this?

shark_turbine/kernel/wave/utils.py
```diff
@@ -366,7 +369,29 @@ def compile_and_invoke(
...
+            for inp in kernel_inputs:
+                inputs.append(
+                    "x".join(
+                        [str(x) for x in inp.shape]
+                        + [TORCH_DTYPE_TO_SIGNED_MLIR_TYPE_ASM[inp.dtype]]
+                    )
+                )
+        else:
+            for inp in kernel_inputs:
+                tf = tempfile.NamedTemporaryFile(delete=False)
```
Hardcode84, line 385: Why `delete=False`?

shark_turbine/kernel/wave/utils.py
```diff
@@ -375,6 +400,8 @@ def compile_and_invoke(
             **benchmark_flags,
         )
         _print_bench_result(benchmark_results, bench_file)
+        for tf in tempfiles:
```
Hardcode84, line 403: The files will closed when `tempfiles` list is destroyed (or or can `del tempfiles` manually), don't need to call `close` on each file manually.

shark_turbine/kernel/wave/utils.py
```diff
@@ -366,7 +369,29 @@ def compile_and_invoke(
         _invoke(ctx.vm_context, device, func, kernel_inputs, kernel_outputs)
 
     if run_bench:
-        inputs = [inp.numpy() for inp in kernel_inputs]
+        bench_with_constant_weights = config.get("backend", False)
```
Hardcode84, line 372: Wrong config option name.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -101,6 +101,7 @@ def test(
         },
         canonicalize=True,
         run=True,
+        run_bench=True,
```
Hardcode84, line 104: This should not be enabled by default, we need a separate option. Also, where bench results are saved?

shark_turbine/kernel/wave/utils.py
erman-gurses: Do you mean when we have "delete=True", we do not need to close the file?

shark_turbine/kernel/wave/utils.py
erman-gurses: Can you elaborate?

Hardcode84 (approved):

PR #184, opened 2024-10-02
[TKW] Fix sympy expr lowering and add some more igemm test shapes

Hardcode84:
* Rework how we are lowering `rational` sympy expressions, instead of delayed materialization via lambdas introduce `_Rational` type and propagate `numerator/denominator` values independently. Division will only be materialized on explicit `sympy.floor/ceiling` op.
* Rework how igemm test cases are generated and introduce few real shapes.
* Use custom pytest markers to separate perf/non-perf tests


Hardcode84:
@harsh-nod @erman-gurses I found a way to separate perf/non-perf tests using custom markers, don't need any external files, see the updated files and `--runperf` option.

tests/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--runperf", action="store_true", default=False, help="run performace tests"
```
harsh-nod, line 12: typo: performance

tests/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+
+
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers", "perf_only: performace test, runs only with '--runperf'"
+    )
+
+
+def pytest_collection_modifyitems(config, items):
```
harsh-nod, line 22: very cool! thanks for discovering this! much cleaner than other options.

shark_turbine/kernel/wave/codegen.py
```diff
@@ -200,76 +201,135 @@ def add_emitter_subs(emitter: WaveEmitter) -> dict[IndexSymbol, Any]:
...
+            arg = arg.owner.opview
 
-            if is_mul:
-                operation = arith_d.MulIOp(*_broadcast(operation, arg))
-            else:
-                operation = arith_d.AddIOp(*_broadcast(operation, arg))
+        if isinstance(arg, arith_d.ConstantOp):
+            value = arg.attributes["value"]
+            if isinstance(value, IntegerAttr):
+                return int(value)
```
harsh-nod, line 243: Throw error if it is not an int?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -200,76 +201,135 @@ def add_emitter_subs(emitter: WaveEmitter) -> dict[IndexSymbol, Any]:
...
 
-        for arg in callables:
-            operation = arg(operation, is_mul)
+        return None
 
-        stack.append(operation)
+    def muli_fold(lhs, rhs):
```
harsh-nod, line 247: When does this come up? Is it when you are handling with cases like ```a * (1 / b) ``` where you first do ``` a * 1 ``` and then divide by b? But wouldn't sympy convert that to ```a/b``` ?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -200,76 +201,135 @@ def add_emitter_subs(emitter: WaveEmitter) -> dict[IndexSymbol, Any]:
...
-            *_broadcast(mul(x) if is_mul else add(x), denominator)
-        )
+    def _ceiling(value):
+        if isinstance(value, _Rational):
+            value = arith_d.ceildivsi(*_broadcast(value.numerator, value.denominator))
+
+        return value
+
+    def _group_rationals(stack, count):
+        rationals = []
```
harsh-nod, line 307: Could you add a comment explaining why we need non rationals first and then rationals? I am assuming that is the purpose of this function?

shark_turbine/kernel/wave/codegen.py
```diff
@@ -200,76 +201,135 @@ def add_emitter_subs(emitter: WaveEmitter) -> dict[IndexSymbol, Any]:
...
+        if get_const_val(lhs) == 1:
+            return rhs
+
+        if get_const_val(rhs) == 1:
+            return lhs
+
+        return arith_d.muli(lhs, rhs)
+
+    # `x + (a/b)` transformed into `(x*b + a) / b`
+    def _add(lhs, rhs):
```
harsh-nod, line 257: I really like the idea of the Rational and late materialization. It was a much better approach than our previous lambda based approach. Thanks!

harsh-nod (approved):
I really like the refactor of gen_sympy_index for Rationals. Thank you! Just some minor comments but otherwise looks g2g!

shark_turbine/kernel/wave/codegen.py
Hardcode84: We can actually encounter vector constants now, so this code is intended to fall through to the `return None`

shark_turbine/kernel/wave/codegen.py
Hardcode84: > But wouldn't sympy convert that to a/b ?

I'm still seeing a bunch of `muli %a, 1` without this, so apparently not. They are benign as they will be cleaned up later by canonicalizations, but it also means we will have to update all our codegen lit tests. With this early folding lit codegen IR is identical.

shark_turbine/kernel/wave/codegen.py
Hardcode84: added comment

PR #186, opened 2024-10-02
[TKW] Thread Shape analysis

raikonenfnu:
The motivation of this pass is to generalize the register analysis pass which is used to determine the thread shape of TKW.Register, to all other operations.

One main use case for such is to allow reduction, and later on "broadcast" to use thread shape information from the kernel as opposed to relying on vector_shape which may not always be valid.

We generalize the register analysis metho by finding  a few anchor ops who's thread shape information is determined, and then propagate to it's successors and ancestors.

In addition to that we also implemented a couple helper function/attributes.

1. Control_fn on BFS, ForwardSlice, BackwardSlice. This is to make it easier for us to control/stop the search when we hit ops we do not want to explore. In this case, we do not want to explore/propagate onto other anchor ops and their children.

2. Introducing parent_op to IterArg and region of Reduction, for developer ergonomics.

3. Move handling of IterArg and GetUser in BackwardSlice/BFS's get_input exploration phase to be handled individually as opposed to being handled when its' consumer is being explored. Previously to explore/propagate IterArg/GetUser, we need to explore its' consumer, just exploring IterArg/GetUser will not get handled correctly. This is useful for the case where we want to propagate/explore mma.acc (usually IterArg) directly.


shark_turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -0,0 +1,118 @@
...
+from ..ops.wave_ops import *
+from ..lang.global_symbols import *
+from .utils import capture_forward_slice, capture_backward_slice
+
+logger = get_logger("turbine.wave.thread_shape_analysis")
+
+
+@dataclass(order=True)
+class IndexSize:
+    index: IndexSymbol
```
harsh-nod, line 19: minor but to avoid the confusion with index (which is the index sequence in the rest of the code), I would call this dim since it is an IndexSymbol. Then some of the other lines get easier to understand like
```
custom.index[target.dim].size = target.size
```
I think you might want to rename other functions from index -> dim.

shark_turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -0,0 +1,118 @@
...
+    nonPropagatableTypes = anchorOpTypes + noHandleTypes
+
+    def is_anchor_op(node: fx.Node):
+        return isinstance(get_custom(node), anchorOpTypes)
+
+    def propagatable_op(node: fx.Node):
+        return not isinstance(get_custom(node), nonPropagatableTypes)
+
+    anchor_ops = trace.walk(is_anchor_op)
+    thread_size_to_ops: dict[IndexSymbol, set[CustomOp]] = {}
```
harsh-nod, line 74: This is a pretty important variable. Can you add some documentation describing the key and value with some samples like in general it could look like
```
{frozenset({IndexSize(index=M, size=1), IndexSize(index=K, size=4)}): {read_shared_0_0_0}, frozenset({IndexSize(index=N, size=1), IndexSize(index=K, size=4)}): {read_shared_0_0_0}, frozenset({IndexSize(index=N, size=1), IndexSize(index=M, size=4)}): {acc_0_0_0}, frozenset({IndexSize(index=M, size=1), IndexSize(index=N, size=1)}): {}}
```

shark_turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -0,0 +1,118 @@
...
+                index_sizes, set([])
+            ).union(bwd_slice)
+        elif isinstance(custom, MMA):
+            lhs_bwd_slice = capture_backward_slice(custom.lhs, propagatable_op)
+            rhs_bwd_slice = capture_backward_slice(custom.rhs, propagatable_op)
+            acc_slice = capture_forward_slice(custom.acc, propagatable_op)
+            acc_slice.union(capture_backward_slice(custom.acc, propagatable_op))
+            acc_index = get_index_sizes(custom.acc_index)
+            lhs_index = get_index_sizes(custom.lhs_index)
+            rhs_index = get_index_sizes(custom.rhs_index)
```
harsh-nod, line 97: Just curious, how come we have to remove custom.fx_node from the backward slice of Write and forward slice of Read, but for MMA we don't need to remove anything from the forward or backward slices? Basically why dont we have anything like
```
bwd_slice.remove(custom.fx_node)
```
in the MMA case?

shark_turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def is_anchor_op(node: fx.Node):
+        return isinstance(get_custom(node), anchorOpTypes)
+
+    def propagatable_op(node: fx.Node):
+        return not isinstance(get_custom(node), nonPropagatableTypes)
+
+    anchor_ops = trace.walk(is_anchor_op)
+    thread_size_to_ops: dict[IndexSymbol, set[CustomOp]] = {}
+    for anchor_op in anchor_ops:
```
harsh-nod, line 75: Nicely done! This looks like a good way to propagate the thread shapes :)

shark_turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -0,0 +1,118 @@
...
+    for target in target_index_sizes:
+        if target.index not in custom.index:
+            raise NotImplementedError(
+                "NYI: Handle when source target index size is not found in target/user index."
+            )
+        custom.index[target.index].size = target.size
+
+
+# Function called on op post propagation for extra processing/handling.
+def post_propagation(custom: CustomOp, target_index_sizes: list[IndexSize]):
```
harsh-nod, line 45: Why is this function required? Don't the forward and backward slices go beyond the boundaries of the reduction and propagate from iter args to init args?

harsh-nod (commented):
Overall looks great! Thanks I think this will be a good foundation to build on. Just some minor comments, but otherwise looks good!

shark_turbine/kernel/wave/thread_shape_analysis.py
raikonenfnu: haha we don't have to, I guess because it already have the index size so just slightly redundant to set the index size again. But I think you're right, it's not very expensive to set index and would be cleaner without it. :)

shark_turbine/kernel/wave/thread_shape_analysis.py
raikonenfnu: backward slice does not go through from IterArg to init args. Shall we modify getInputs to do that instead(it's quite doable)? Since I am not the only user of bfs/backwardSlice I'd like to be mindful of modifiying too much  haha

shark_turbine/kernel/wave/thread_shape_analysis.py
raikonenfnu: Done!

shark_turbine/kernel/wave/thread_shape_analysis.py
raikonenfnu: Added more docs, lmk what you think! :)

shark_turbine/kernel/wave/thread_shape_analysis.py
raikonenfnu: done!

shark_turbine/kernel/wave/thread_shape_analysis.py
raikonenfnu: @harsh-nod I modified it, but not 100% sure this would work for other's use cases, would probably want your eyes on this one :)

shark_turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -57,6 +50,38 @@ def determine_thread_shapes(trace: CapturedTrace):
     3. We bucket these ops to Variadic(Index->elem_per_thread) mapping.
     4. At every bucket of (index -> elem_per_thread), we apply these information
        by updating their indexSequence size.
+
+    We stored the buckets above in a variable/dict called `thread_size_to_ops`.
```
harsh-nod, line 54: very nice! thanks!

shark_turbine/kernel/wave/thread_shape_analysis.py
harsh-nod: Yes I think that makes sense. Looks good to me!

shark_turbine/kernel/wave/thread_shape_analysis.py
harsh-nod: looks great, thanks!

harsh-nod (approved):
lgtm! once you get the tests passing :)

PR #187, opened 2024-10-03
[TKW] IGEMM Benchmarking

Hardcode84:
Initial version of IGEMM benchmarking.

* If `--runperf` pytest option is set, generate IREE ref code and run both TKW and ref code with `run_bench=True`
* Add `--dump-perf-files-path` option to save perf info files into provided directory (filenames based on test name and params)


shark_turbine/kernel/wave/iree_utils.py
```diff
@@ -23,6 +24,23 @@ def get_mmt_asm(lhs_type: str, rhs_type: str, acc_type: str) -> str:
     return matmul_function
 
 
+def get_conv_asm(
+    conv_type: str, lhs_type: str, rhs_type: str, res_type: str, stride: int
+) -> str:
+    res_dtype = res_type.split("x")[-1]
+    return f"""
+    func.func @conv_{conv_type}(%lhs: tensor<{lhs_type}>, %rhs: tensor<{rhs_type}>) -> tensor<{res_type}> {{
```
harsh-nod, line 32: Not related to this patch, but one way to enforce that iree using the "best" values for lowering pipeline etc would be to get the transform dialect scripts from the sdxl-scripts repo.

shark_turbine/kernel/wave/utils.py
```diff
@@ -388,16 +388,11 @@ def compile_and_invoke(
                 )
         else:
             for inp in kernel_inputs:
-                tf = tempfile.NamedTemporaryFile()
-                torch.save(inp, tf)
+                tf = tempfile.NamedTemporaryFile(suffix=".npy")
+                numpy.save(tf, inp.numpy())
```
harsh-nod, line 392: is this required because iree-benchmark-module expects np files?

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -940,6 +946,18 @@ def repeat(acc: tkl.Register[M, NF, tkl.f32]) -> tkl.Register[M, NF, tkl.f32]:
 
     config = {"backend": "rocm", "device": "hip", "target": "gfx942"}
 
+    run_bench = request.config.getoption("--runperf")
+    dump_perf = request.config.getoption("--dump-perf-files-path")
+    if run_bench:
+        config["benchmark_batch_size"] = 10
```
harsh-nod, line 952: One question here, how much difference do you see in performance between 10 and 1000?

shark_turbine/kernel/wave/utils.py
Hardcode84: Not sure what was broken exactly, benchmarking module wasn't throwing any errors, it was just returning empty benchmarking info.

tests/kernel/wave/wave_e2e_test.py
Hardcode84: If you look into perf dump, most tests only did few 10s of iterations, the fastest ones did 200-300, batch size 1000 will take literal hours.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -829,11 +830,16 @@ def repeat(acc: tkl.Register[M, NF, tkl.f32]) -> tkl.Register[M, NF, tkl.f32]:
     perf_test(2, 64, 64, 640, 3, 3, 640, 1),
 ]
 
+_mem_spaces = [
+    pytest.param(GLOBAL_ADDRESS_SPACE, id="global"),
+    pytest.param(SHARED_ADDRESS_SPACE, id="shared"),
+]
```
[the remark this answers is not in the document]
Hardcode84: It allows for tests to have proper names like `[shared-2-64-64-640-3-3-640-1]` instead of `[mem_space1-2-64-64-640-3-3-640-1]`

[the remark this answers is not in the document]
harsh-nod: nice!

harsh-nod (approved):
thanks!

PR #188, opened 2024-10-03
[TKW][CI] Specify rocm lib directory

raikonenfnu:


raikonenfnu:
Fixed locally by modifying:
```
sudo vim /etc/ld.so.conf.d/10-rocm-opencl.conf
# correct the path in the file above
sudo ldconfig #reload ld paths
```

PR #190, opened 2024-10-03
Set `fail-fast: false`

Hardcode84:
Our tests are flaky, `fail-fast: false` won't allow failing builds abort other.


harsh-nod (approved):
thanks!

PR #193, opened 2024-10-04
Add padding to reduce shared memory bank conflicts

harsh-nod:
This PR adds padding to shared memory allocations
to minimize bank conflicts.

### DIFF
--- a/shark_turbine/kernel/wave/promotion.py
+++ b/shark_turbine/kernel/wave/promotion.py
@@ -15,6 +15,25 @@ from .utils import subs_idxc
 logger = get_logger("turbine.wave.promotion")
 
 
+def apply_padding(
+    shape: tuple[IndexSymbol | int], dtype: DataType
+) -> tuple[IndexSymbol | int]:
+    """
+    When accessing shared memory, we need to be cognizant of bank conflicts
+    that can have a significant impact on performance. One way to mitigate
+    these conflicts is by applying padding to the shared memory allocation.
+    This function applies padding of 64 bits to the shared memory allocation.
+    While this approach accomplishes the goal of reducing bank conflicts, it
+    is inefficient in terms of memory usage. A more sophisticated approach
+    would involve swizzling of the shared memory access patterns.
+    """
+    padding = 64 // dtype.bitwidth()
+    return tuple(
+        value + padding if i == len(shape) - 1 else value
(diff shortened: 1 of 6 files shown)


shark_turbine/kernel/wave/promotion.py
```diff
@@ -15,6 +15,25 @@
...
+    When accessing shared memory, we need to be cognizant of bank conflicts
+    that can have a significant impact on performance. One way to mitigate
+    these conflicts is by applying padding to the shared memory allocation.
+    This function applies padding of 64 bits to the shared memory allocation.
+    While this approach accomplishes the goal of reducing bank conflicts, it
+    is inefficient in terms of memory usage. A more sophisticated approach
+    would involve swizzling of the shared memory access patterns.
+    """
+    padding = 64 // dtype.bitwidth()
+    return tuple(
```
raikonenfnu, line 31: what if it's already aligned?

shark_turbine/kernel/wave/promotion.py
```diff
@@ -15,6 +15,25 @@
...
+        value + padding if i == len(shape) - 1 else value
```
raikonenfnu, line 32: can we do something like `shape[-1] += padding`?

raikonenfnu (commented):
Overall looks great just couple minor NITs/Qs.

shark_turbine/kernel/wave/promotion.py
harsh-nod: For this approach, we just apply a blanket padding value that is independent  of whether the shape is aligned or not. We could also making this a tuning parameter to see which value gives the best performance for a given shape.

shark_turbine/kernel/wave/promotion.py
harsh-nod: Shape is a tuple so unfortunately not.

shark_turbine/kernel/wave/promotion.py
raikonenfnu: makes sense!

shark_turbine/kernel/wave/promotion.py
raikonenfnu: ah I see

raikonenfnu (approved):

PR #194, opened 2024-10-04
Add ability to dump intermediates

harsh-nod:
This PR adds a flag to dump intermediates which include .ll and .s files to see what instructions were
generated.

### DIFF
--- a/shark_turbine/kernel/wave/utils.py
+++ b/shark_turbine/kernel/wave/utils.py
@@ -334,6 +334,9 @@ def compile_and_invoke(
     if config.get("print_ir_after_all", False):
         flags.append("--mlir-print-ir-after-all")
 
+    if config.get("dump_intermediates", False):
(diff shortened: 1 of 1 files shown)


shark_turbine/kernel/wave/utils.py
```diff
@@ -334,6 +334,9 @@ def compile_and_invoke(
     if config.get("print_ir_after_all", False):
         flags.append("--mlir-print-ir-after-all")
 
+    if config.get("dump_intermediates", False):
+        flags.append("--iree-hal-dump-executable-intermediates-to=tmp/intermediates")
```
Hardcode84, line 338: It's better to get path from the config, I think, `config["dump_intermediate"] = "some/path"`

harsh-nod: sure I will change it so that if you specify "dump_intermediates", you also need to specify a path.

Hardcode84 (approved):

PR #195, opened 2024-10-04
Split TK CI from main CI

Hardcode84:
* Main CI is flaky, add a separate pipeline, which tests only TK as temp solution
* Make `pytest` output more verbose
* Remove unnecessary stuff from perf pipeline

### DIFF
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -16,7 +16,7 @@ concurrency:
 
 jobs:
   test:
-    name: "Unit Tests and Type Checking"
+    name: "[TKW] Unit Tests and Type Checking"
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -1,13 +1,10 @@
-name: PERF
+name: CI
 
 on:
-  workflow_dispatch:
   pull_request:
   push:
     branches:
     - main
-  schedule:
-  - cron: '30 5 * * *'
 
 concurrency:
   # A PR number if a pull request and otherwise the commit hash. This cancels
@@ -24,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         version: [3.11]
-        os: [nodai-amdgpu-mi300-x86-64]
+        os: [ubuntu-latest, nodai-amdgpu-mi300-x86-64]
     runs-on: ${{matrix.os}}
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -55,13 +55,13 @@ jobs:
...
       run: |
-        pytest -n 4 .
+        pytest -n 4 --capture=tee-sys -vv .
 
     - name: Run e2e tests on MI300
       if: "contains(matrix.os, 'mi300') && !cancelled()"
       run: |
         export WAVE_RUN_E2E_TESTS=1
-        pytest -n 4 --capture=tee-sys ./tests/kernel/wave/
+        pytest -n 4 --capture=tee-sys -vv ./tests/kernel/wave/
```
harsh-nod, line 64: Should we remove Run e2e tests on MI300 from ci.yaml since it is already in ci-tk.yaml?

.github/workflows/ci-tk.yaml
```diff
@@ -0,0 +1,74 @@
...
+      run: |
+        export WAVE_RUN_E2E_TESTS=1
+        pytest -n 4 --capture=tee-sys -vv ./tests/kernel/wave/
+
+    - name: Run LIT tests
+      if: ${{ !cancelled() }}
+      run: |
+        lit lit_tests/ -v
+
+    - name: MyPy Type Checking
```
harsh-nod, line 71: Can we remove mypy from here so it is only run once in ci.yaml?

.github/workflows/ci.yaml
Hardcode84: I'm not sure. My motivation to keep it was that in the bright future, when our CI is not flaky anymore, we can just delete `TK CI` pipeline file and won't need any changes to main CI.

.github/workflows/perf.yaml
```diff
@@ -54,21 +54,10 @@ jobs:
...
-        pytest -n 1 ./tests/kernel/wave/
-    - name: Run LIT tests
-      if: ${{ !cancelled() }}
-      run: |
-        lit lit_tests/ -v
-    - name: MyPy Type Checking
-      if: ${{ !cancelled() }}
-      run: |
-        mypy
+        pytest -n 1 --capture=tee-sys -vv ./tests/kernel/wave/
```
harsh-nod, line 63: Don't we need to add a --runperf flag here?

.github/workflows/ci-tk.yaml
Hardcode84: No, I want to keep lit and mypy tests here too, as if they fail in main CI, people can miss it as it fails every time anyway.

.github/workflows/perf.yaml
Hardcode84: We should, but let's do it in dedicated PR

.github/workflows/ci.yaml
harsh-nod: I would say we can still do that in the future, but instead of deleting tk ci, we can merge ci and tk ci, though there might be something to be said about compartmentalization of different parts of the codebase. (think changes in one folder only trigger tests related to that) I think since we are trying to save time lets remove as much extraneous work being done on the builders.

.github/workflows/ci-tk.yaml
harsh-nod: makes sense.

.github/workflows/ci.yaml
Hardcode84: OK, makes sense. Done.

harsh-nod (approved):
thanks for putting this together!

.github/workflows/ci-tk.yaml
ScottTodd: Static type checking (especially across the whole repository) could also go in a standalone "lint" workflow.

Flaky CI builds need to be fixed though. Ignoring "main CI" because it "fails every time" is extremely suspect.

PR #201, opened 2024-10-07
[TKW] Test multiple igemm layouts

Hardcode84:
* Test `nchw_fchw` and `nhwc_hwcf` igemm conv layouts.
* Perf test will use `nhwc_hwcf` as IREE seems to produce the best result for it.


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -984,6 +876,7 @@ def repeat(acc: tkl.Register[M, NF, tkl.f32]) -> tkl.Register[M, NF, tkl.f32]:
         run_bench=run_bench,
         run_config=config,
     ):
+        out = torch.zeros_like(out_ref)
```
harsh-nod, line 879: why is this needed?

harsh-nod (approved):
lgtm!

tests/kernel/wave/wave_e2e_test.py
Hardcode84: just moved the var closer to it usage

PR #202, opened 2024-10-07
[tests][aot] Add test for externalized parameters

vinayakdsci:
Progresses towards https://github.com/iree-org/iree/issues/18564.

Adds a test for the example given in [Using external parameters](https://iree.dev/guides/ml-frameworks/pytorch/#using-external-parameters) section in the IREE PyTorch guide.

### DIFF
--- a/tests/aot/params_test.py
+++ b/tests/aot/params_test.py
@@ -12,6 +12,9 @@ import unittest
 import torch
 import torch.nn as nn
 
+import iree.runtime as rt
+import iree.compiler as ireec
+
 from iree.turbine.aot import (
     export,
     externalize_module_parameters,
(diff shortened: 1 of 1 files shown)


kumardeepakamd (approved):
LG

monorimet (approved):
Thanks. Overall this looks OK, but I think in the future we might want more than just validating numerics.
A follow-up might break this down further or add more tests that ensure parameter externalization occurs as expected (make sure we don't have copies of buffers/parameters or leftover inlined parameters in the generated IR). These are nitpicks based on problems I've run into often on the SD modeling side. ( I believe some of these are covered in existing tests )

Also, If we're adding in tests for external parameters like so, perhaps a follow-up for AOT save/load on disk may cover more use cases.

The sort of problems I hope we validate a path for:

- saving and loading .safetensors or .irpa from a torch module to a parameter index
- intake of parameter files from common sources, ensure seamless plug-in of parameters to appropriately constructed compiled modules.

The motive behind these points are cases like Stable Diffusion deployment, where checkpoints are sourced by users and "plugged in" to a given inference solution. If we don't seamlessly intake those parameters, we either have to generate new IR for them or save a copy of parameters with modified parameter keys.

LGTM though. Thanks.

vinayakdsci:
@monorimet sure, I agree with all you say. I can add tests in follow up PRs too, for each of the cases that you mention.

PR #203, opened 2024-10-07
[TKW] Igemm opt

Hardcode84:


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -757,14 +757,14 @@ def test_igemm_conv(n, h, w, c, hf, wf, nf, stride, mem_space, layout, request):
         inputs={
             N: i // SZ_OUT,
             C: j // (HF * WF),
-            H: (i % SZ_OUT) % W_OUT * stride + (j % (HF * WF)) % WF,
-            W: (i % SZ_OUT) // W_OUT * stride + (j % (HF * WF)) // WF,
+            W: (i % SZ_OUT) % W_OUT * stride + (j % (HF * WF)) % WF,
```
harsh-nod, line 760: was this a bug? how were the tests passing if we had H and W switched?

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -803,10 +803,10 @@ def test_igemm_conv(n, h, w, c, hf, wf, nf, stride, mem_space, layout, request):
...
     # Expose user-constraints
     constraints: list[tkw.Constraint] = []
-    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
-    constraints += [tkw.WorkgroupConstraint(NF, BLOCK_N, 1)]
-    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
-    constraints += [tkw.WaveConstraint(NF, BLOCK_N)]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
+    constraints += [tkw.WorkgroupConstraint(NF, BLOCK_N, 0)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M / 2)]
+    constraints += [tkw.WaveConstraint(NF, BLOCK_N / 2)]
```
harsh-nod, line 809: If you update the wave constraint, you also have to update the waves_per_block to be (2, 2, 1)

tests/kernel/wave/wave_e2e_test.py
Hardcode84: This changes how we unroll conv window (rows vs colums first), it should produce semantically equivalent result as long as it synchronized between all mapping. But it also affects memory access pattern and this way it's slightly faster.

tests/kernel/wave/wave_e2e_test.py
Hardcode84: done

PR #206, opened 2024-10-08
Add preliminary tikz overview figure 

martin-luecke:
Adds preliminary tikz overview figure and an initial description of the fx tracing process


harsh-nod (approved):
looks good! thanks!

PR #208, opened 2024-10-10
[tkw] Add figure to outline the expansion process

martin-luecke:
The figure consists of the graphs of a traced gemm before and after expansion.
I started coloring the edges depending in which step of the expansion the corresponding nodes are created to aid in explaining the expansion process.
This still needs the coloring for the last steps of the expansion of the reduction dimension.


harsh-nod (approved):
looks nice!

PR #211, opened 2024-10-11
Add initial plot for attention

erman-gurses:


iree/turbine/kernel/wave/docs/mlsys/scripts/plot_results_attention.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+
+import matplotlib.pyplot as plt
+import seaborn as sns
+import numpy as np
+
+sns.set_theme()
+
+shapes = (
+    "4.0\n16.0\n1024.0\nd128",
```
harsh-nod, line 10: I would remove the d, and can you combine the first 2 numbers into one number, so instead of 4x16 just 64, also replace \n with a x

erman-gurses: Per our conversation, we keep the `\n` to avoid mixing the numbers on the X-axis.

PR #216, opened 2024-10-14
[TKW] Make shared mem read/writes using non masked ops

Hardcode84:
Add a pass which adjusts shared mem reads/writes and MMA indices to never had to partial accesses. This is need to remove masking ops on shared mem.

Also, update `minimize_global_loads.py` to use ceildiv instead if floor div in case of unaligned sizes.

### DIFF
--- a/iree/turbine/kernel/wave/utils.py
+++ b/iree/turbine/kernel/wave/utils.py
@@ -271,9 +271,18 @@ def remove_global_indexing(
     """
     tiling_constraints = [c for c in constraints if isinstance(c, TilingConstraint)]
     workgroup_ids = [WORKGROUP_0, WORKGROUP_1, WORKGROUP_2]
-    new_index = {key: index[key].subs({w: 0 for w in workgroup_ids}) for key in index}
+    subs = {w: 0 for w in workgroup_ids}
+
+    key_subs = {
+        c.dim: (c.count * c.tile_size)
+        for c in constraints
+        if isinstance(c, (TilingConstraint, WorkgroupConstraint))
+        and subs_idxc(c.dim) != subs_idxc(c.count * c.tile_size)
+    }
+
+    new_index = {safe_subs(key, key_subs): safe_subs(index[key], subs) for key in index}
     for key in new_index:
-        for constraint in tilingConstraints:
+        for constraint in tiling_constraints:
             new_index[key] = new_index[key].subs({constraint.induction_var: 0})
     return new_index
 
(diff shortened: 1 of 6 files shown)


iree/turbine/kernel/wave/utils.py
```diff
@@ -262,21 +262,44 @@ def get_hardware_vector_map(constraints: list[Constraint]) -> dict[IndexSymbol,
...
+    This function aligns index vars with Workgroup/Tiling constraints so it never
+    need partial reads/writes.
+    """
+    key_subs = {
+        c.dim: (c.count * c.tile_size)
+        for c in constraints
+        if isinstance(c, (TilingConstraint, WorkgroupConstraint))
+        and subs_idxc(c.dim) != subs_idxc(c.count * c.tile_size)
+    }
+    return {safe_subs(key, key_subs): index[key] for key in index}
```
harsh-nod, line 300: So this is interesting. Stepping through the code, it seems like the original index 
```
{M: $T0 + BLOCK_M*floor($T0/64), N: $T1*BLOCK_N + 4*$T1 : 4 : 1}
```
would get transformed to 
```
{BLOCK_M*ceiling(M/BLOCK_M): $T0 + BLOCK_M*floor($T0/64), BLOCK_N*ceiling(N/BLOCK_N): $T1*BLOCK_N + 4*$T1 : 4 : 1}
```
But the problem with this is that, downstream passes and transformations can no longer query the original dimensions "M" and "N". Is there a way we can avoid modifying the dimension names?

Hardcode84: Yes, this is true, and it was causing problems in thread shape analysis. And that's why I'm doing it almost at the very end of the pipeline.

harsh-nod: I think that makes sense, but I am wondering whether it is necessary. So why do you need to rewrite M -> BLOCK_M * ceiling(M / BLOCK_M) ? How is that used in the codegen?

harsh-nod: Couldn't you modify the value of M in the IndexingContext ?

Hardcode84: `find_index_bounds` compares dim with aligned WG/Tilied size to determine if op need masking. We need original dim value to make making on global mem.

Hardcode84: added a comment

harsh-nod (approved):
I would like to explore other ways to surface this information to earlier transformations, but this is good for now. thanks!

PR #218, opened 2024-10-14
[TKW] Add overflow flags to index calculations

Hardcode84:
Index calculations should never overflow. Add corresponding flags to `arith.addi/muli`.


harsh-nod (approved):
lgtm! thanks!

PR #219, opened 2024-10-15
Add support for more MFMA intrinsics

harsh-nod:
This PR adds support for the 32x32x8 F16 intrinsic, as well as the 16x16x32 and 32x32x16 F8 intrinsics.


iree/turbine/kernel/wave/scheduling/loop_reconstruction.py
```diff
@@ -283,9 +283,11 @@ def push_rotating_registers(
                     f"Mapped orig: {node_map[node]} / mapped: {iter_arg} to stage {mapped_stage}."
                 )
             else:
+                mapped_stage = stage + len(registers) - i - 1
```
raikonenfnu, line 286: are the changes in this file related to adding of new mfma intrinscis?

harsh-nod: So that one is not. There are basically 2 types of changes in this PR:
1. Adding new intrinsics
2. Fixes to scheduling to get scheduling to work with the new intrinsics

raikonenfnu: I understand how adding casts need changes in the graph_util and resources, but not completely sure how the new intrinsics would incite these changes in loop_reconstruction. Can you give a little more context here?

harsh-nod: So adding the new intrinsic helped uncover a bug in the scheduling code. The issue was that when we push rotating registers into the argument context, the rotating registers need to be pushed into a different stage when emitting the loop body vs the epilogue (prologue doesn't matter because we don't have rotating registers). As a result of this not all the results from the loop body were being used in the epilogue but with the fix, seems to be working now.

raikonenfnu: ah nice! :)

tests/kernel/wave/wave_gemm_test.py
```diff
@@ -51,10 +52,43 @@ def get_test_shapes(test_name: str) -> list[tuple[int]]:
     return default_test_shapes[test_name]
 
 
+def get_load_elems_per_thread(mfma_variant: MMAType) -> int:
```
raikonenfnu, line 55: Seems useful in life, maybe we cna move this into utils?

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -80,25 +81,42 @@ def has_strided_access(node: fx.Node) -> bool:
...
+            key=lambda item: item[1],
+        )
         with custom.graph.inserting_before(operator):
             for i in range(elements_per_thread):
+                # Non-contiguous access patterns can have varying offsets. We
+                # handle that here.
+                gpr_offset = [
+                    expr
+                    for expr in simplified_index[max_stride_dim].start.args
+                    if expr.has(GPR_NUM)
```
raikonenfnu, line 97: just curious wouldn't the indexSequence.start.arg be singular symbols? so wouldn't `expr == GPR_NUM` be more apt here?

raikonenfnu: or would expr be:
```
[ (8 * floor(GPR_NUM / 4) % 32), 4 * floor(lane / 32), (GPR_NUM % 4)] ?
 ```

File: iree/turbine/kernel/wave/index_sequence_analysis.py [hunk 4]
```diff
@@ -80,25 +81,42 @@ def has_strided_access(node: fx.Node) -> bool:
...
+                ]
+                if not gpr_offset:
+                    gpr_offset = i
+                else:
+                    gpr_offset = sympy.Add(*gpr_offset).subs({GPR_NUM: i})
```
raikonenfnu, line 102: IIUC, we are trying to sub in `GPR_NUM: i` on simplified_index[max_stride_dim].start, BUT without the lane information. Wouldn't be more accurate with the lane info i,e (`4 * floor(lane / 32)`)? if that is not the case, can you explain why not? 🤔

tests/kernel/wave/wave_gemm_test.py
harsh-nod: sure, will move into utils.

iree/turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: The idea is to separate the offset into a gpr_offset and lane_offset. We compute the gpr_offset here and add it to the lane offset later.

iree/turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: You are right. We need both the lane offset and gpr offset. The lane offset is added later when computing write.index below.

tests/kernel/wave/wave_gemm_test.py
harsh-nod: done!

raikonenfnu (approved):

PR #220, opened 2024-10-15
WIP Add in-place variant of transfer_to_logical_device

sogartar:
The purpose of this is to insert IREE ABI affinity attributes for function arguments.
iree.flow.transfer is not sufficient for cases of in-place updates of function arguments.

An alternative approach is to set the attribute of a Torch FX graph function argument. Then during/after conversion to MLIR to set the corresponding attribute on the MLIR function.

### DIFF
--- a/iree/turbine/ops/iree.py
+++ b/iree/turbine/ops/iree.py
@@ -6,13 +6,19 @@
 
 """Custom ops for built-in IREE functionality."""
 from typing import cast
+import torch
 
 from ..support.ir_imports import (
     Attribute,
+    ArrayAttr,
+    Block,
+    BlockArgument,
+    DictAttr,
     RankedTensorType,
     StringAttr,
     Value,
     flow_d,
+    func_d,
     tensor_d,
 )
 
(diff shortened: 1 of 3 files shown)


sogartar:
The output IR in the test will look like this

```mlir
  func.func @main(%arg0: !torch.tensor<[9,8],f32> {iree.abi.affinity = #hal.device.promise<@__device_1>}) -> !torch.vtensor<[9,8],f32> attributes {torch.assume_strict_symbolic_shapes} {
    %0 = torch.copy.to_vtensor %arg0 : !torch.vtensor<[9,8],f32>
    %1 = torch_c.to_builtin_tensor %0 : !torch.vtensor<[9,8],f32> -> tensor<9x8xf32>
    %2 = torch_c.from_builtin_tensor %1 : tensor<9x8xf32> -> !torch.vtensor<[9,8],f32>
    %int1 = torch.constant.int 1
    %int1_0 = torch.constant.int 1
    %3 = torch.aten.add.Scalar %2, %int1, %int1_0 : !torch.vtensor<[9,8],f32>, !torch.int, !torch.int -> !torch.vtensor<[9,8],f32>
    torch.overwrite.tensor.contents %3 overwrites %arg0 : !torch.vtensor<[9,8],f32>, !torch.tensor<[9,8],f32>
    return %3 : !torch.vtensor<[9,8],f32>
  }
```

rsuderman:
We should try to include execution based testing. Having a simple script that performs the transfer, does some simple math, then transfers back running via iree would tell us more about functionality. E.g. IR is good, execution is better.

PR #222, opened 2024-10-16
[TKW] Refactor gather indices generation

Hardcode84:
Previously, we could end generating dynamic indices both for the gather `start_indices` and `offsets_vec` which is suboptimal.

Now, detect if we need dynamic `offsets_vec`, and generate `start_indices` as 0 in this case, encoding entire index calculation in `offsets_vec`.

### DIFF
--- a/iree/turbine/kernel/wave/codegen.py
+++ b/iree/turbine/kernel/wave/codegen.py
@@ -639,9 +639,16 @@ def _construct_gather_scatter_indices(
 
     offsets_vec_type = VectorType.get([elements_per_thread], IndexType.get())
     if need_dynamic_offsets:
+        # In case we need dynamic `offsets_vec`, set all `start_indices` to 0
+        # and encode entire index info in `offsets_vec`.
         result_index = {key: 0 for key in symbolc_shape}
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/wave/codegen.py
```diff
@@ -626,22 +632,29 @@ def _construct_gather_scatter_indices(
...
-    )
-
-    for i, off in dynamic_offsets:
-        pos = arith_d.ConstantOp(IndexType.get(), i)
-        offsets_vec = vector_d.insertelement(off, offsets_vec, position=pos)
+    if need_dynamic_offsets:
+        result_index = {key: 0 for key in symbolc_shape}
+        start_indices = _build_start_indices(emitter, result_index)
+        subs = [(sym, idx) for sym, idx in zip(iters.keys(), start_indices_orig)]
+        subs[-1] = (
```
harsh-nod, line 645: I understand what is happening here, but it's a little hard to read with the [-1][0] etc. Can you add a comment explaining this? (my understanding is you are adding [0,1,2,3] to the innermost dimension) ?

harsh-nod (approved):
lgtm! just one request for a comment, thanks

PR #223, opened 2024-10-17
Add op to cast between dtypes

harsh-nod:
This PR adds an op that can cast between
integer and float types of different bitwidths.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -1160,7 +1160,7 @@ class ShuffleOp(CustomOp):
         return get_custom(self.arg).indexing_dims
 
     @property
-    def type(self) -> Memory:
+    def type(self) -> Register:
         src_type = get_custom(self.arg).type
         return src_type
 
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -123,6 +123,10 @@ def shuffle(src: "Register", offset: int, width: int) -> "Register":
     ...
 
 
+def cast(src: "Register", dtype: DataType) -> "Register":
+    ...
+
+
 def define_op(op_name: str) -> Callable[[T], T]:
     def decorator(cls: T) -> T:
         cls.tkw_op_name = op_name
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1159,3 +1163,23 @@ def indexing_dims(self) -> list[IndexSymbol]:
...
+
+    arg: fx.Node
+    dtype: DataType
+
+    @property
+    def indexing_dims(self) -> list[IndexSymbol]:
+        return get_custom(self.arg).indexing_dims
+
+    @property
+    def type(self) -> Memory:
```
Hardcode84, line 1183: `-> Register`

Hardcode84 (approved):

iree/turbine/kernel/wave/codegen.py
```diff
@@ -1198,3 +1200,64 @@ def handle_get_result(emitter: WaveEmitter, node: fx.Node):
...
+
+    cast_ops = {
+        (True, True): arith_d.extf,
+        (True, False): arith_d.extsi,
+        (False, True): arith_d.truncf,
+        (False, False): arith_d.trunci,
+    }
+
+    dtype = (
+        get_float_type(dst_elem_type.width)
```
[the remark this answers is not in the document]
raikonenfnu: actually misrread sorry

iree/turbine/kernel/wave/codegen.py
```diff
@@ -1198,3 +1200,64 @@ def handle_get_result(emitter: WaveEmitter, node: fx.Node):
...
+        if is_dst_float
+        else IntegerType.get_signless(dst_elem_type.width)
+    )
+    converted_vector = conversion_ops[(is_src_float, is_dst_float)](
+        VectorType.get(src_vector_type.shape, dtype), vector_src
```
raikonenfnu, line 1253: why can't we use dst_elem_type directly instead of `dtype`?

iree/turbine/kernel/wave/codegen.py
```diff
@@ -1198,3 +1200,64 @@ def handle_get_result(emitter: WaveEmitter, node: fx.Node):
...
+        (True, True): arith_d.extf,
+        (True, False): arith_d.extsi,
+        (False, True): arith_d.truncf,
+        (False, False): arith_d.trunci,
+    }
+
+    dtype = (
+        get_float_type(dst_elem_type.width)
+        if is_dst_float
+        else IntegerType.get_signless(dst_elem_type.width)
```
raikonenfnu, line 1250: don't we want src_elem_type.width for the intermediate cast between float and int?

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: thanks, fixed.

iree/turbine/kernel/wave/codegen.py
harsh-nod: yes, good catch. fixed.

iree/turbine/kernel/wave/codegen.py
raikonenfnu: nvm

harsh-nod: good point, I was under the impression that the bit width has to be the same for the conversion ops but that is not true. Fixed.

PR #224, opened 2024-10-17
Perf test experiments

Hardcode84:


.github/workflows/perf.yaml
```diff
@@ -19,7 +16,8 @@ concurrency:
 
 jobs:
   test:
-    name: "Unit Tests and Type Checking"
+    name: "Perf Tests"
+    if: ${{ github.event.issue.pull_request }} && contains(github.event.comment.body, '/turbine_run_perf')
```
ScottTodd, line 21: You may want to look for a label being applied or validate the author of the comment somehow. This looks like any GitHub user (project member or not) could trigger the workflow by adding this comment. That increases the surface area for malicious behavior.

Hardcode84: I doesn't work anyway yet :)

Hardcode84:
turbine_run_perf

.github/workflows/perf.yaml
Hardcode84: I will just probably create separate personal repo and experiment on it first

ScottTodd: https://github.com/iree-org/iree/blob/7d60397cbe62e8563a36ec286c41aa40d7e4e58d/.github/workflows/benchmark_trigger.yml might also be useful to reference (I'm sure there are other good references in other github projects, but that's what we used in iree-org/iree for a while)

PR #229, opened 2024-10-18
Restrict to torch < 2.5.0 and change channel

marbre:
The CPU requirements restrict torch to a version < 2.5.0, whereas the ROCm requirements did not so far. The current iree-turbine is incompatible with torch 2.5.0 as `torch.export.dynamic_dim()` was removed with pytorch/pytorch@b454c51. Version 2.4 releases for ROCm 6.0 are available via a different channel, changing that accordingly.

### DIFF
--- a/pytorch-rocm-requirements.txt
+++ b/pytorch-rocm-requirements.txt
@@ -1,5 +1,5 @@
 --pre
---index-url https://download.pytorch.org/whl/nightly/rocm6.0
-torch>=2.3.0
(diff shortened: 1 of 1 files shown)


marbre:
Instead of pointing to https://download.pytorch.org/whl/rocm6.0, we could also point to https://download.pytorch.org/whl/rocm6.1 but the previously set URL pointed to ROCm 6.0.

Hardcode84 (approved):

PR #231, opened 2024-10-18
Add device affinities for arguments in AOT

sogartar:
We don't have support for providing device affinities for function arguments, which need to end up as MLIR function argument attributes.

This change adds a class DeviceAffinity and provides the ability to supply affinities when exporting Torch functions/modules or when tracing in IREE-Trubine itself.

### DIFF
--- a/iree/turbine/aot/compiled_module.py
+++ b/iree/turbine/aot/compiled_module.py
@@ -44,6 +44,8 @@ from .support.ir_utils import (
     ModuleBuilderOptions,
 )
 
+from .tensor_traits import DeviceAffinity
+
 
 __all__ = [
     "CompiledModule",
@@ -107,12 +109,27 @@ class PyOnlyDef:
         return self.py_value(*args, **kwargs)
 
 
+class ExportTargetDef:
+    def __init__(
+        self,
+        target: Union[Callable, ExportedProgram],
+        *,
+        arg_device: dict[int, DeviceAffinity] | None = None,
+    ):
+        self.target = target
+        self.arg_device = arg_device
+
+    def __call__(self, *args, **kwargs):
+        return self.target(*args, **kwargs)
+
+
 class ExportProcDef:
     __slots__ = [
         "callable",
         "export_name",
         "signature",
         "file_line_loc",
+        "arg_device",
     ]
 
     def __init__(
@@ -122,14 +139,22 @@ class ExportProcDef:
         *,
         signature,
         file_line_loc: Optional[Tuple[str, int]] = None,
+        arg_device: dict[int, DeviceAffinity] | None = None,
     ):
         self.export_name = export_name
         self.callable = callable
         self.signature = signature
         self.file_line_loc = file_line_loc
+        self.arg_device = arg_device
 
     def copy(self) -> "ExportProcDef":
-        return ExportProcDef(self.export_name, self.callable, signature=self.signature)
+        return ExportProcDef(
+            self.export_name,
+            self.callable,
+            signature=self.signature,
+            file_line_loc=self.file_line_loc,
+            arg_device=self.arg_device,
+        )
 
     def __repr__(self):
         return f"<def {self.export_name}({self.signature})>"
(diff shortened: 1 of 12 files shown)


iree/turbine/aot/compiled_module.py
```diff
@@ -107,12 +112,27 @@ def __call__(self, *args, **kwargs):
         return self.py_value(*args, **kwargs)
 
 
+class ExportTargetDef:
```
sogartar, line 115: I am not sure if I should drop the `ExportTargetDef` and use half-initialized ExportProcDef and ExportedProgramDef directly.

sogartar:
This PR removes the need for #220, which kind of abuses the function generation mechanism.

stellaraccident:
Thanks. I will review this in a few minutes

iree/turbine/aot/compiled_module.py
dan-garvey: What is the benefit to having this separate class?

sogartar: In the place where we need a structure to store this data we are not ready to construct ExportProcDef or ExportedProgramDef. If we are to use them directly they will have a more complicated multi-step initialization.

iree/turbine/aot/compiled_module.py
```diff
@@ -207,6 +240,21 @@ def globals_defs(self) -> Generator[Tuple[str, GlobalsDef], None, None]:
         )  # type: ignore
 
     def def_attribute(self, key, value):
+        if isinstance(value, ExportTargetDef):
+            if isinstance(value.target, ExportedProgram):
+                value = ExportedProgramDef(
```
rsuderman, line 245: Shouldn't this be returned as well? It is setting the value but this falls through. I see it is handled further down, if so we should handle it right before the follow up case.

iree/turbine/aot/exporter.py
```diff
@@ -49,6 +50,21 @@
 SaveableTarget = Union[str, Path, None, Output]
 
 
+class DeviceAffinity:
+    """This is used to provide device affinities to exported function arguments."""
+
+    def __init__(self, moniker: str):
```
rsuderman, line 56: Make the moniker an int. All cases around just specify it that way anyway.

iree/turbine/aot/compiled_module.py
```diff
@@ -568,6 +627,22 @@ def save_mlir(inst: "CompiledModule", path: Union[Path, str]):
 
     jittable = staticmethod(builtins.jittable)
 
+    @staticmethod
+    def annotate(
```
rsuderman, line 631: Name this something better than `annotate`. Even `signature_info` would be sufficient to explain that its adding additional information to the function signature.

rsuderman (changes requested):

iree/turbine/aot/compiled_module.py
```diff
@@ -107,12 +112,27 @@ def __call__(self, *args, **kwargs):
...
 
+class ExportTargetDef:
+    def __init__(
+        self,
+        target: Union[Callable, ExportedProgram],
+        *,
+        argument_device_affinities: dict[int, "DeviceAffinity"] | None = None,
+    ):
+        self.target = target
+        self.argument_device_affinities = argument_device_affinities
```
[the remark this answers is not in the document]
sogartar: Done.

iree/turbine/aot/compiled_module.py
```diff
@@ -633,6 +708,7 @@ def __new__(
                 ep_def.exported_program,
                 symbol_name=ep_def.export_name or "main",
                 symbol_visibility=None if ep_def.public else "private",
+                argument_device_affinities=ep_def.argument_device_affinities or {},
```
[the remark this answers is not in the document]
sogartar: Done.

tests/aot/fx_programs_test.py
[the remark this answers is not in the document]
sogartar: I moved the new test to a new file as you suggested.

iree/turbine/aot/compiled_module.py
sogartar: I moved this case down before the handling of `ExportedProgramDef`.

iree/turbine/aot/exporter.py
sogartar: I changed it to `ordinal: int`.

stellaraccident (approved):
Thanks. A minor comment on code organization.

iree/turbine/aot/exporter.py
```diff
@@ -49,6 +50,21 @@
 SaveableTarget = Union[str, Path, None, Output]
 
 
+class DeviceAffinity:
```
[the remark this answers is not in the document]
sogartar: I moved it there. Why I did not see this.

sogartar:
I squashed and rebased to prepare for merging.

rsuderman (approved):

PR #233, opened 2024-10-18
Remove deprecated constraints and dynamic dim support with test fixes

IanNod:
The backing APIs have been deprecated for a year and were removed from PyTorch 2.6 nightlies, causing import errors.

torch.export and torch-mlir have long since provided a supported way for exporting dynamic dims.


saienduri:
With this, we can remove the torch restriction here: https://github.com/iree-org/iree-turbine/blob/main/pytorch-rocm-requirements.txt#L3

saienduri (approved):

PR #234, opened 2024-10-21
[TKW] Add support for multiple/local reduceOp

raikonenfnu:
In order to support flash attention, we'd need to be able to expand ReduceOps in the reduction dimension as well. We will do this by expanding the source of ReduceOp and locally reduce all of them. In that effort, we introduce this PR(1st out of 2) that add support of locally reducing over multiple variables.  

The second PR on the way would be expansion of ReduceOp.


lit_tests/kernel/wave/codegen.py
```diff
@@ -1336,6 +1336,68 @@ def test(
...
+            ADDRESS_SPACE: tkl.AddressSpace.GLOBAL_MEMORY.value,
+        },
+        canonicalize=True,
+    ):
+        print(test(a, b, c).module_op)
+        # CHECK: %[[LHS:.+]] = vector.load {{.*}} : memref<256x128xf16
+        # CHECK: %[[RHS:.+]] = vector.load {{.*}} : memref<256x128xf16
+        # CHECK: %[[LHS0:.+]] = vector.extract_strided_slice %[[LHS]] {offsets = [0], sizes = [1], strides = [1]}
+        # CHECK: %[[LHS1:.+]] = vector.extract_strided_slice %[[LHS]] {offsets = [1], sizes = [1], strides = [1]}
+        # CHECK: %[[REDUC_0:.+]] = arith.addf %[[LHS0]], %[[LHS1]] : vector<1xf16>
```
harsh-nod, line 1394: Why can't we do 
```
%res0 = arith.addf %lhs, %rhs : vector<2xf16>
%slice0 = vector.extract_strided_slice %res0 {offsets = [0], ...}
%slice1 = vector.extract_strided_slice %res0 {offsets = [1], ...}
%res1 = arith.addf %slice0, %slice1 : vector<1xf16>
```
Then you end up with 2 addf instead of 3?

raikonenfnu: good point, done!

iree/turbine/kernel/wave/decompose_reduce_ops.py
```diff
@@ -71,11 +77,12 @@ def decompose_reduce_ops(
 ):
     """
     The lowering for multi_reduction is done in two steps:
-      1. Local Reduce: Each thread reduces all elements carried by it along
+      1. Source Reduce: Each thread reduce locally all it's sources.
+      2. Local Reduce: Each thread reduces all elements carried by it along
          the reduction dimensions.
-      2. Thread Reduce: Each thread reduces result of step 1 across threads
+      3. Thread Reduce: Each thread reduces result of step 1 across threads
```
harsh-nod, line 83: should this be step 2?

iree/turbine/kernel/wave/decompose_reduce_ops.py
```diff
@@ -37,17 +37,23 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph):
     return custom
 
 
+def emit_sources_reduction(
+    binary_fn: Callable, src: list[fx.Node], graph: fx.Graph
+) -> fx.Node:
+    init = src[0]
+    for i in range(1, len(src)):
+        init = get_graph_node(binary_fn(init, src[i]), graph)
+    init.index = src[0].index
```
harsh-nod, line 46: are the indices the same for all src[i] ?

harsh-nod (approved):
lgtm! thanks!

iree/turbine/kernel/wave/decompose_reduce_ops.py
harsh-nod: I mean in the comment, should it be
```
 3. Thread Reduce: Each thread reduces result of step 2 across threads
```

lit_tests/kernel/wave/codegen.py
harsh-nod: thanks!

iree/turbine/kernel/wave/decompose_reduce_ops.py
raikonenfnu: Ahhh you're right! :)

PR #238, opened 2024-10-22
Add support for chained matmuls

harsh-nod:
This PR adds support for chained matmuls. In
order to do this, we implement the following
- Support for op-specific vector_shapes
- Support for op-specific dim expansions (reduction
  dims of mmas will now always have chained expansions,
  even if they are not the induction variable of
  reduction ops)
- Adds a permute operator
- Adds anchor and vector_shapes to each op
- Adds reduction_dim to MMA ops

### DIFF
--- a/iree/turbine/kernel/wave/expansion.py
+++ b/iree/turbine/kernel/wave/expansion.py
@@ -117,10 +117,7 @@ def expand_graph(
     Create a graph that represents the expanded version of the wave function.
     The expansion is done in the dimensions specified by the constraints.
     """
-    if isinstance(constraints_or_scaling, dict):
-        dim_scaling = constraints_or_scaling
-    else:
-        dim_scaling = get_dim_scaling(constraints_or_scaling)
+    get_node_dim_scaling = partial(get_dim_scaling, constraints_or_scaling)
 
     # Start from the back and expand in the corresponding indexing dimensions of a node
     # Then proceed to the operands
@@ -146,6 +143,7 @@ def expand_graph(
         if node.__class__ not in leaf_nodes:
             continue
 
+        dim_scaling = get_node_dim_scaling(node)
         for dim_combination in get_dim_combinations(dim_scaling, node.indexing_dims):
             expand_dims = {
                 dim: val for dim, val in zip(dim_scaling.keys(), dim_combination)
@@ -157,6 +155,7 @@ def expand_graph(
                 expand_dims,
                 dim_scaling,
                 expansion_context,
+                get_node_dim_scaling,
             )
 
 
@@ -166,6 +165,7 @@ def _expand_node(
     dim_query: dict[IndexSymbol, int],
     dim_scaling: dict[IndexSymbol, int],
     context: ExpandedNodeMap,
+    get_node_dim_scaling: Callable[[fx.Node], dict[IndexSymbol, int]],
     res_idx: int = 0,
 ) -> CustomOp:
     """Expand a single node or list of nodes in specific dimensions and recursively proceed to its inputs."""
@@ -177,8 +177,9 @@ def _expand_node(
                     elem,
                     trace,
                     dim_query,
-                    dim_scaling,
+                    get_node_dim_scaling(elem),
                     context,
+                    get_node_dim_scaling,
                     res_idx,
                 ).fx_node
             )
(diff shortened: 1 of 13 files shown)


iree/turbine/kernel/wave/expansion.py
```diff
@@ -187,8 +188,25 @@ def _expand_node(
     if (node, get_indexed_dims(dim_query, node), res_idx) in context:
         logger.debug(f"Already expanded node: {node} in {dim_query}")
         return context[(node, get_indexed_dims(dim_query, node), res_idx)]
+    elif isinstance(node, MMA):
+        # Handle expansion of MMA nodes whose reduction dim is not the same as the reduction
+        # dim of the parent reduction op.
```
harsh-nod, line 193: Handle the case where there is an mma without any reduction.

iree/turbine/kernel/wave/expansion.py
```diff
@@ -230,15 +248,34 @@ def _expand_node(
...
-            new_arg = _expand_node(
-                arg,
-                trace,
-                restricted_dims,
-                dim_scaling,
-                context,
-                res_idx,
-            )
+        new_arg = None
+        if not isinstance(arg, Sequence):
```
harsh-nod, line 252: if not sequence arg = [arg]

iree/turbine/kernel/wave/expansion.py
```diff
@@ -325,13 +367,107 @@ def _expand_reduction(
...
+    logger.debug(f"Expanding MMA reduction: {mma} in dims: {dim_query}")
+    idxc = IndexingContext.current()
+    for dim in mma.indexing_dims:
+        if dim not in dim_scaling and mma.vector_shapes[dim] > 0:
+            tile_size = idxc.get_static_value(dim)
+            dim_scaling[dim] = tile_size // mma.vector_shapes[dim]
+
+    expand_dims = set(mma.indexing_dims) - set([mma.reduction_dim])
+
+    # Store the original mma node and accumulator value for expansion.
```
harsh-nod, line 411: Explain this a little bit more as to why it is required.

iree/turbine/kernel/wave/expansion.py
harsh-nod: done

iree/turbine/kernel/wave/expansion.py
harsh-nod: done

iree/turbine/kernel/wave/expansion.py
harsh-nod: done

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -124,6 +129,7 @@ def set_node_indices(trace: CapturedTrace, constraints: list[Constraint]):
     mma_index, mma_slices = get_mma_dimensional_mapping(
         trace, get_hardware_constraint(constraints)
     )
+    trace.walk(partial(set_vector_shapes, constraints, mma_index, mma_slices))
```
raikonenfnu, line 132: No Action required: in the future perhaps we can do `set_vector_shapes` and `set_node_index` in a single walk. (still keeping the two functions apart though)

harsh-nod: good point. will keep that for a future PR.

iree/turbine/kernel/wave/utils.py
```diff
@@ -219,18 +219,30 @@ def is_mma(node):
...
-        hardware_constraint.vector_shapes[m] = M
-        hardware_constraint.vector_shapes[n] = N
-        hardware_constraint.vector_shapes[k] = K
-
-    return mapping, capture_mma_slices([get_custom(x) for x in mma_nodes])
+        if custom not in mapping:
+            mapping[custom] = {}
+        mapping[custom][m] = 0
+        mapping[custom][n] = 1
+        mapping[custom][k] = 2
```
raikonenfnu, line 226: No actions for now: Would be nice in the future to have an enum to represent [0, 1, 2] as [M, N, K] for better legibiltiy

harsh-nod: makes sense, will add that as well.

PR #240, opened 2024-10-24
Add `DeviceTensorTrait` to export annotations

rsuderman:
Exported globals can now be parked to belong to a specific device. This works with sharding tooling to ensure globals are not misassigned to the default affinity device


sogartar (approved):
We are inconsistent with the `DeviceAffinity` being an ordinal and this is using a the whole device name directly.

rsuderman:
> We are inconsistent with the `DeviceAffinity` being an ordinal and this is using a the whole device name directly.

Addressed

PR #241, opened 2024-10-25
Modify the shuffle op to use information from the index sequence

harsh-nod:
This PR modifies the shuffle op global reduction so that it determines the number of threads and the stride between them from the reduction source's index.

### DIFF
--- a/iree/turbine/kernel/wave/index_sequence_analysis.py
+++ b/iree/turbine/kernel/wave/index_sequence_analysis.py
@@ -23,6 +23,7 @@ from .utils import (
     get_hardware_constraint,
     subs_idxc,
     specialize_index_sequence,
+    capture_backward_slice,
 )
 import torch.fx as fx
 import numpy as np
@@ -238,6 +239,11 @@ def set_node_index(
     dimensions M, N and K, but allow each mapping to be piecewise conditioned
     on the operand.
     """
+    custom = get_custom(node)
+    anchor = custom.anchor
+    if isinstance(custom, (Reduction, Placeholder)) and not isinstance(custom, IterArg):
+        return
+
     hardware_constraint = [get_hardware_constraint(constraints)]
     workgroup_constraints = {
         c.dim: c for c in constraints if isinstance(c, WorkgroupConstraint)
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -242,13 +248,16 @@ def set_node_index(
...
-
-    if isinstance(custom, (Reduction, Placeholder)) and not isinstance(custom, IterArg):
-        return
+    # For elementwise operations that do not have an elements per thread attribute,
+    # look back to the backward slice to see if they can find an appropriate value.
+    if elements_per_thread is None:
+        backward_slice = capture_backward_slice(node)
+        for bwd_node in backward_slice:
+            custom_node = get_custom(bwd_node)
+            elements_per_thread = getattr(custom_node, "elements_per_thread", None)
```
raikonenfnu, line 258: shouldn't this be taken care of in thread shape analysis?

harsh-nod: Good question. So the reason I needed this was because the binary py op was not getting the correct index from the read and in order to get the correct index the binary py op needs some value for elements_per_thread. This is a short-term fix but I think combining index_sequence_analysis and thread_shape_propagation will solve all of these issues.

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -242,13 +248,16 @@ def set_node_index(
...
-    custom = get_custom(node)
-    anchor = custom.anchor
-
     elements_per_thread = getattr(custom, "elements_per_thread", None)
-
-    if isinstance(custom, (Reduction, Placeholder)) and not isinstance(custom, IterArg):
-        return
+    # For elementwise operations that do not have an elements per thread attribute,
+    # look back to the backward slice to see if they can find an appropriate value.
+    if elements_per_thread is None:
```
harsh-nod, line 254: Add TODO:

harsh-nod: done

raikonenfnu (approved):

PR #242, opened 2024-10-25
Minor cleanups and refactoring

harsh-nod:
This PR address comments from a previous PR, namely
- now only one pass through the graph is required to set the vector shapes and index
- the MMA index is now specified using an enum, the MMAOperand, which allows for better verification


raikonenfnu (approved):
LGTM!

PR #243, opened 2024-10-26
[TKW] Teach expansion to handle non direct acc and ReduceOp on reduction dim.

raikonenfnu:
In flash attention, we need to enable non direct acc matmul, and also expansion of reduceOp in reduction dimension. The former is needed in FA since we are applying some scaling to the acc of second MMA before feeding it in. The second case is required in FA because ReduceOp/MaxOp is in the backward slice of second MMA's LHS, which would require it to be expanded in K2/reduction dim as well.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -615,7 +615,7 @@ class BinaryPyOp(CustomOp, ABC):
             raise ValueError(
                 "BinaryPyOp requires lhs and rhs shape to be at least broadcastable."
             )
-        broadcasted_type = lhs_type if lhs_dim_set > rhs_dim_set else rhstype
+        broadcasted_type = lhs_type if lhs_dim_set > rhs_dim_set else rhs_type
         return broadcasted_type
 
 
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -409,11 +409,15 @@ class CustomOp(ABC):
             new_node.reduction_dim = self.fx_node.reduction_dim
         return get_custom(new_node)
 
-    def replace_all_uses_with(self, new_node: CustomOp | fx.Node):
+    def replace_all_uses_with(
+        self,
+        new_node: CustomOp | fx.Node,
+        filter_fn: Callable[[fx.node], bool] = lambda x: True,
+    ):
         """Replace all uses of the current node with the new node."""
         if isinstance(new_node, CustomOp):
             new_node = new_node.fx_node
-        self.fx_node.replace_all_uses_with(new_node)
+        self.fx_node.replace_all_uses_with(new_node, filter_fn)
 
     def erase(self):
         """Erase the current node from the graph where it exists."""
@@ -1321,6 +1325,10 @@ class ReduceOp(CustomOp, ABC):
             return len(self.dim)
         else:
             return 1
+
+    @property
+    def reduction_dim(self) -> IndexSymbol:
+        return self.dim
 
 
 # TODO: Add support for more shuffle types.
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -409,15 +409,11 @@ class CustomOp(ABC):
             new_node.reduction_dim = self.fx_node.reduction_dim
         return get_custom(new_node)
 
-    def replace_all_uses_with(
-        self,
-        new_node: CustomOp | fx.Node,
-        filter_fn: Callable[[fx.node], bool] = lambda x: True,
-    ):
+    def replace_all_uses_with(self, new_node: CustomOp | fx.Node):
         """Replace all uses of the current node with the new node."""
         if isinstance(new_node, CustomOp):
             new_node = new_node.fx_node
-        self.fx_node.replace_all_uses_with(new_node, filter_fn)
+        self.fx_node.replace_all_uses_with(new_node)
 
     def erase(self):
         """Erase the current node from the graph where it exists."""
@@ -1304,7 +1300,12 @@ class ReduceOp(CustomOp, ABC):
             from ..wave.utils import all_equal
 
             src_types = [get_custom(arg).type for arg in self.arg]
-            if not all_equal(src_types):
+            ref_shape = src_types[0].symbolic_shape
+            ref_dtype = src_types[0].dtype
+            if not all(
+                src_type.symbolic_shape == ref_shape and src_type.dtype == ref_dtype
+                for src_type in src_types
+            ):
                 raise NotImplementedError(
                     "NYI: Only support case where all inputs to ReduceOp to have same type."
                 )
--- a/iree/turbine/kernel/wave/expansion.py
+++ b/iree/turbine/kernel/wave/expansion.py
@@ -553,6 +553,53 @@ def get_dim_scaling(
     return dim_scaling
 
 
+def _expand_mma_tiled_reduction(
+    mma: MMA,
+    trace: CapturedTrace,
+    dim_query: dict[IndexSymbol, int],
+    dim_scaling: dict[IndexSymbol, int],
+    context: ExpandedNodeMap,
+    get_node_dim_scaling: Callable[[fx.Node], dict[IndexSymbol, int]],
+    res_idx: int,
+) -> CustomOp:
+    latest_reduced_op = mma
+    for scale_idx in range(1, dim_scaling[mma.reduction_dim]):
+        dim_query[mma.reduction_dim] = scale_idx
+        # Temporarily replace the loop carried arg here to avoid
+        # duplicated expansion. Otherwise we have the following situation:
+        # Suppose we have:
+        #   mma_0_0_0(..., acc_0_0_0)
+        #   mma_0_0_1(..., mma_0_0_0)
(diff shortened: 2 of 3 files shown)


iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -409,11 +409,15 @@ def copy(
             new_node.reduction_dim = self.fx_node.reduction_dim
         return get_custom(new_node)
 
-    def replace_all_uses_with(self, new_node: CustomOp | fx.Node):
+    def replace_all_uses_with(
+        self,
+        new_node: CustomOp | fx.Node,
+        filter_fn: Callable[[fx.node], bool] = lambda x: True,
+    ):
```
harsh-nod, line 416: Can you add a comment describing what the filter function means? Does it mean that we will replace all uses except the ones that the filter function returns false for?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -611,7 +615,7 @@ def type(self) -> Memory:
             raise ValueError(
                 "BinaryPyOp requires lhs and rhs shape to be at least broadcastable."
             )
-        broadcasted_type = lhs_type if lhs_dim_set > rhs_dim_set else rhstype
+        broadcasted_type = lhs_type if lhs_dim_set > rhs_dim_set else rhs_type
```
harsh-nod, line 618: Nice! Not sure how that got by code review :)

iree/turbine/kernel/wave/expansion.py
```diff
@@ -553,6 +553,53 @@ def get_dim_scaling(
...
+            trace,
+            dim_query,
+            dim_scaling,
+            context,
+            get_node_dim_scaling,
+            res_idx,
+        )
+
+        # This expansion always happens, user should never be reused
+        assert new_node != latest_reduced_op
```
harsh-nod, line 593: I would modify this comment to say that the node is always cloned and that is why new node will never be equal to latest reduced op

iree/turbine/kernel/wave/expansion.py
```diff
@@ -553,6 +553,53 @@ def get_dim_scaling(
...
+    mma: MMA,
+    trace: CapturedTrace,
+    dim_query: dict[IndexSymbol, int],
+    dim_scaling: dict[IndexSymbol, int],
+    context: ExpandedNodeMap,
+    get_node_dim_scaling: Callable[[fx.Node], dict[IndexSymbol, int]],
+    res_idx: int,
+) -> CustomOp:
+    latest_reduced_op = mma
+    for scale_idx in range(1, dim_scaling[mma.reduction_dim]):
```
harsh-nod, line 566: Just wondering why this has to start from 1?

iree/turbine/kernel/wave/expansion.py
```diff
@@ -566,53 +613,63 @@ def _handle_reduction_dim(
...
-
-                saved_arg = user.node_args[index]
-                user.update_arg(index, dummy)
-                new_node = _expand_node(
-                    user,
+        dim_scaling = get_node_dim_scaling(root_op)
+        dims = dict(root_op.fx_node.expanded_dims)
+        latest_reduced_op = root_op
+        op_output_index = get_output_index(root_op)
+        if dim_scaling[reduction.axis] <= 1:
```
harsh-nod, line 644: Do we need this check? Shouldn't it be automatically handled?

iree/turbine/kernel/wave/expansion.py
```diff
@@ -566,53 +613,63 @@ def _handle_reduction_dim(
...
+                dims,
+                dim_scaling,
+                context,
+                get_node_dim_scaling,
+                res_idx,
+            )
+        elif isinstance(root_op, ReduceOp):
+            original_src = latest_reduced_op.arg
+            for scale_idx in range(1, dim_scaling[reduction.axis]):
+                dims[root_op.reduction_dim] = scale_idx
```
harsh-nod, line 659: this is pretty cool, so if i understand this right, you are expanding all the inputs and replacing the original argument with all the expanded args?

iree/turbine/kernel/wave/expansion.py
```diff
@@ -566,53 +613,63 @@ def _handle_reduction_dim(
     # TODO: Register iter args with the reduction initially so accessing them is easier
     iter_args: list[CustomOp] = []
     reduction_subgraph = trace.get_subgraph(reduction.subgraph_name)
+
+    # TODO: Add support for case where we process MMA before returning to IterArg.
```
harsh-nod, line 617: Don't fully follow this comment, can you explain?

harsh-nod (changes requested):
Overall, looks good! Thanks I like the changes to expansion. Just some comments about documentation .

iree/turbine/kernel/ops/wave_ops.py
raikonenfnu: Ah, this is dead code, Used this previously to make `replace_all_uses_with` work more similar to `replace_all_uses_except` where the `filter_fn` is the lambda that skips over op if it's true.

iree/turbine/kernel/ops/wave_ops.py
raikonenfnu: haha yee :)

iree/turbine/kernel/wave/expansion.py
raikonenfnu: Added the doc here, it was from the original code
`The initial nodes are expanded in the first dimension, so we start from 1`

iree/turbine/kernel/wave/expansion.py
raikonenfnu: Cleared up the comment to "Handle case where MMAs/ReduceOps do not have Output as direct consumer."
LMK if it's still not well worded :)

iree/turbine/kernel/wave/expansion.py
raikonenfnu: removed

iree/turbine/kernel/wave/expansion.py
raikonenfnu: That is correct!

iree/turbine/kernel/wave/expansion.py
raikonenfnu: done!

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: Yes that's what it looked like to me when I first saw it. Sounds like a useful function to have but we can add that later.

iree/turbine/kernel/wave/expansion.py
harsh-nod: cool thanks

iree/turbine/kernel/wave/expansion.py
harsh-nod: thanks

iree/turbine/kernel/wave/expansion.py
harsh-nod: yes that makes a lot more sense, thanks

iree/turbine/kernel/wave/expansion.py
```diff
@@ -327,33 +327,35 @@ def _expand_reduction(
...
+                arg,
+                trace,
+                dims,
+                get_node_dim_scaling(arg),
+                context,
+                get_node_dim_scaling,
+                res_idx,
             )
+            # Proceed with expansion inside the reduction
+            if expanded_output in new_output_args:
```
harsh-nod, line 340: when do you get this situation where you expanded something that is already in the new_output_args? is it even worth expanding if it is not needed? can we use the context to detect this and prevent expanding?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1300,7 +1300,12 @@ def type(self) -> Memory:
             from ..wave.utils import all_equal
 
             src_types = [get_custom(arg).type for arg in self.arg]
-            if not all_equal(src_types):
+            ref_shape = src_types[0].symbolic_shape
```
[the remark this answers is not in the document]
raikonenfnu: For some reason, I have cases where `Register[M, K2].of(f32)) != `Register[M, K2].of(f32))`. I think something to do with the metaclass

iree/turbine/kernel/wave/expansion.py
raikonenfnu: It's not getting re-expanded, but for some reason it's getting reinserted

harsh-nod: same as above, maybe add a TODO explaining this and we can look into this later.

harsh-nod (approved):
overall looks good and we can address the open questions later. Thanks!

iree/turbine/kernel/ops/wave_ops.py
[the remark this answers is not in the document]
raikonenfnu: done! :)

iree/turbine/kernel/wave/expansion.py
raikonenfnu: done!

iree/turbine/kernel/ops/wave_ops.py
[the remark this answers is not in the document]
harsh-nod: thanks!

PR #244, opened 2024-10-29
[TKW] Fix indexing of permute to enable attention

raikonenfnu:
Most of the time in GPU programming, we would only materialize "transposes"/"permutes" of data during reads and writes. When doing transposes/permutations of data in GPU registers, it is most of the time free/no-op, since threads will still own the same data but just symbolically different.


harsh-nod (approved):
this is awesome! our first ever attention! thanks for making this happen! :)

PR #247, opened 2024-10-29
Add support for varying vector shapes

harsh-nod:
This PR adds support for expanding operators
with varying vector shapes, specifically for the
MMA case where either the same dimension has different vector shapes in different mmas or if different instructions are being used.

The idea is to insert a reshape operator whenever such a shape mismatch is discovered.  The reshape operator lowers to an extract or concatenate operation, depending on the context.

### DIFF
--- a/iree/turbine/kernel/wave/codegen.py
+++ b/iree/turbine/kernel/wave/codegen.py
@@ -67,6 +67,7 @@ from ..ops.wave_ops import (
     scheduling_group_barrier,
     cast,
     permute,
+    reshape,
 )
 from ..lang.wave_types import IndexMapping, IndexSymbol
 from ..compiler.base import CodegenError, ValidationError, NDEBUG
@@ -1310,3 +1311,36 @@ def handle_permute(emitter: WaveEmitter, node: fx.Node):
         raise ValidationError("Malformed arguments") from e
     vector_src = cast_py_value(emitter, register)
     emitter.bind_node_proxy(node, vector_src)
+
+
+@handle_op(reshape)
+def handle_reshape(emitter: WaveEmitter, node: fx.Node):
+    try:
+        args, target_vector_shapes = node.args
+    except ValueError as e:
+        raise ValidationError("Malformed arguments") from e
+    custom = get_custom(node)
+    innermost_dim = custom.type.symbolic_shape[-1]
+    offset = custom.expanded_dims[innermost_dim]
+
+    # Determine whether to extract or combine.
+    if len(args) == 1:
+        # Extract the appropriate slice.
+        size = (
+            target_vector_shapes[innermost_dim] // custom.vector_shapes[innermost_dim]
+        )
+        vector = cast_vector(emitter, args[0])
+        result_type = VectorType.get([size], vector.type.element_type)
+        slice = vector_d.extract_strided_slice(
+            result_type,
+            vector,
+            [offset * size],
+            [size],
+            [1],
+        )
+        emitter.bind_node_proxy(node, IRProxyValue(slice))
+        return
+
+    raise NotImplementedError(
+        "reshape: Currently only handles cases where target_vector_shapes > custom.vector_shapes"
+    )
--- a/iree/turbine/kernel/wave/codegen.py
+++ b/iree/turbine/kernel/wave/codegen.py
@@ -1324,23 +1324,23 @@ def handle_reshape(emitter: WaveEmitter, node: fx.Node):
     offset = custom.expanded_dims[innermost_dim]
 
     # Determine whether to extract or combine.
-    if len(args) == 1:
-        # Extract the appropriate slice.
-        size = (
-            target_vector_shapes[innermost_dim] // custom.vector_shapes[innermost_dim]
+    if len(args) > 1:
+        raise NotImplementedError(
+            "reshape: Currently only handles cases where target_vector_shapes > custom.vector_shapes"
         )
-        vector = cast_vector(emitter, args[0])
-        result_type = VectorType.get([size], vector.type.element_type)
-        slice = vector_d.extract_strided_slice(
-            result_type,
-            vector,
-            [offset * size],
-            [size],
-            [1],
-        )
-        emitter.bind_node_proxy(node, IRProxyValue(slice))
-        return
 
-    raise NotImplementedError(
-        "reshape: Currently only handles cases where target_vector_shapes > custom.vector_shapes"
+    # Extract the appropriate slice. The offset is obtained from the expanded_dim
+    # and so corresponds to the dim_query during expansion. To obtain the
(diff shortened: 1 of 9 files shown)


iree/turbine/kernel/wave/codegen.py
```diff
@@ -1310,3 +1311,36 @@ def handle_permute(emitter: WaveEmitter, node: fx.Node):
...
+    try:
+        args, target_vector_shapes = node.args
+    except ValueError as e:
+        raise ValidationError("Malformed arguments") from e
+    custom = get_custom(node)
+    innermost_dim = custom.type.symbolic_shape[-1]
+    offset = custom.expanded_dims[innermost_dim]
+
+    # Determine whether to extract or combine.
+    if len(args) == 1:
```
raikonenfnu, line 1327: Can we just exit out early instead of guarding the rest of the codegeneration?

iree/turbine/kernel/wave/codegen.py
```diff
@@ -1310,3 +1311,36 @@ def handle_permute(emitter: WaveEmitter, node: fx.Node):
...
+        # Extract the appropriate slice.
+        size = (
+            target_vector_shapes[innermost_dim] // custom.vector_shapes[innermost_dim]
+        )
+        vector = cast_vector(emitter, args[0])
+        result_type = VectorType.get([size], vector.type.element_type)
+        slice = vector_d.extract_strided_slice(
+            result_type,
+            vector,
+            [offset * size],
```
raikonenfnu, line 1337: would be good to have small doc/comment on `offset = offset * size`. Perhaps also include info on how the current `offset` is really `dim_query` offset, and we need to multiply by size to get actual offset.

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -182,6 +185,19 @@ def is_contiguous_dim(
...
+    for arg in custom.node_args.values():
+        if not isinstance(arg, Sequence):
+            arg = [arg]
+        for subarg in arg:
+            # These are ops in the parent graph that have not had their vector shapes set yet.
+            if subarg.vector_shapes is None:
+                continue
+            if subarg.vector_shapes != custom.vector_shapes:
+                with custom.graph.inserting_before(custom.fx_node):
+                    Reshape(subarg, custom.vector_shapes).add_to_graph(custom.graph)
```
raikonenfnu, line 198: Dead code?

iree/turbine/kernel/wave/utils.py
```diff
@@ -234,6 +250,19 @@ def is_mma(node):
             custom.vector_shapes.update(hardware_constraint.vector_shapes)
         custom.anchor = custom
         custom.reduction_dim = k
+        if last_vector_shapes and last_vector_shapes != custom.vector_shapes:
```
raikonenfnu, line 253: Can't we check the the "vector_shapes" per operand of the MMA, as opposed to checking what is the latest vector shape we come across?

raikonenfnu (commented):
Super awesome stuff Harsh! this is exciting! I have a few Qs tho :)

iree/turbine/kernel/wave/codegen.py
harsh-nod: sure will change.

iree/turbine/kernel/wave/codegen.py
harsh-nod: sure will add a comment explaining this.

iree/turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: good catch, will remove

iree/turbine/kernel/wave/utils.py
harsh-nod: Yes, currently this is limited to only checking the current and previous mma. Let me try and modify this to account for that.

harsh-nod: Modified to handle this.

iree/turbine/kernel/wave/expansion.py
```diff
@@ -496,6 +510,84 @@ def _expand_mma_reduction(
...
+            arg_dim_query,
+            get_node_dim_scaling(custom_arg.fx_node),
+            context,
+            get_node_dim_scaling,
+            res_idx,
+        )
+        new_args.append(new_node.fx_node)
+
+    reshape.update_arg("args", new_args)
+
```
raikonenfnu, line 589: Do we need to do anything special to ensure this ordering of expansion is correct?

iree/turbine/kernel/wave/utils.py
```diff
@@ -243,6 +258,32 @@ def is_mma(node):
...
+    # chained matmuls when the vector shapes of the operands in one matmul
+    # differ from those in another matmul.
+    for src in mma_nodes:
+        custom_src = get_custom(src)
+        for dst in mma_nodes:
+            if src == dst:
+                continue
+            custom_dst = get_custom(dst)
+            lhs_slice = capture_backward_slice(custom_dst.lhs)
+            rhs_slice = capture_backward_slice(custom_dst.rhs)
```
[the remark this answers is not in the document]
harsh-nod: yes we can, but then we would have to iterate through all the operators in the slice to see if any of them has an mma as an argument. That would end up with a triple-nested-loop. Unless you are thinking of something else?

iree/turbine/kernel/wave/expansion.py
harsh-nod: I don't think so, but do you have an example in mind?

iree/turbine/kernel/wave/utils.py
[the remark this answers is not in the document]
harsh-nod: okay I think I have a cleaner version now. let me know what you think.

iree/turbine/kernel/wave/utils.py
```diff
@@ -243,6 +258,45 @@ def is_mma(node):
...
+        slice = list(capture_backward_slice(node))
+        for arg in reversed(slice):
+            prev_mma = get_custom(arg)
+            if isinstance(prev_mma, MMA):
+                return prev_mma
+        return None
+
+    for mma in mma_nodes:
+        custom_mma = get_custom(mma)
+        prev_mma = find_mma_in_slice(custom_mma.lhs) or find_mma_in_slice(
```
raikonenfnu, line 294: Looks awesome! just one last small thing then should be all good to merge, what if lhs and rhs come from two different MMAs? can we refactor the code a bit to handle that case?

harsh-nod: sure will do.

raikonenfnu (approved):
Looks awesome, merge please!

PR #248, opened 2024-11-01
Add support for more intrinsics in attention

harsh-nod:
This PR adds support for the 32x32x8 intrinsic in
attention. This should allow for a wider search space when tuning attention.

In order to do so, the following changes were required:
- Add reshape as an anchor op in thread shape analysis, primarily to block the backward propagation of thread shapes from the MMA operands

- Modify assignment of anchor ops to take indexing dims into account

### DIFF
--- a/iree/turbine/kernel/wave/decompose_reduce_ops.py
+++ b/iree/turbine/kernel/wave/decompose_reduce_ops.py
@@ -48,8 +48,6 @@ def determine_shuffle_config(
 
     """
     access_pattern = index[reduction_dim]
-    elements_per_thread = access_pattern.size
-    cluster_size = vector_shapes[reduction_dim] // elements_per_thread
 
     # Since we are only concerned with what happens within a subgroup,
     # we can ignore the TID_1 and TID_2 components of the index. We can
@@ -67,11 +65,11 @@ def determine_shuffle_config(
     offset = access_pattern.start.subs({k: 0 for k in ignore})
     offset = subs_idxc(offset)
     offset_table = [offset.subs({THREAD_0: i}) for i in range(subgroup_size)]
-    # Determine the thread ids participating in the shuffle.
+    unique_offsets = list(dict.fromkeys(offset_table))
+    cluster_size = len(set(offset_table))
     thread_ids = []
-    for i in range(cluster_size):
-        thread_ids.append(offset_table.index(i * elements_per_thread))
-
+    for thread_offset in unique_offsets:
+        thread_ids.append(offset_table.index(thread_offset))
     cluster_stride = [x - y for x, y in zip(thread_ids[1:], thread_ids[:-1])]
     assert all_equal(cluster_stride), f"Cluster stride must be equal across threads."
     return cluster_size, cluster_stride[0]
--- a/iree/turbine/kernel/wave/decompose_reduce_ops.py
+++ b/iree/turbine/kernel/wave/decompose_reduce_ops.py
@@ -66,7 +66,9 @@ def determine_shuffle_config(
     offset = subs_idxc(offset)
     offset_table = [offset.subs({THREAD_0: i}) for i in range(subgroup_size)]
     unique_offsets = list(dict.fromkeys(offset_table))
-    cluster_size = len(set(offset_table))
+    # The cluster size represents the number of unique threads that are participating in a shuffle.
+    # We can obtain this information by just computing the number of unique entries in the offset table.
+    cluster_size = len(unique_offsets)
     thread_ids = []
     for thread_offset in unique_offsets:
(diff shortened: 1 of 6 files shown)


iree/turbine/kernel/wave/decompose_reduce_ops.py
```diff
@@ -67,11 +65,11 @@ def determine_shuffle_config(
     offset = access_pattern.start.subs({k: 0 for k in ignore})
     offset = subs_idxc(offset)
     offset_table = [offset.subs({THREAD_0: i}) for i in range(subgroup_size)]
-    # Determine the thread ids participating in the shuffle.
+    unique_offsets = list(dict.fromkeys(offset_table))
+    cluster_size = len(set(offset_table))
```
raikonenfnu, line 69: this is neat! so we count how many unique threads there are in a layout dimension? can we actually add a comment here to make it easier to understand for future selves?

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -219,6 +219,9 @@ def set_vector_shapes(
             or node in mma_slices[mma][MMA_LHS]
             or node in mma_slices[mma][MMA_RHS]
         ):
+            # Ensure that the operators indexing dims are present in the anchor.
```
raikonenfnu, line 222: Can you remind me what happen if we don't observe this again?

iree/turbine/kernel/wave/utils.py
```diff
@@ -641,7 +642,8 @@ def bfs(
     filter_fn: Callable[[fx.node], bool],
 ) -> set[fx.Node]:
     """
-    Run BFS on the graph to capture the forward slice of a node.
+    Run BFS on the graph. The filter function is not applied to
+    the incoming node.
```
raikonenfnu, line 646: Is this still needed?

iree/turbine/kernel/wave/decompose_reduce_ops.py
```diff
@@ -67,11 +65,11 @@ def determine_shuffle_config(
     offset = access_pattern.start.subs({k: 0 for k in ignore})
     offset = subs_idxc(offset)
     offset_table = [offset.subs({THREAD_0: i}) for i in range(subgroup_size)]
-    # Determine the thread ids participating in the shuffle.
+    unique_offsets = list(dict.fromkeys(offset_table))
```
raikonenfnu, line 68: I don't quite understand this, isn't offset_table of type `list[IndexExpr]`?

raikonenfnu: Ah, we're just trying to get a set of the "unique" offsets

iree/turbine/kernel/wave/decompose_reduce_ops.py
raikonenfnu: actually, is this not the same as len(unique_offsets)?

harsh-nod: The mistaken the original code was making was that the cluster size was being determined from the index. However the cluster size represents the number of threads that are participating in the shuffle and I don't think that can be queried directly. I will add a comment. Yes it is the same as len(unique_ offsets), will change.

iree/turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: sure will update the comment.

iree/turbine/kernel/wave/utils.py
harsh-nod: Just updating the comment to be more accurate and specify that the filter function is not applied to the incoming note.

raikonenfnu: ah right, that makes sense!

iree/turbine/kernel/wave/decompose_reduce_ops.py
harsh-nod: Good question. Yes you are correct that we do not use set so that we can maintain the ordering. But the ordering is not important for expansion or figuring out the indexing dimensions. In this case maintaining the ordering is important so that when we compute the cluster stride we get the correct value. For example, say we have the following unique threads [0, 4, 8, 12]. If we do not maintain the ordering, then we could end up with the following set [0, 8, 4, 12] and now when we do a diff to compute the cluster stride we would get [8, -4, 8] as opposed to [4, 4, 4] which is what we really want.

raikonenfnu: makes sense, thanks!

raikonenfnu (approved):

PR #252, opened 2024-11-05
Add type inference

harsh-nod:
This PR adds a type inference pass to wave. Previously, the types were infered by looking up types from neighbors resulting in inefficient type inference.

Instead, we now introduce a pass that infers the types for all operators in the graph and the inferred type is then stores in the node. New nodes that are constructed in downstream passes are responsible for annotating types for the new operators.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -338,7 +338,7 @@ class CustomOp(ABC):
         vars_str = ", ".join(vars_list)
         return f"{self.tkw_op_name}({vars_str})"
 
-    def add_to_graph(self, region_graph: RegionGraph) -> fx.Node:
+    def add_to_graph(self, region_graph: RegionGraph, type: Any = None) -> fx.Node:
         arg_list = tuple([value for _, value in vars(self).items()])
         self.graph = region_graph
         self.fx_node = region_graph.create_node(
@@ -350,6 +350,10 @@ class CustomOp(ABC):
         self.fx_node.tkw_op = self.__class__
         self.fx_node.tkw_op_name = self.tkw_op_name
         self.fx_node.index = None
+        if type is None:
+            get_custom(self.fx_node).infer_type()
+        else:
+            self.fx_node.type = type
         return self.fx_node
 
     def _add_proxy_to_graph(self, region_graph: RegionGraph):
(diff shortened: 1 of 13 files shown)


iree/turbine/kernel/wave/type_inference.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def infer_types(trace: CapturedTrace | fx.Graph):
+    inferer = TypeInferer()
+    # First, infer the types for all the nodes.
+    for subgraph in trace.region_graph.subgraphs.values():
+        for node in subgraph.nodes:
+            custom = get_custom(node)
+            inferer.infer_types(custom)
+    # Then, set the types.
```
harsh-nod, line 83: Set type within one loop.

harsh-nod: done

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -556,6 +560,23 @@ def vector_shapes(self) -> dict[IndexSymbol, int]:
...
+    @type.setter
+    def type(self, value: Any):
+        self.fx_node.type = value
+
+    def infer_type(self):
+        """
+        Infer the type of this operator using the types
+        of its arguments.
+        """
+        pass
```
raikonenfnu, line 578: maybe we can raise NotImplemented here?

iree/turbine/kernel/wave/expansion.py
```diff
@@ -336,7 +336,7 @@ def _expand_reduction(
             # Add GetResult nodes for the corresponding dimensions
             reduction.graph.inserting_after(reduction.fx_node)
             new_node = GetResult(reduction.fx_node, len(new_output_args))
-            new_node.add_to_graph(reduction.graph)
+            new_node.add_to_graph(reduction.graph, arg.type)
```
raikonenfnu, line 339: Just curious what/why is the regular GetResult get_infer_type not working for this case?

raikonenfnu (commented):
Overall looks awesome! thanks for the refactoring, this is really clean! just couple Qs

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: Good question. So the idea is that most operators will implement infer_types except for Placeholders, etc. For those operators, since they already have a type, we want this to just pass through instead of throwing an error.

iree/turbine/kernel/wave/expansion.py
harsh-nod: Sure I can add that. The short summary is that it would require us to find the output node to get the right type since this is during expansion and the init_args have not been set yet. So since we have access to arg, we just use that instead.

raikonenfnu (approved):

PR #263, opened 2024-11-09
Update package version to 2.9.1.

ScottTodd:
I think this will let source builds appear as newer than the stable PyPI release that was just published (https://pypi.org/project/iree-turbine/2.9.0/). We can iterate on the release process sequencing next week. Maybe we shouldn't touch this version, or maybe we should update the minor version instead of the patch version, I'm not sure.

### DIFF
--- a/version_info.json
+++ b/version_info.json
@@ -1,3 +1,3 @@
 {
-    "package-version": "2.9.0"
+    "package-version": "2.9.1"
 }


marbre:
I would suggest to increment the minor version instead but as 3.0.0 is to be released soon, we even might want to set the version to 3.0.0.dev already which nightlies building as 3.0.0rc202411*. I will add docs to iree.dev soon.

PR #265, opened 2024-11-12
[TKW] Refactor test run config

Hardcode84:
Move default run config to common function


harsh-nod (approved):
thanks for the cleanup!

PR #274, opened 2024-11-15
Add support for dynamic parallel dims in GEMMs

harsh-nod:
This PR adds tests for dynamic M and N dims in
GEMMs. This works out of the box for the most part
    and just requires moving the align index pass
    after scheduling.

### DIFF
--- a/iree/turbine/kernel/wave/codegen.py
+++ b/iree/turbine/kernel/wave/codegen.py
@@ -272,6 +272,14 @@ def gen_sympy_index(dynamics: dict[IndexSymbol, Any], expr: sympy.Expr) -> OpRes
         if get_const_val(rhs) == 0:
             return lhs
 
+        if (rhs_val := get_const_val(rhs)) and rhs_val < 0:
+            # If rhs_val is negative, since it is of index type and we have
+            # overflow flags, we need to express it as a sub to get the
+            # correct result.
+            return arith_d.subi(
+                lhs, _get_const(-rhs_val), overflow_flags=overflow_flags
+            )
+
         return arith_d.addi(lhs, rhs, overflow_flags=overflow_flags)
 
     # `x + (a/b)` transformed into `(x*b + a) / b`
@@ -320,7 +328,18 @@ def gen_sympy_index(dynamics: dict[IndexSymbol, Any], expr: sympy.Expr) -> OpRes
 
     def _ceiling(value):
         if isinstance(value, _Rational):
-            value = arith_d.ceildivsi(*_broadcast(value.numerator, value.denominator))
+            # TODO: This is the expansion of ceildivui, but we should figure
+            # out how to run the arith-expand pass instead of doing this.
+            # ceildivui(x, y) = x == 0 ? 0 : ((x - 1) / y) + 1
+            one = _get_const(1)
+            zero = _get_const(0)
+            lhs_minus_one = arith_d.subi(*_broadcast(value.numerator, one))
+            div = arith_d.divui(*_broadcast(lhs_minus_one, value.denominator))
+            result = arith_d.addi(*_broadcast(div, one))
+            cmp = arith_d.cmpi(
+                arith_d.CmpIPredicate.eq, *_broadcast(value.numerator, zero)
+            )
+            value = arith_d.select(cmp, zero, result)
 
         return value
 
@@ -416,6 +435,16 @@ def gen_sympy_index(dynamics: dict[IndexSymbol, Any], expr: sympy.Expr) -> OpRes
                 _enforce_non_rational(rhs, term)
                 _enforce_non_rational(lhs, term)
                 res = arith_d.andi(*_broadcast(lhs, rhs))
+                stack.append(res)
+            case sympy.Max():
+                rhs = stack.pop()
+                lhs = stack.pop()
+                _enforce_non_rational(rhs, term)
+                _enforce_non_rational(lhs, term)
+                if _is_integer_like_type(rhs.type):
+                    res = arith_d.maxsi(*_broadcast(lhs, rhs))
+                else:
+                    res = arith_d.maximumf(*_broadcast(lhs, rhs))
                 stack.append(res)
             case sympy.logic.boolalg.BooleanFalse():
                 res = arith_d.constant(IntegerType.get_signless(1), 0)
@@ -1062,7 +1091,10 @@ def handle_reduction(emitter: WaveEmitter, node: fx.Node):
 
     # For now, we assume that dimensions that have tiling constraints on them,
     # do not have any other constraints.
-    end = arith_d.constant(IndexType.get(), int(node.count))
+    if isinstance(node.count, sympy.Expr):
+        end = gen_sympy_index(add_emitter_subs(emitter), node.count)
+    else:
+        end = arith_d.constant(IndexType.get(), int(node.count))
 
     # Since we divide the end by the tile size, we need to make sure that the
     # step is 1.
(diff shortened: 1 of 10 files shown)


lit_tests/kernel/wave/codegen.py
```diff
@@ -1184,6 +1184,102 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+            a_reg = tkw.read(a, elements_per_thread=LOAD_ELEMS_PER_THREAD)
+            b_reg = tkw.read(b, elements_per_thread=LOAD_ELEMS_PER_THREAD)
+            acc = tkw.mma(a_reg, b_reg, acc)
+            return acc
+
+        tkw.write(repeat, c, elements_per_thread=STORE_ELEMS_PER_THREAD)
+
+    with tk.gen.TestLaunchContext(
+        {
+            BLOCK_M: 64,
```
stellaraccident, line 1222: Am I reading this right: you've now got a dynamic kernel where you bind the concrete indices at launch? Just making sure to not invest my enthusiasm yet if that is not actually what this is doing :)

harsh-nod: yes you are right, we can have dynamic parallel and reduction dimensions and bind their values at launch time :) This PR just does it for GEMM, but I was going to put another one for attention for paged attention and their variants.

iree/turbine/kernel/wave/codegen.py
```diff
@@ -272,6 +272,14 @@ def addi(lhs, rhs):
         if get_const_val(rhs) == 0:
             return lhs
 
+        if (rhs_val := get_const_val(rhs)) and rhs_val < 0:
+            # If rhs_val is negative, since it is of index type and we have
+            # overflow flags, we need to express it as a sub to get the
```
raikonenfnu, line 277: Can you explain a little more why we'd need the overflow_flags? in theory, wouldn't be rhs_val  always be <= lhs?

iree/turbine/kernel/wave/scheduling/schedule.py
```diff
@@ -112,7 +118,12 @@ def schedule_reduction(
     )
 
     # Update new reduction count.
-    new_reduction.count = max_induction_variable - (scheduler.num_stages - 1)
+    if ivar_is_number:
+        new_reduction.count = max_induction_variable - (scheduler.num_stages - 1)
+    else:
+        new_reduction.count = sympy.Max(
+            0, max_induction_variable - (scheduler.num_stages - 1)
```
raikonenfnu, line 125: Just curious, why wouldn't the count be 1 instead of 0?

tests/kernel/wave/wave_gemm_test.py
```diff
@@ -161,6 +167,19 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
 
+    dynamic_symbols = []
+    dynamic_symbols_map = {}
+    if dynamic_dims:
+        dynamic_symbols_map[M] = hyperparams[M]
+        dynamic_symbols_map[N] = hyperparams[N]
+        dynamic_symbols_map[K] = hyperparams[K]
+        dynamic_symbols.append(M)
+        dynamic_symbols.append(N)
+        dynamic_symbols.append(K)
```
raikonenfnu, line 178: maybe we should move this to the hyper param?

raikonenfnu: actually nvm, hyperparams require the key to be symbols, doesn't have to be in this PR but would be nice to think of some way to clean this up, i.e make this a bit more ergonomic for kernel/torch model authors.

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -55,7 +55,8 @@ def isolated_test_call(
         argument_dims = get_dynamic_dims(
             sig.kernel_buffer_input_bindings, dynamic_symbols
         )
-        input_tensors += [IndexType.get() for _ in argument_dims]
+        # Only add types for the unique dynamic symbols.
+        input_tensors += [IndexType.get() for _ in set(argument_dims)]
```
raikonenfnu, line 59: funny, I also have this in my zero copy PR to handle dynamic writable argument 😁

iree/turbine/kernel/wave/codegen.py
harsh-nod: So basically what happens is that we are trying to do 
```
x - 2
```

Now this lowers to
```
%c_2 = arith.constant -2 : index
addi %x, %c_2, nsw, nuw
``` 
because sympy represents these are additions.
So the problem here is that we end up -2 with index type which is not really -2. What we really want is
```
%c2 = arith.constant 2 : index
subi %x, %c2 
``` 

Now the interesting this is that if we get rid of nsw, nuw, then the addi example works because we allow the overflow to happen. But we want to avoid overflow and so this was a solution to it, but happy to discuss other better solutions.

raikonenfnu: I see I see, fwiw, I don't think we'd need a full discussion on this, just more for knowledge and maybe NIT haha.

but looks like your current change should generate:
```
%c2 = arith.constant 2 : index
subi %x, %c2 
```
with the `arith_d.subi` right? Dosn't that mean we won't need the "overflow" flag anymore?

raikonenfnu:
mostly looks great to me btw! just some minor Qs and NITs :)

iree/turbine/kernel/wave/codegen.py
Hardcode84: That's an interesting example (BTW, we can also check this issue on alive2 https://alive2.llvm.org/ce/z/dpybHs). There may be still similar issues lurking around, but I'd prefer to keep overflow flags.

iree/turbine/kernel/wave/scheduling/schedule.py
harsh-nod: Because this is to handle situations where we don't have enough iterations (count < 0), but based on our discussion I will add an Assumption to handle this.

tests/kernel/wave/wave_gemm_test.py
harsh-nod: This is only required for the pytest. For kernel authors, they would not specify the dimensions in the hyper parameter dictionary.

iree/turbine/kernel/compiler/host_codegen.py
harsh-nod: yes its a bug and I will change to preserve insertion order.

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -260,7 +261,7 @@ def set_node_index(
         c.dim: c for c in constraints if isinstance(c, WorkgroupConstraint)
     }
     other_constraints = [
-        c for c in constraints if not isinstance(c, HardwareConstraint)
+        c for c in constraints if not isinstance(c, (HardwareConstraint, Assumption))
```
raikonenfnu, line 264: BTW, how is the assumption being checked/asserted? .apply?

harsh-nod: Right now, the assumption can be any sympy expression because it is only used to answer queries. So if there are errors in the assumption, they would come from sympy's error checking. But as long as it is a valid sympy expression it should be a valid assumption.

harsh-nod:
I created a PR in IREE to add the Arith-expand pass. https://github.com/iree-org/iree/pull/19200

raikonenfnu (approved):

PR #276, opened 2024-11-17
[TKW] Modify Wave Codegen and runtime to run natively on ROCM tensors

raikonenfnu:


iree/turbine/kernel/wave/utils.py
```diff
@@ -404,6 +404,40 @@ def _invoke(vm_context, device, entry_function, inputs, outputs):
...
+    def push_tensor_to_arg_list(arg_tensor: torch.Tensor):
+        if not arg_tensor.is_contiguous():
+            arg_tensor = arg_tensor.contiguous()
+        capsule = arg_tensor.__dlpack__(None)
+        arg_tensor_bv = device.from_dlpack_capsule(capsule)
+        arg_list.push_ref(arg_tensor_bv)
+
+    # Linearize arguments, since we modify output buffer in place.
+    # In linearized arg_list, we first push in all inputs, and then
+    # we push in the outputs.
```
raikonenfnu, line 421: add note to push in the dynamic dims after the outputs

Hardcode84:
`.to("cuda")` copypaste in tests is not ideal. I think it's better to create our own versions of `randn` and `zeros` in utils (which call torch impls and immediately do `to("cuda")`) and use them.

raikonenfnu:
Thanks for the comment @Hardcode84, do you perhaps some examples on this? I can't think of a way at the top of my mind on how to do it while making it look cleaner than just .to("cuda") haha

raikonenfnu:
@Hardcode84 I see I see, will do that, thanks! :)

raikonenfnu:
> something like
> 
> ```
> def randn(*args, **kwargs):
>     return torch.randn(*args, **kwargs).to("cuda")
> ```
> 
> And put into `wave/utils.py`

Done(I think), thanks for the tip! :)

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -50,12 +51,11 @@ def isolated_test_call(
     dynamic_symbols: list[IndexSymbol] = [],
 ):
     with InsertionPoint(mb.body_block), Location.unknown():
-        input_types = [b.as_mlir_type() for b in sig.kernel_buffer_input_bindings]
+        input_types = [b.as_mlir_type() for b in sig.kernel_buffer_bindings]
         input_tensors = memref_to_tensor(input_types)
```
harsh-nod, line 55: Since now the input types have all the buffers do we need the code below for output types and output tensors?

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -50,12 +51,11 @@ def isolated_test_call(
...
-        argument_dims = get_dynamic_dims(
-            sig.kernel_buffer_input_bindings, dynamic_symbols
-        )
-        input_tensors += [IndexType.get() for _ in argument_dims]
+        argument_dims = get_dynamic_dims(sig.kernel_buffer_bindings, dynamic_symbols)
+        # Adding unique dynamic dims as inputs.
+        input_tensors += [IndexType.get() for _ in list(dict.fromkeys(argument_dims))]
```
harsh-nod, line 58: This makes sense and is better than using set() since it preserves the insertion order.

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -80,13 +80,22 @@ def isolated_test_call(
             dispatch = SymbolRefAttr.get([exe.sym_name.value, entrypoint])
             entrypoints = ArrayAttr.get([dispatch])
 
+            buffer_binding_count = len(sig.kernel_buffer_bindings)
+            offset_to_out_binding = len(sig.kernel_buffer_input_bindings)
```
harsh-nod, line 84: minor: how about renaming this to input_binding_count ?

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -80,13 +80,22 @@ def isolated_test_call(
...
+                ]
+            )
             out = flow_d.DispatchOp(
                 output_tensors,
                 [dynamic_argument_map[dim] for dim in dynamic_symbols],
                 entrypoints,
                 entry_block.arguments,
                 [dynamic_argument_map[dim] for dim in argument_dims],
                 [dynamic_argument_map[dim] for dim in result_dims],
+                tied_operands=tied_operands,
```
harsh-nod, line 98: This is pretty cool. So if operand 2 is tied, does that mean that the output will be written to the memory allocated for operand 2 and so we won't have to allocate new memory?

iree/turbine/kernel/wave/utils.py
```diff
@@ -419,14 +453,32 @@ def _print_bench_result(result, filename):
...
 def compile_and_invoke(
     asm: str,
     func_name: str,
     config: dict[str, str],
     kernel_inputs: list[torch.Tensor],
     kernel_outputs: list[torch.Tensor],
+    kernel_dynamic_dims: list[int] = [],
     run: bool = False,
     run_bench: bool = False,
+    inplace: bool = False,
```
harsh-nod, line 481: Shouldn't we default this to True? This seems like the preferred mode of execution. Do you see any downsides to turning this on by default?

iree/turbine/kernel/wave/utils.py
```diff
@@ -493,7 +548,17 @@ def compile_and_invoke(
...
+            _inplace_invoke(
+                ctx.vm_context,
+                device,
+                func,
+                kernel_inputs,
+                kernel_outputs,
+                kernel_dynamic_dims,
+            )
+        else:
+            _invoke(ctx.vm_context, device, func, kernel_inputs, kernel_outputs)
```
harsh-nod, line 561: Do we need to modify the signature of _invoke to include the kernel dynamic dims since they are no longer included in the kernel inputs?

iree/turbine/kernel/wave/wave.py
```diff
@@ -352,8 +352,11 @@ def test_execute(self, args, kwargs):
                     kernel_outputs.append(arg)
 
             dynamic_symbols_map = kwargs.get("dynamic_symbols_map", {})
+            kernel_dynamic_dims = []
             if dynamic_symbols:
-                kernel_inputs += [dynamic_symbols_map[sym] for sym in dynamic_symbols]
+                kernel_dynamic_dims = [
```
harsh-nod, line 357: minor: Can we use dynamic_symbols_map.values() instead?

tests/kernel/wave/wave_attention_test.py
```diff
@@ -196,7 +198,7 @@ def repeat(
         generate_iree_ref(
             "chain_mmt", [q, k, v], [iree_ref], config, run_bench=run_bench
         )
-        assert_close(output, iree_ref)
+        assert_close(output.cpu(), iree_ref)
```
harsh-nod, line 201: Instead of moving the output to the cpu, should we do check_device=False, which automatically moves the output to the cpu? https://pytorch.org/docs/main/testing.html

harsh-nod (changes requested):
Looks great thanks for doing this. Being able to modify tensors in-place will be really important. Just some minor comments.

iree/turbine/kernel/compiler/host_codegen.py
raikonenfnu: exactly! :)

iree/turbine/kernel/wave/utils.py
raikonenfnu: Great question, I'd need to do some modifications to iree_ref_utils to always enable this to true, but I was hoping we can do that in a follow-on PR. :)

iree/turbine/kernel/wave/utils.py
raikonenfnu: Good catch, yes, I'll update that :)

iree/turbine/kernel/compiler/host_codegen.py
raikonenfnu: yeah, we still need it to set up the result types :)

.github/workflows/ci-tk.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
     - name: Run e2e tests on MI300
       if: "contains(matrix.os, 'mi300') && !cancelled()"
       run: |
+        pip install --no-compile --upgrade -r pytorch-rocm-requirements.txt
```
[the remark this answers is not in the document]
raikonenfnu: Let me try that, I had some issues with CI using the old cpu pytorch for some reason, but may be unrelated issue/fix.

iree/turbine/kernel/wave/utils.py
harsh-nod: That works as well. Sure, we can save that for the next PR.

iree/turbine/kernel/compiler/host_codegen.py
harsh-nod: very nice :)

iree/turbine/kernel/compiler/host_codegen.py
raikonenfnu: done!

iree/turbine/kernel/wave/utils.py
raikonenfnu: done!

iree/turbine/kernel/wave/wave.py
raikonenfnu: nice idea, done!

.github/workflows/ci-tk.yaml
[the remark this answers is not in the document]
raikonenfnu: removed the `--upgrade` and it's still working, looks like some unrelated issue, done! :)

harsh-nod (approved):
lgtm! this should go a long way in getting our kernels more production ready :)

PR #279, opened 2024-11-19
[TKW] Add CDNA2 + CDNA3 Int8 intrinsics and refactor intrinsic enums

raikonenfnu:
- Added CDNA2 int8 intrinsic layouts
- Modified iree_ref to handle int gemms
- Modified certain e2e test to require certain GPU arch to be available
- Modified enum for easy handling in the future
- Get default architecture function
- Borrowed device_randint from Ivan
- Turn on CDNA2 runner for TK-CI

Manually tested that the generated iree_ref for int gemms are working as expected!


iree/turbine/kernel/wave/iree_utils.py
```diff
@@ -93,7 +94,7 @@ def get_mmt_asm(
     if not cast_fp8:
         matmul_function = f"""
         func.func @{func_name}(%lhs: tensor<{lhs_type}>, %rhs: tensor<{rhs_type}>) -> tensor<{acc_type}> {{
-          %c0 = arith.constant 0.0 : {acc_dtype}
+          %c0 = arith.constant {zero} : {acc_dtype}
```
Hardcode84, line 97: nit: you probably can just do `zero = "0.0" if acc_dtype.starts_with("f") else "0"` instead of passing it as parameter.

iree/turbine/kernel/wave/utils.py
```diff
@@ -912,6 +924,10 @@ def device_randn(*args, **kwargs):
     return torch.randn(*args, **kwargs).to("cuda")
 
 
+def device_randint(*args, **kwargs):
+    return torch.randint(*args, **kwargs).to("cuda")
```
Hardcode84, line 928: As we are handling `torch.cuda.is_available() == False` in `get_default_arch()`, probably should also handle it here and use cpu tensors.

raikonenfnu: done!

iree/turbine/kernel/wave/iree_utils.py
raikonenfnu: nice idea, done!

iree/turbine/kernel/wave/iree_utils.py
```diff
@@ -138,7 +138,7 @@ def get_conv_asm(
 
 
 def dtype_str(dtype: torch.dtype) -> str:
-    dtype_str = TORCH_DTYPE_TO_MLIR_TYPE_ASM.get(dtype, None)
+    dtype_str = TORCH_DTYPE_TO_IREE_TYPE_ASM[dtype]
```
raikonenfnu, line 141: FYI: This is because `TORCH_DTYPE_TO_MLIR_TYPE_ASM` converts `torch.int32` into `si32`, but IREE do not handle `si32` very well. Hence we use `TORCH_DTYPE_TO_IREE_TYPE_ASM` that generates `i32`.

Hardcode84 (approved):
Couple more comments, but otherwise LGTM

tests/kernel/wave/wave_gemm_test.py
```diff
@@ -219,6 +227,299 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
 
 
 @require_e2e
+@require_cdna2
```
harsh-nod, line 230: This is cool. We should talk to Sai and try to get a MI250 builder to verify our CDNA2 tests.

harsh-nod (changes requested):
This looks great and I am very surprised that you were able to reuse all the other intrinsics. One final ask - can you add a lit test showing the IR form of the mfmas?

iree/turbine/kernel/wave/utils.py
```diff
@@ -908,12 +920,23 @@ def all_equal(input_list: list[Any]) -> bool:
     return all(elem == input_list[0] for elem in input_list)
 
 
+def get_default_device():
+    return "cuda" if torch.cuda.is_available() else "cpu"
```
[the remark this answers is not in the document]
raikonenfnu: done, thanks!

tests/kernel/wave/wave_gemm_test.py
```diff
@@ -219,6 +227,299 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+        MMA_DELAY: 1,
+        VALU_DELAY: 1,
+        SHUFFLE_DELAY: 1,
+        SHARED_MEMORY_UNITS: 4,
+        GLOBAL_MEMORY_UNITS: 4,
+        MMA_UNITS: 4,
+        VALU_UNITS: 8,
+        SHUFFLE_UNITS: 8,
+    }
+    config = {"backend": "rocm", "device": "hip", "target": "gfx942"}
```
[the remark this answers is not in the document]
raikonenfnu: > did you mean gfx90a
Ah good catch, done!

> get_default_run_config() to use get_default_arch()

Yeah, I'll do in separate PR because that'll also impact LIT tests haha

iree/turbine/kernel/wave/constraints.py
```diff
@@ -15,11 +15,45 @@
...
+  * For AMD:
+    * 0 = CDNA1
+    * 1 = CDNA2
+    * 2 = CDNA3
+    * 8 = RDNA3
+* C = element type of A-matrix:
+  * 0 = 64-bit float (e.g. IEEE754 double precision)
+  * 1 = 32-bit float (e.g. IEEE754 single precision, and "xf32" fast variants)
+  * 2 = 16-bit float (incl. IREE754 half and bf16)
+  * 3 = 8-bit float (incl. f8E5M2, f8E4M3, and "FNUZ" variants)
```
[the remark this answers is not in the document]
raikonenfnu: In IREE, we are not doing that, I think we should be OK for now haha :)

iree/turbine/kernel/wave/constraints.py
```diff
@@ -89,13 +123,13 @@ def get_thread_id_from_workgroup_dim(self, workgroup_dim: int) -> IndexSymbol:
     def mma_matrix_shapes(self) -> tuple[int]:
         # TODO: Eventually the shapes and indices should be provided by a tool
         match self.mma_type:
-            case MMAType.F32_16x16x16_F16:
+            case MMAType.F32_16x16x16_F16 | MMAType.I32_16x16x16_I8:
```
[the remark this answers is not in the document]
raikonenfnu: yeah, saw the layouts and they are indeed the same! 😁

tests/kernel/wave/wave_gemm_test.py
raikonenfnu: that'll be very nice! Initially, I added that flag S.T the cdna3 exclusive tests wouldn't fail when @Giuseppe5 is running on his machine haha.

.github/workflows/perf.yaml
```diff
@@ -61,3 +61,10 @@ jobs:
         export WAVE_RUN_E2E_TESTS=1
         export TEST_PARAMS_PATH="tests/kernel/wave/test_param.json"
         pytest -n 1 --capture=tee-sys -vv ./tests/kernel/wave/
+
+    - name: Run e2e tests on MI250
```
saienduri, line 65: You can remove this step completely, and just keep the one before it. Just hange its name from Run e2e tests on MI300 to Run e2e tests on AMD GPU or something and remove the `if: "contains(matrix.os, 'mi300') && !cancelled()"` line

Hardcode84:
For correctness check probably should add to `ci-tk.yaml` instead. `perf.yaml` was meant for performance, but this work was never finished.

raikonenfnu:
> For correctness check probably should add to ci-tk.yaml instead. perf.yaml was meant for performance measurement, but this work was never finished.

Makes sense! thanks!

.github/workflows/perf.yaml
raikonenfnu: Can you further explain? If I understand correctly, you want to only have one test? but we need to run both MI250 and MI300 though

raikonenfnu: actually I see what you mean!

raikonenfnu: Done! thanks you again for the tip! :)

tests/kernel/wave/wave_gemm_test.py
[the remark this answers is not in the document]
raikonenfnu: it's now in the same PR, thanks!

tests/kernel/wave/wave_gemm_test.py
raikonenfnu: done, I think! :)

tests/kernel/wave/wave_gemm_test.py
```diff
@@ -219,6 +227,299 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+
+
+@require_e2e
+@require_cdna3
+@pytest.mark.parametrize("shape", get_test_shapes("test_gemm"))
+@pytest.mark.parametrize("enable_scheduling", [False, True])
+@pytest.mark.parametrize(
+    "mfma_variant",
+    [
+        MMAType.F32_16x16x32_F8,
```
Giuseppe5, line 398: Shouldn't this be I8?

tests/kernel/wave/wave_gemm_test.py
```diff
@@ -219,6 +227,299 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+        run=True,
+        run_bench=run_bench,
+        run_config=config,
+        schedule=enable_scheduling,
+        use_scheduling_barriers=enable_scheduling_barriers,
+        dynamic_symbols=dynamic_symbols,
+        dynamic_symbols_map=dynamic_symbols_map,
+    ):
+        randint_hi = 4
+        a = device_randint(randint_hi, (shape[0], shape[2]), dtype=torch.int16)
```
Giuseppe5, line 371: Would it be possible to pass here directly a torch.int8?

tests/kernel/wave/wave_gemm_test.py
raikonenfnu: ahh, good catch! it was passing because the generated instructions between them are technically in the mlir vector level are the same haha, will fix that. :)

raikonenfnu: done!

tests/kernel/wave/wave_gemm_test.py
raikonenfnu: yeap! done!

harsh-nod (approved):
thanks! this looks great!

PR #280, opened 2024-11-20
Add dynamic attention

harsh-nod:
This PR adds support for dynamic attenton dims
for attention (except K1).

### DIFF
--- a/iree/turbine/kernel/compiler/kernel_codegen.py
+++ b/iree/turbine/kernel/compiler/kernel_codegen.py
@@ -20,6 +20,7 @@ from enum import Enum
 from dataclasses import dataclass
 
 import torch.fx as fx
+import sympy
 
 from .._support.indexing import (
     IndexingContext,
@@ -79,6 +80,10 @@ class BindingType(Enum):
     SYMBOL_VALUE = 2
 
 
+def _is_symbolic(value: list[sympy.Expr | int]) -> bool:
+    return any(isinstance(v, sympy.Expr) for v in value)
+
+
 @dataclass
 class BindingDesc:
     # The unique reference object that this is derived from. This will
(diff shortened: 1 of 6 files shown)


iree/turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -120,7 +125,7 @@ def sym_to_dim_asm(s: IndexSymbol) -> str:
                 # Unranked. Not well supported, but for completeness.
                 spec_asm = element_type_asm
             strides = strides_from_symbolic_shape(idx_context, kb_t.symbolic_shape)
-            if strides is None:
+            if strides is None or _is_symbolic(strides):
```
Hardcode84, line 128: If we have a symbolic (i.e. dynamic) strides, shouldn't memref type be `memref<{spec_asm}, strided<[?, ?, ...], offset: ?>`?

iree/turbine/kernel/compiler/utils.py
```diff
@@ -6,6 +6,7 @@
 def strides_from_symbolic_shape(
     indexing_context: IndexingContext,
     symbolic_shape: Optional[list[IndexSymbol]],
+    allow_mixed_shapes: bool = False,
```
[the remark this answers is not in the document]
harsh-nod: Yes, some other tests (outside of wave) fail if I remove this.

iree/turbine/kernel/compiler/kernel_codegen.py
harsh-nod: Good point, I will fix this.

harsh-nod: done

Hardcode84 (approved):

PR #282, opened 2024-11-20
Update LLVM patch instructions

martin-luecke:
This updates the instructions for patching LLVM with the fixes to preserve our scheduling decisions.

The old instructions do not fetch the commit anymore. The cherry-pick yielded the error: `fatal: bad object ee52732cddae42deed2e3387a83b20ec05860b4e`.
I changed this to only fetch the relevant commit.


harsh-nod (approved):
Thanks for fixing this!

PR #286, opened 2024-11-21
Fix bugs in benchmarking

harsh-nod:
This PR fixes issues in GEMM benchmarking related to shapes, device of tensors and passing inputs and outputs to iree-benchmark-module.


iree/turbine/kernel/wave/iree_utils.py
```diff
@@ -208,4 +208,5 @@ def generate_iree_ref(
         kernel_outputs,
         run=True,
         run_bench=kwargs.get("run_bench", False),
+        pass_outputs_to_kernel=False,
```
Hardcode84, line 211: Semantically, it should be named `(not)inplace_outputs`, but we are using `inplace=False` only in `generate_iree_ref` so I think we can just unify those 2 params. Also, while we are here, can you pass `inplace=False` explicitly in `generate_iree_ref`.

harsh-nod: Correct me if I'm wrong but looks like inplace is always False. It has not been turned on yet. So I don't think we can unify them unless I turn it on?

harsh-nod: Sorry my mistake, yes I will unify and change.

Hardcode84: `test_execute` uses `inplace=True`

Hardcode84 (approved):

harsh-nod:
So there are 2 more outstanding issues:
1. Unknown HIP target 'cpu' (which I believe is coming from our automatic device detection?)
2. Accuracy issue with attention on MI250

harsh-nod:
> So there are 2 more outstanding issues:
> 
> 1. Unknown HIP target 'cpu' (which I believe is coming from our automatic device detection?)
> 2. Accuracy issue with attention on MI250

Okay restarting the builders seemed to fix these issues.

PR #288, opened 2024-11-21
[WIP] Propagate dim index

Hardcode84:
Refactor `thread_shape_analysis` to take into account entire index instead of just elements per thread count.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -1349,6 +1349,10 @@ class ReduceOp(CustomOp, ABC):
     def reduction_dim(self) -> IndexSymbol:
         return self.dim
 
+    @property
+    def elements_per_thread(self) -> int:
+        return 1
+
 
 # TODO: Add support for more shuffle types.
 @define_op("shuffle")
--- a/iree/turbine/kernel/wave/thread_shape_analysis.py
+++ b/iree/turbine/kernel/wave/thread_shape_analysis.py
@@ -197,7 +197,7 @@ def determine_thread_shapes(trace: CapturedTrace):
                 bwd_slice = capture_backward_slice(custom.init, propagatable_op)
             reduce_dims = frozenset(
                 [
-                    DimIndex(dim, IndexSequence(seq.start, 1, 1))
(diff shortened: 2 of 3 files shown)


iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1349,6 +1349,10 @@ def num_reduction_dims(self) -> int:
     def reduction_dim(self) -> IndexSymbol:
         return self.dim
 
+    @property
```
harsh-nod, line 1352: Can you add a comment explaining why this is necessary?

iree/turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -51,7 +57,7 @@ def set_index_size(custom: CustomOp, target_dim_sizes: list[DimSize]):
 # Anchor Indicies and Conflict resolution helpers
 #################################################################
 
-anchorOpTypes = (Read, Write, MMA, ReduceOp, Reshape)
+anchorOpTypes = (Read, Write, MMA, ReduceOp, Reshape, Permute)
 noHandleTypes = (Placeholder, Output, ExtractSlice, Allocate)
```
harsh-nod, line 61: Also a comment explaining why Permute is added as an anchor op?

harsh-nod (changes requested):
lgtm! just some requests for comments and also if you could check if we still need the propagation in index_sequence_analysis of elements_per_thread? thanks!

iree/turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -51,7 +57,7 @@ def set_index_size(custom: CustomOp, target_dim_sizes: list[DimSize]):
 # Anchor Indicies and Conflict resolution helpers
 #################################################################
 
-anchorOpTypes = (Read, Write, MMA, ReduceOp, Reshape)
+anchorOpTypes = (Read, Write, MMA, ReduceOp, Reshape, Permute)
```
raikonenfnu, line 60: Can we add the forward propagation of permute, just as safety measure to ensure we won't be generating "valid" but incorrect IRs. Speaking from experience, would be much better for program to crash than to debug why MLIR is wrong and where the wrong is coming from. 😄

PR #290, opened 2024-11-22
[TKW] Clear cache after every test for robustify test

raikonenfnu:
Tests on MI250 is randomly passing and failing. In order to robustify them, this PR adds manual_seed in places that are missing, and add cuda.clear_cache() to ensure we don't run out of memory.

### DIFF
--- a/tests/kernel/wave/wave_attention_test.py
+++ b/tests/kernel/wave/wave_attention_test.py
@@ -13,10 +13,7 @@ import iree.turbine.kernel as tk
 import iree.turbine.kernel.lang as tkl
 import iree.turbine.kernel.wave as tkw
 from iree.turbine.kernel.lang.global_symbols import *
-from iree.turbine.kernel.wave.iree_utils import (
-    generate_iree_ref,
-    initialize_seed_and_cache,
-)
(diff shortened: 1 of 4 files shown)


tests/kernel/wave/wave_attention_test.py
```diff
@@ -193,6 +193,8 @@ def repeat(
         schedule=enable_scheduling,
         use_scheduling_barriers=enable_scheduling_barriers,
     ):
+        torch.cuda.empty_cache()
```
harsh-nod, line 196: what do you think of moving this to a 
```
initialize_test(seed=0)
```
function in utils.py?

harsh-nod (changes requested):
thanks! how can we validate that this is really the root cause and that it fixes the issue?

tests/kernel/wave/wave_attention_test.py
raikonenfnu: as in a function that has both torch.cuda.empty_cache() + torch.manual_seed(seed)?

raikonenfnu: done, but renamed to `initialize_seed_and_cache` for better explanation to what it does. :)

PR #292, opened 2024-11-22
Add propagation of thread dependent index sequences

harsh-nod:
This PR adds a major refactor of the index sequence analysis. Specifically,
1. Index sequence computation is broken down into two phases - a thread independent index which is calculated with one pass through the graph.
2. And a thread dependent index which is computed by propagating indices from nodes such as MMA, Read or Write.
3. A heuristic is added to determine how to propagate information in case of multiple nodes competing for a given node based on dimensional analysis
4. An additional unit test is added for attention expansion
5. An e2e test for attention with bias is added

### DIFF
--- a/iree/turbine/kernel/wave/index_sequence_analysis.py
+++ b/iree/turbine/kernel/wave/index_sequence_analysis.py
@@ -28,13 +28,13 @@ from .utils import (
     get_mma_dimensional_mapping,
     get_hardware_constraint,
     subs_idxc,
-    specialize_index_sequence,
-    capture_backward_slice,
+    get_inputs,
+    get_users,
 )
 import torch.fx as fx
 import numpy as np
 from functools import partial
-from typing import Sequence
+from typing import Sequence, Callable
 from ...support.logging import get_logger
 import sympy
 from itertools import groupby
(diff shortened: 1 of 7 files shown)


iree/turbine/kernel/wave/thread_shape_analysis.py
```diff
@@ -68,6 +68,17 @@ def propagatable_op(node: fx.Node):
     )
 
 
+def propagate_resolutions(
+    custom_node: CustomOp, dst_op: CustomOp = None
+) -> list[fx.Node]:
+    propagated_resolutions = capture_forward_slice(custom_node.fx_node, propagatable_op)
+    if dst_op:
+        for node in propagated_resolutions:
+            get_custom(node).index = dst_op.index
```
raikonenfnu, line 77: When would we use it without dst_op?

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -462,6 +379,226 @@ def set_node_index(
...
+        )
+    return [(node, index, hardware_constraint.vector_shapes)]
+
+
+def combine_indices(
+    thread_independent_index: dict[IndexSymbol, IndexSequence],
+    thread_dependent_index: dict[IndexSymbol, IndexSequence],
+) -> dict[IndexSymbol, IndexSequence]:
+    combined_index = {k: v for k, v in thread_independent_index.items()}
+    for k in combined_index:
```
raikonenfnu, line 473: Would be nice to add some docs, to different examples/cases on when we'd add the start offsets. (perhaps read -> read or offseted_read + mma?)

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -462,6 +379,226 @@ def set_node_index(
...
+        for arg in args:
+            custom = get_custom(arg)
+            if isinstance(custom, (Allocate, Placeholder)) and not isinstance(
+                custom, IterArg
+            ):
+                continue
+            vector_shapes = (
+                custom.vector_shapes if custom.vector_shapes else source_vector_shapes
+            )
+            sources.append((custom, source_index, vector_shapes))
```
raikonenfnu, line 507: why not use source.index? Seems like the only time sourc.index != source_index is when source.index has underlying indices before getting propagated. But wouldn't we want that?

raikonenfnu (approved):
Overall looks good, just couple clarifying questions

Hardcode84 (approved):

iree/turbine/kernel/wave/thread_shape_analysis.py
harsh-nod: Good question. This was when I was using it in 2 places where dst_op was optional. I can modify this so that it doesn't handle that case.

iree/turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: In this PR, we split the index assignment into 2 phases. The first is a thread-independent index assignment where we set the indices based on work group constraints and tiling constraints and anything that in general does not have any thread level dependence. Once we have this index then we go through and propagate the thread dependent index which comes either from MMA nodes or if there are no MMA nodes it comes from reads and writes and we then add the thread dependent index to the thread independent index. So we are always going to be adding these two offsets together.

iree/turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: So source.index contains the unified index which is the sum of the thread dependent index and the thread independent index. During propagation we only want to propagate the thread dependent index and that's why we only propagate source_index.

PR #295, opened 2024-11-26
[TKW] Move IGEMM conv impl to common place.

Hardcode84:
Move TKW IGEMM conv impl from the test folder to some common place to allow it to be reused outside the tests (e.g. in iree-kernel-benchmark).

Not sure what the proper place for it, suggestions are welcome.

### DIFF
--- a/iree/turbine/kernel/kernels.py
+++ b/iree/turbine/kernel/kernels.py
@@ -1,0 +1,158 @@
+# Copyright 2024 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import iree.turbine.kernel as tk
+import iree.turbine.kernel.lang as tkl
+import iree.turbine.kernel.wave as tkw
+from typing import Any
+from iree.turbine.kernel.lang.global_symbols import *
+
+
+def get_igemm_conv2d(
+    layout: str,
+    n: int,
+    h: int,
+    w: int,
+    c: int,
+    hf: int,
+    wf: int,
+    nf: int,
+    stride: int,
+    mem_space: tkl.IndexSymbol = SHARED_ADDRESS_SPACE,
+) -> tuple["LaunchableWave", dict[Any, Any]]:
+    cf = c
+    padding = 0  # TODO: only pad=0 is supported for now
+
+    sym = tkl.sym
+    N, C, H, W = sym.N, sym.C, sym.H, sym.W
+    NF, HF, WF = sym.NF, sym.HF, sym.WF
+
+    H_OUT = (H + 2 * padding - HF) // stride + 1
+    W_OUT = (W + 2 * padding - WF) // stride + 1
+    SZ_OUT = H_OUT * W_OUT
+
+    K = HF * WF * C
+    M = SZ_OUT * N
+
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/kernels.py
```diff
@@ -0,0 +1,158 @@
+# Copyright 2024 The IREE Authors
```
raikonenfnu, line 1: Maybe we can put it in `iree/turbine/kernel/wave/templates/igemm.py` or `iree/turbine/kernel/wave/templates/conv.py`. Here are some of my thoughts/ideas behind this:

1. Reasoning under `iree/turbine/kernel/wave/` because it is specifically a `wave` kernel and with the current state `iree/turbine/kernel/` is not wave exclusive. 
2. Under `iree/turbine/kernel/wave/template` because it will be somewhat uniform to the way shark-ai stack is organizing their kernels/template kernels. [(example link)](https://github.com/nod-ai/shark-ai/tree/main/sharktank/sharktank/kernels)
3. Should be in it's own file `igemm.py`/`conv.py`, because in the future, we may store different variants of igemm or conv (such as igemm, direct convolution, backward convs, etc). In a similar fashion that in the future, we may have `attention.py` that has different variants such as vanilla, paged, quantized, evoform.

iree/turbine/kernel/kernels.py
```diff
@@ -0,0 +1,158 @@
...
+                elements_per_thread=ELEMS_PER_THREAD,
+            )
+            acc = tkw.mma(a_reg, b_reg, acc)
+            return acc
+
+        tkw.write(
+            repeat, out, mapping=out_mapping, elements_per_thread=ELEMS_PER_THREAD
+        )
+
+    symbols = {
```
raikonenfnu, line 135: Can we move the `symbols: dict[Symbol, int]` and other non-symbolic/materialized variables out of the kernels.py file into the driver file? 

Looking at it from a perspective of a library writer, I'd probably want to expose ratios, block_m, block_n, block_k and all these other stuff in the driver such that I can come up with my own heuristics/tunings/configs for igemm depending on the input size.

raikonenfnu (commented):
This is a nice idea Ivan! I have some notes/suggestions though. :)

iree/turbine/kernel/kernels.py
Hardcode84: Good idea, done.

iree/turbine/kernel/kernels.py
Hardcode84: Without this there is no way for user to know which symbols are actually used inside the kernel. My idea was to return symbols dict with some sensible defaults and user then can tweak it on their side.

iree/turbine/kernel/wave/templates/conv.py
```diff
@@ -0,0 +1,158 @@
...
+def get_igemm_conv2d(
+    layout: str,
+    n: int,
+    h: int,
+    w: int,
+    c: int,
+    hf: int,
+    wf: int,
+    nf: int,
+    stride: int,
```
raikonenfnu, line 23: can we add block_m, block_n, block_k, ratio_m, ratio_n as optional arg with default values?

raikonenfnu: We can add assertions and heuristics in upcoming PR :)

Hardcode84: Done

raikonenfnu (approved):

PR #296, opened 2024-11-26
[TKW] Hoist loop invariant reads

raikonenfnu:
In flash attention, Q's reduction dimension is typically relatively small, and hence we do only have reduction tile across K2 dimension/reduction dimension of 2nd gemm(P and V). Hence, an optimization we can do is to hoist reading of Q from global memory out of the for loop, this actually generates quite a big speedup (hoistQ + use global->register for Q gives typically 2x speed up.)

To implement the optimization above, we needed to add:

1. Expand `hoisting.py` to also look for `Read` that is:
     - independent of induction variable
     - has memory that is read-only/do not have write as users (i.e important for correctness since this guarantee data being read is constant/not changing with loop)
     - has memory who is a captured_var (i.e memory can be traced to outside the loop) 
2. Implement method to hoist reads properly:
    - Copy Read to rootOp
    - replace rootOp's memory who is a `captured_var` with it's counterpart in the RootOp by querying reduction's `implicit_capture`
    - Remove unused `captured_var` from `Reduction` otherwise scf.for will be indexing/loading from the wrong bindings.
3. Updated specifically chained_gemm_tests in `lit_tests/codegen.py` to test for the hoisted reads from global. 
4. Updated `lit_tests/attention.py` since this change generates new schedule

### DIFF
--- a/iree/turbine/kernel/wave/hoisting.py
+++ b/iree/turbine/kernel/wave/hoisting.py
@@ -43,13 +43,15 @@ def get_hoistable_ops(
                 continue
             # Only handle case where we are not writing to the same memory.
             # Counterproof: we may expect different read if we write to same memory.
-            for memory_user in custom_node.memory.users:
-                if isinstance(get_custom(memory_user), Write):
-                    continue
+            if any(
+                isinstance(get_custom(mem_user), Write)
+                for mem_user in custom_node.memory.users
+            ):
+                continue
             # Only hoist Read that is loop invariant.
-            dims_indexing = [ind.start for ind in custom_node.index.values()]
-            dim_depends_on_ivar = [ind.has(induction_variable) for ind in dims_indexing]
-            if any(dim_depends_on_ivar):
+            if any(
+                ind.start.has(induction_variable) for ind in custom_node.index.values()
+            ):
                 continue
             hoistable_ops.append(custom_node)
         else:
--- a/iree/turbine/kernel/wave/hoisting.py
+++ b/iree/turbine/kernel/wave/hoisting.py
@@ -4,6 +4,8 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+from .constraints import Constraint
+from .utils import get_induction_variable
 from ...support.logging import get_logger
 from iree.turbine.kernel._support.tracing import CapturedTrace
 import torch.fx as fx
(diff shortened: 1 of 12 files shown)


lit_tests/kernel/wave/codegen.py
```diff
@@ -1051,8 +1051,12 @@ def repeat(
         output = torch.zeros(8, 64, 128, dtype=torch.float32)
         print(chained_gemm(q, k, v, output).module_op)
 
-        # CHECK-LABEL:     func.func @chained_gemm(
+        # CHECK-LABEL:     func.func @chained_gemm
+        # CHECK-SAME:        (%[[ARG0:[a-zA-Z0-9_]+]]: !stream.binding
```
Hardcode84, line 1055: nit: `[[ARG0:.*]]`

lit_tests/kernel/wave/codegen.py
```diff
@@ -1051,8 +1051,12 @@ def repeat(
...
+        # CHECK:             %[[GLOBAL_0:.+]] = stream.binding.subspan %[[ARG0]]
+        # CHECK-COUNT-4:     vector.load %[[GLOBAL_0]]
         # CHECK:             {{.*}} = scf.for
+        # CHECK-COUNT-4:       {{.*}} = vector.load %alloc
```
Hardcode84, line 1059: `%alloc` SSA name may not be stable, either `%{{.*}}` or explicitly match alloc before

iree/turbine/kernel/wave/hoisting.py
```diff
@@ -13,25 +15,85 @@
...
+                    continue
+            # Only hoist Read that is loop invariant.
+            dims_indexing = [ind.start for ind in custom_node.index.values()]
+            dim_depends_on_ivar = [ind.has(induction_variable) for ind in dims_indexing]
+            if any(dim_depends_on_ivar):
+                continue
+            hoistable_ops.append(custom_node)
+        else:
+            continue
+    all_hoistables_ops = hoistable_allocs + hoistable_ops
```
Hardcode84, line 57: I think we can unify them into single list

Hardcode84 (approved):
LGTM besides couple comments

iree/turbine/kernel/wave/hoisting.py
raikonenfnu: The main reason why it is important to have two list is because we want to ensure that allocs are processed/hoisted out first. This is to avoid using alloc before defined/non-dominating behavior. (e.g hoisting read from global to shared before shared alloc is defined.)

lit_tests/kernel/wave/codegen.py
raikonenfnu: I too prefer (and started with) this haha, but for some reason with` :.*` it is failing to find it 🤔

Hardcode84: Maybe it matched multiple args? You can try to explicitly match all the `chained_gemm` args

iree/turbine/kernel/wave/hoisting.py
harsh-nod: oh I missed the other comment, yes in that case makes sense to separate them out.

PR #298, opened 2024-11-26
[TKW] Move default scheduling params to common place

Hardcode84:
Moved default scheduling params to `get_default_scheduling_params()`.

### DIFF
--- a/tests/kernel/wave/wave_e2e_test.py
+++ b/tests/kernel/wave/wave_e2e_test.py
@@ -14,6 +14,7 @@ from iree.turbine.kernel.wave.iree_utils import generate_iree_ref
 from iree.turbine.kernel.wave.utils import (
     get_default_arch,
     get_default_run_config,
+    get_default_scheduling_params,
     device_randn,
     device_randint,
     device_randperm,
@@ -922,7 +923,7 @@ def test_igemm_conv(n, h, w, c, hf, wf, nf, stride, mem_space, layout, request):
     else:
         raise ValueError(f"Invalid layout: {layout}")
 
-    conv, symbols = get_igemm_conv2d(
+    conv, hyperparams = get_igemm_conv2d(
         layout=layout,
         n=n,
         h=h,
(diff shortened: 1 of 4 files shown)


harsh-nod (approved):
thanks for the cleanup! Was meaning to do this, but never got to it.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -994,6 +995,22 @@ def test_igemm_conv(n, h, w, c, hf, wf, nf, stride, mem_space, layout, request):
...
+        NF: nf,
+        WF: wf,
+        HF: hf,
+        BLOCK_M: 64,
+        BLOCK_N: 128,
+        BLOCK_K: 32,
+        ELEMS_PER_THREAD: 4,
+        ADDRESS_SPACE: mem_space,
+    }
+    hyperparams.update(get_default_scheduling_params())
```
raikonenfnu, line 1012: can't we just integrate into the backend directly? since we may use this area to override default variables

raikonenfnu (changes requested):

tests/kernel/wave/wave_e2e_test.py
raikonenfnu: Would be good to add it here
https://github.com/iree-org/iree-turbine/blob/e3b6c878f3f0ee26660e6d4762879bff0c8394a3/iree/turbine/kernel/wave/scheduling/resources.py#L32-L40

So instead of `return np.array([int(subs_idxc(x)) for x in resources]) ` we can do something like:
```python
default_resources =  get_default_scheduling_params()
return np.array([int(subs_idxc(x)) if x in subs_map else default_resources[x] for x in resources]) 
```

and similar method for the delays as well. I think if we can do this way we would not need to update hyperparameter for every kernel and it makes overwriting default values cleaner. :)

Hardcode84: Those values are not really 'default' as they can be different for different GPUs, it's more like `get_default_device_scheduling_params()` so I think it's better to set them explicitly.

raikonenfnu (approved):
discussed offline, would add default params to backend resource.py in future

PR #300, opened 2024-11-27
[TKW] Add `create_vmfb_file` option

Hardcode84:
Add a `create_vmfb_file` flag to run full IREE pipeline to get vmfb file without actually invoking anything.

This is a quick hack to get iree-kernel-benchmark going until we have a proper AOT compilation API.

### DIFF
--- a/iree/turbine/kernel/wave/utils.py
+++ b/iree/turbine/kernel/wave/utils.py
@@ -39,6 +39,7 @@ from .constraints import (
 from .assumptions import Assumption
 import torch.fx as fx
 import iree.turbine.kernel.lang as tkl
+from pathlib import Path
 
 
 import tempfile
@@ -522,6 +523,7 @@ def compile_and_invoke(
     kernel_dynamic_dims: list[int] = [],
     run: bool = False,
     run_bench: bool = False,
+    create_vmfb_file: Optional[Path] = None,
     inplace: bool = False,
 ):
     backend = config["backend"]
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/wave/utils.py
```diff
@@ -572,23 +572,25 @@ def compile_and_invoke(
...
-    vm_modules = [
-        mod,
-        rt.create_hal_module(vm_instance, device),
-    ]
-    ctx = rt.SystemContext(
-        vm_modules=vm_modules,
-        config=rt_config,
-    )
+
+    if run or run_bench:
```
harsh-nod, line 576: minor: Following llvm style, maybe exit early instead?
```
if not (run or run_bench):
   return
```

iree/turbine/kernel/wave/wave.py
```diff
@@ -336,7 +336,8 @@ def test_execute(self, args, kwargs):
 
         run = kwargs.get("run", False)
         run_bench = kwargs.get("run_bench", False)
-        if run or run_bench:
+        dry_run = kwargs.get("dry_run", False)
```
harsh-nod, line 339: Do we even need this flag? If you set run = False and run_bench = False and set dump_vmfb_file, doesn't that work?

harsh-nod (changes requested):

iree/turbine/kernel/wave/wave.py
Hardcode84: If we set both `run` and `run_bench` to `False` we will never enter this path. Although, as you mentioned it, I think instead of adding `dry_run` flag I can instead promote `dump_vmfb` flag from run config to general flag. Let me try it.

Hardcode84: done

harsh-nod: thanks

iree/turbine/kernel/wave/utils.py
```diff
@@ -39,6 +39,7 @@
 from .assumptions import Assumption
 import torch.fx as fx
 import iree.turbine.kernel.lang as tkl
+from pathlib import Path
```
harsh-nod, line 42: unused import?

harsh-nod (approved):
lgtm!

PR #301, opened 2024-11-27
[TKW] Add dtype to igemm conv template

Hardcode84:
Add `input_dtype`/`output_dtype` to igemm_conv template API, only f16xf16xf32 is supported for now but it will be extended in future.


harsh-nod (approved):
lgtm!

PR #304, opened 2024-12-02
[TKW] Detect contiguous access pattern in mapped reads/writes

Hardcode84:
Add utility to detect contiguous access pattern in mapped reads/writes so we can use contiguous vector ops instead of gathers/scatters during lowering.

Iterate over `elements_per_thread` and check if every subsequent mapped index have only diff of 1 in fastest changing dim.

Also, it need some custom sympy simplifications to successfully work on IGEMM conv.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -976,6 +976,35 @@ class Read(CustomOp):
         if is_shared_mem_access(self):
             self.index = align_index_vars(self.index, constraints)
 
+    def has_identity_mapping(self) -> bool:
+        mapping = self.mapping
+        if mapping is None:
+            return True
+
+        mem_shape = self.memory.type.symbolic_shape
+        if mapping.is_identity() and mapping.input_shape == mem_shape:
+            return True
+
+        return False
+
+    def is_contiguous_vec(self) -> bool:
+        if self.has_identity_mapping():
+            return True
+
+        mapping = self.mapping
+
+        mem_shape = self.memory.type.symbolic_shape
+
+        from ..wave.utils import check_is_mapping_contiguous
+
+        return check_is_mapping_contiguous(
+            mapping=mapping,
+            symbolc_shape=mem_shape,
+            index=self.index,
+            elements_per_thread=self.elements_per_thread,
+            is_read=True,
+        )
+
 
 @define_op("reduction")
 @dataclass
@@ -1151,6 +1180,34 @@ class Write(CustomOp):
 
         if is_shared_mem_access(self):
             self.index = align_index_vars(self.index, constraints)
+
+    def has_identity_mapping(self) -> bool:
+        mapping = self.mapping
+        if mapping is None:
+            return True
+
+        mem_shape = self.memory.type.symbolic_shape
+        if mapping.is_identity() and mapping.output_shape == mem_shape:
+            return True
+
+        return False
+
+    def is_contiguous_vec(self) -> bool:
+        if self.has_identity_mapping():
+            return True
+        mapping = self.mapping
+
+        mem_shape = self.memory.type.symbolic_shape
+
+        from ..wave.utils import check_is_mapping_contiguous
+
+        return check_is_mapping_contiguous(
+            mapping=mapping,
+            symbolc_shape=mem_shape,
+            index=self.index,
+            elements_per_thread=self.elements_per_thread,
+            is_read=False,
+        )
 
 
 @define_py_op(operator.getitem)
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -977,6 +977,7 @@ class Read(CustomOp):
             self.index = align_index_vars(self.index, constraints)
 
     def has_identity_mapping(self) -> bool:
+        """Check if mapping between input memory and output register is identity."""
         mapping = self.mapping
         if mapping is None:
             return True
@@ -988,6 +989,9 @@ class Read(CustomOp):
         return False
 
     def is_contiguous_vec(self) -> bool:
+        """Check if op can be lowered to contiguous vector ops
+
+        If False we will have to lower it to gather"""
         if self.has_identity_mapping():
             return True
 
@@ -1182,6 +1186,7 @@ class Write(CustomOp):
             self.index = align_index_vars(self.index, constraints)
 
     def has_identity_mapping(self) -> bool:
+        """Check if mapping between input register and output memory is identity."""
         mapping = self.mapping
         if mapping is None:
             return True
@@ -1193,6 +1198,9 @@ class Write(CustomOp):
         return False
 
     def is_contiguous_vec(self) -> bool:
+        """Check if op can be lowered to contiguous vector ops
+
+        If False we will have to lower it to gather"""
         if self.has_identity_mapping():
             return True
         mapping = self.mapping
--- a/iree/turbine/kernel/wave/codegen.py
+++ b/iree/turbine/kernel/wave/codegen.py
@@ -610,9 +610,9 @@ def _construct_gather_scatter_indices(
 
     # As we only support identity input/output mapping for now, we can directly
     # substitute iterators with corresponding expanded index.
-    subs = list(idxc.subs.items()) + [
+    subs = [
         (sym, expr.start) for sym, expr in zip(iters.keys(), index.values())
-    ]
+    ] + list(idxc.subs.items())
 
     # Contruct input/output index, substituting iterators in input mapping with
     # expanded index.
--- a/iree/turbine/kernel/wave/codegen.py
+++ b/iree/turbine/kernel/wave/codegen.py
@@ -177,6 +177,20 @@ class WaveEmitter:
         )
         self._node_values[node] = proxies
 
+    def get_induction_vars_and_syms(self) -> tuple[list[OpResult], list[IndexExpr]]:
+        induction_var_syms = []
+        induction_vars = []
+        if self.induction_vars:
+            for constraint in self.constraints:
+                if isinstance(constraint, TilingConstraint):
+                    assert (
+                        constraint.dim in self.induction_vars
+                    ), f"Could not find induction var for {constraint.dim} dimension"
+                    induction_var_syms.append(constraint.induction_var)
+                    induction_vars.append(self.induction_vars[constraint.dim])
+
+        return induction_vars, induction_var_syms
+
 
 def get_type_or_element_type(operand_type: IrType):
     assert isinstance(operand_type, IrType)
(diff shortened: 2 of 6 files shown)


iree/turbine/kernel/wave/utils.py
```diff
@@ -1043,3 +1044,208 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
...
+            if not isinstance(arg, (sympy.floor, sympy.Mod)):
+                return None
+
+        return ret
+
+    def transform_mod(expr):
+        """Move constant outside of Mod expr
+
+        Example:
+        (floor(a) * 4 + 3) % 16 -> (floor(a) * 4) % 16 + 3
```
harsh-nod, line 1089: This transformation isn't true even for positive numbers.
See this example:
```
(2 + 2) % 3 = 1 which is not equal to (2 % 3) + (2 % 3)
```
However, what you can say is
```
(a + b) % M = ((a % M) + (b % M)) % M
```

Hardcode84: I'm matching expr `(floor(x) * a + b) % c` and checking `c % a == 0` and `b < a`

iree/turbine/kernel/wave/utils.py
```diff
@@ -1043,3 +1044,208 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
...
+            if not isinstance(arg, (sympy.floor, sympy.Mod)):
+                return None
+
+        return ret
+
+    def transform_floor(expr):
+        """Simplify rational addition inside floor expr
+
+        Example:
+        floor(floor(a)/3 + 1/6) -> floor(floor(a)/3)
```
harsh-nod, line 1146: Here as well, we can only state that
```
floor(x + y) = floor(x)
```
if 
```
0 <= y < 1 - (x - floor(x)) 
```

iree/turbine/kernel/wave/utils.py
```diff
@@ -1043,3 +1044,208 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
...
+) -> list[IndexExpr]:
+    start_indices = []
+    for dim_indexing in src_indices:
+        i = _get_start_index(src_indices[dim_indexing])
+        start_indices.append(i)
+
+    return start_indices
+
+
+def _simplify_sympy_expr(expr: IndexExpr) -> IndexExpr:
```
harsh-nod, line 1067: Screenshot 2024-12-03 at 3 16 46 PM
 How about something like this?

harsh-nod: To accomplish this, you would do the following
```
> next_indices[2] - prev_indices[2]
-floor(floor((Mod($T0, 4))/80 + floor($ARGC*HF*WF*BLOCK_K/8)/80)/3) + floor(floor((Mod($T0, 4))/80 + floor($ARGC*HF*WF*BLOCK_K/8)/80 + 1/640)/3)

sympy.cse(next_indices[2] - prev_indices[2])[0]
[(x0, (Mod($T0, 4))/80 + floor($ARGC*HF*WF*BLOCK_K/8)/80)]
sympy.cse(next_indices[2] - prev_indices[2])[1]
[-floor(floor(x0)/3) + floor(floor(x0 + 1/640)/3)]
```

iree/turbine/kernel/wave/utils.py
harsh-nod: Makes sense, but I'm not sure that is enough for correctness. What you want is
```
(floor(x) * a + b) % c == (floor(x) * a) % c + b % c
```
which is only true when either b is a multiple of c, or 
```
 (floor(x) * a) % c + (b % c) < c
```

iree/turbine/kernel/wave/utils.py
Hardcode84: So, as any sane person should do, I wrote a fuzzing tests (see latest commit) and it indeed found some failing cases. But after stricten some checks (and marking induction vars as non-negative) and running 100 000s tests I'm now reasonable sure my transformations are correct.

harsh-nod: Thanks that sounds good, but I think we have been thinking about this the wrong way.

So essentially what we are looking to do here is determine whether the mapping is "contiguity-preserving" (which means that elements that are contiguous pre-mapping will remain contiguous post-mapping). This is a function entirely of the mapping and should not depend at all on the particular access pattern of the underlying data. So I think these patterns to simplify the index sequences are not required and instead we should be looking just at the mapping. For example, looking at one IGEMM mapping which maps a tensor of shape [M, K] to a tensor of shape [N, H, W, C]

```
x_mapping = tkw.IndexMapping(
        num_iterators=2,
        inputs={
            N: i // SZ_OUT,
            C: j % C,
            H: (i % SZ_OUT) % W_OUT * stride + (j // C) % WF,
            W: (i % SZ_OUT) // W_OUT * stride + (j // C) // WF,
        },
        outputs={M: i, K: j},
    )
```

we can immediately deduce that this mapping is contiguity-preserving because j gets mapped to j % C.

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -976,6 +976,35 @@ def align_index(self, constraints: list["Constraint"]) -> None:
...
+    def has_identity_mapping(self) -> bool:
+        mapping = self.mapping
+        if mapping is None:
+            return True
+
+        mem_shape = self.memory.type.symbolic_shape
+        if mapping.is_identity() and mapping.input_shape == mem_shape:
+            return True
+
+        return False
```
harsh-nod, line 988: How about
```
def has_identity_mapping(self) -> bool:
        return self.mapping is None or (
            self.mapping.is_identity() and self.mapping.input_shape == self.memory.type.symbolic_shape
        )
```

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1152,6 +1181,34 @@ def align_index(self, constraints: list["Constraint"]) -> None:
         if is_shared_mem_access(self):
             self.index = align_index_vars(self.index, constraints)
 
+    def has_identity_mapping(self) -> bool:
```
harsh-nod, line 1184: Since these are identical, can we add them to the CustomOp or you could create a new base class for Read and Write?

iree/turbine/kernel/wave/codegen.py
```diff
@@ -621,16 +610,34 @@ def _construct_gather_scatter_indices(
 
     # As we only support identity input/output mapping for now, we can directly
     # substitute iterators with corresponding expanded index.
-    subs = [
+    subs = list(idxc.subs.items()) + [
```
harsh-nod, line 613: Is this only for stylistic reasons or does the order matter?

iree/turbine/kernel/wave/utils.py
```diff
@@ -1055,3 +1056,220 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
...
+        ), "non-identity input mapping is not supported yet"
+        index_mapping = mapping.map_output_indices(symbolc_shape)
+
+    index_mapping = tuple(subs_idxc(i) for i in index_mapping)
+
+    iters = mapping.iters
+
+    subs = [(sym, expr.start) for sym, expr in zip(iters.keys(), index.values())]
+
+    result_index = {key: m.subs(subs) for key, m in zip(symbolc_shape, index_mapping)}
```
harsh-nod, line 1247: This is not used anywhere?

iree/turbine/kernel/wave/utils.py
```diff
@@ -1055,3 +1056,220 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
...
+    expected_diff[-1] = 1
+
+    # Assume fastest changing dim increments in elements_per_thread between individual ops,
+    # This is needed for sympy simplifications.
+    subs[-1] = (subs[-1][0], (subs[-1][1] // elements_per_thread) * elements_per_thread)
+    prev_indices = _get_start_indices(
+        {key: m.subs(subs) for key, m in zip(symbolc_shape, index_mapping)}
+    )
+    for i in range(1, elements_per_thread, 1):
+        # Increment fastests changing dim in unmapped index by one and apply mapping.
```
harsh-nod, line 1261: minor typo fastests

iree/turbine/kernel/wave/utils.py
```diff
@@ -1055,3 +1056,220 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
...
+    expected_diff = [0] * len(index_mapping)
+    expected_diff[-1] = 1
+
+    # Assume fastest changing dim increments in elements_per_thread between individual ops,
+    # This is needed for sympy simplifications.
+    subs[-1] = (subs[-1][0], (subs[-1][1] // elements_per_thread) * elements_per_thread)
+    prev_indices = _get_start_indices(
+        {key: m.subs(subs) for key, m in zip(symbolc_shape, index_mapping)}
+    )
+    for i in range(1, elements_per_thread, 1):
```
harsh-nod, line 1260: Could you add a comment why this starts with 1?

iree/turbine/kernel/wave/utils.py
```diff
@@ -1055,3 +1056,220 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
...
+    result_index = {key: m.subs(subs) for key, m in zip(symbolc_shape, index_mapping)}
+
+    # Iterate over elements_per_thread end check if every subsequent read have
+    # diff 1 in fastest changing dim and 0s in every other.
+    expected_diff = [0] * len(index_mapping)
+    expected_diff[-1] = 1
+
+    # Assume fastest changing dim increments in elements_per_thread between individual ops,
+    # This is needed for sympy simplifications.
+    subs[-1] = (subs[-1][0], (subs[-1][1] // elements_per_thread) * elements_per_thread)
```
harsh-nod, line 1256: Could you add a comment explaining what sympy simplifications this enables?

harsh-nod (changes requested):
Just some minor changes, but overall looks good. How did the cse simplification help?

iree/turbine/kernel/ops/wave_ops.py
Hardcode84: They are not totally identical, I use `mapping.input_shape` for `read` and `mapping.output_shape` for `write`.

iree/turbine/kernel/ops/wave_ops.py
Hardcode84: I really tied to keep it withing 80 char width, not sure if such long conditional expressions are more readable.

iree/turbine/kernel/wave/utils.py
Hardcode84: yeah, some copypaste leftovers, removed

iree/turbine/kernel/wave/codegen.py
Hardcode84: left from some experiments, removed

iree/turbine/kernel/wave/utils.py
Hardcode84: Expanded comments somewhat

iree/turbine/kernel/wave/utils.py
Hardcode84: yeah, `sympy.cse` looks useful, but I weren't able construct working code quickly. I will need to think more on it offline.

harsh-nod: sounds good, let's land this PR. I have a WIP PR that should simplify this a lot use sympy.cse.

iree/turbine/kernel/wave/codegen.py
harsh-nod: thanks

iree/turbine/kernel/wave/utils.py
harsh-nod: thanks

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: I think this can still be refactored but we can save that for a future PR.

harsh-nod (approved):
let's land for now. I think we can greatly simplify the checks by using sympy.cse but we can explore that after landing this PR.

PR #309, opened 2024-12-04
Add support for waves_per_eu flag

harsh-nod:
The waves_per_eu flag is a optimization hint
that specifies the minimum number of waves per
execution unit.


Hardcode84:
Can we infer it from WG/BLOCK sizes constraints?

raikonenfnu:
> Can we infer it from WG/BLOCK sizes constraints?

I think in IREE, we'd just typically default to 2 so would be OK for now

Hardcode84 (approved):

PR #310, opened 2024-12-04
Disable PERF CI

Hardcode84:


ScottTodd:
Why? We can also disable the workflow from the UI: https://github.com/iree-org/iree-turbine/actions/workflows/perf.yaml

image

Hardcode84:
This workflow currently just runs the same tests as normal CI. I hope to fix it in the future but for now we can just disable it to make PR CI faster. I wanted to make code-only changes to reenable it later but I guess disabling it in UI is easier.

ScottTodd:
These tests are different from other workflows: https://github.com/iree-org/iree-turbine/blob/f40670394b060edfe7c6f210afb1fb3250b2722c/.github/workflows/perf.yaml#L58-L63

Hardcode84:
> These tests are different from other workflows: 

Yeah, but it doesn't run with `--runperf` and doesn't actually collect any perf data. And we really should get rid of json files and fully switch all tests to `--runperf`. It is are not very useful in its current form.

PR #311, opened 2024-12-04
Add ml_dtypes as a dependency

harsh-nod:
This will provide access to ML dtypes
like bfloat16 that are not supported by
numpy.


saienduri (approved):

Hardcode84 (approved):

ScottTodd (approved):
Thanks! I was going to send a PR for this myself 😅

PR #318, opened 2024-12-05
[TKW] Introduce ReciprocalOp for faster attention

raikonenfnu:
Doing divf is very expensive on the GPU. We introduce reciprocalOp S.T we can use it to do a single divf on online attention's sum, which we can then apply to the attention output for normalization using mul.

This will allow us to only do a divf on a single f32 as opposed to the entire vector. Which will give us speed up of around 0.1 ms per dispatch on dispatch146(B0: 2, B1: 20, (M, K2): 1024: K1: 64) (after vectorized write land it drops from 5.53ms -> 5.38 ms)

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -102,6 +102,10 @@ def exp2(src: "Register") -> "Register":
     ...
 
 
+def reciprocal(src: "Register") -> "Register":
+    ...
+
+
 def maximum(lhs: "Register", rhs: "Register") -> "Register":
     ...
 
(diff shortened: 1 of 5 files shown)


harsh-nod (approved):
lgtm! One question here - do you think this will still be useful after we have scalar support or is this just a placeholder till then? If it is just a placeholder, could you add a TODO saying remove this after scalar support is added?

harsh-nod:
> I think having reciprocal is still nice/useful even after scalar support. It makes code a little more legible, and easier on shape analysis. Evidently both numpy and pytorch still have reciprocal desipte being able to support 1/torch.Tensor or np.array! :)

Makes sense, thanks

PR #319, opened 2024-12-05
[TKW] Implement pipeline to enable llvm_func_attrs and denormal_math

raikonenfnu:
For flash attention, denormal-fp-math-f32" = "preserve-sign" is very important optimization. With this flag, we are able to drop latency from 5.38ms to 4.75ms on dispatch146(B0: 2, B1: 20, (M, K2): 1024: K1: 64).

Added compile_config S.T we can use it as a parameter for kernel cacher as well.

### DIFF
--- a/iree/turbine/kernel/compiler/dispatch_codegen.py
+++ b/iree/turbine/kernel/compiler/dispatch_codegen.py
@@ -171,7 +171,7 @@ class StreamExecutable:
                         iree_codegen_d.DispatchLoweringPassPipeline.None_
                     )
                     translation_config = None
-                    if len(llvm_configs) > 0:
+                    if llvm_configs:
                         # Add llvm_func_attrs to translation config if any is specified.
                         llvm_func_attrs = DictAttr.get(
                             {k: StringAttr.get(str(v)) for k, v in llvm_configs.items()}
--- a/iree/turbine/kernel/compiler/dispatch_codegen.py
+++ b/iree/turbine/kernel/compiler/dispatch_codegen.py
@@ -21,6 +21,7 @@ from .builder import (
 from .ir import (
     Attribute,
     Block,
+    DictAttr,
     FunctionType,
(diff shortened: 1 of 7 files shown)


iree/turbine/kernel/compiler/dispatch_codegen.py
```diff
@@ -168,10 +170,23 @@ def abi_type(binding: BindingDesc):
                     pipeline_attr = iree_codegen_d.DispatchLoweringPassPipelineAttr.get(
                         iree_codegen_d.DispatchLoweringPassPipeline.None_
                     )
+                    translation_config = None
+                    if len(llvm_configs) > 0:
```
harsh-nod, line 174: minor: you can write this is as if not llvm_configs:

harsh-nod (approved):
lgtm! thanks I like these kind of optimizations :)

harsh-nod:
One request and doesn't have to be in this PR, but can you also add this to the evoformer? Same for the other optimizations.

iree/turbine/kernel/compiler/dispatch_codegen.py
raikonenfnu: sg and done, although it is `if llvm_configs` since we are checking for not 0.

raikonenfnu:
> One request and doesn't have to be in this PR, but can you also add this to the evoformer? Same for the other optimizations.

Done! :)

Hardcode84 (approved):

iree/turbine/kernel/compiler/dispatch_codegen.py
harsh-nod: thanks

harsh-nod:
> > One request and doesn't have to be in this PR, but can you also add this to the evoformer? Same for the other optimizations.
> 
> Done! :)

thanks

PR #320, opened 2024-12-06
[TKW] Propagate dynamic mapping index

Hardcode84:


iree/turbine/kernel/_support/indexing.py
```diff
@@ -411,18 +411,23 @@ class IndexSequence:
...
+        return value
 
-    def subs(self, map: dict[IndexSymbol, int]):
+    def subs(self, map: dict[IndexExpr, IndexExpr]):
         start = self._subs(self.start, map)
         size = self._subs(self.size, map)
         stride = self._subs(self.stride, map)
         return IndexSequence(start, size, stride)
 
+    def apply_expr(self, symbol: IndexExpr, expr: IndexExpr):
```
harsh-nod, line 426: Should we add some checks here to ensure that the size > 0 and stride is not negative?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -580,6 +580,14 @@ def align_index(self, constraints: list["Constraint"]) -> None:
         """
         pass
 
+    def transform_index_backwards(
```
harsh-nod, line 583: Can you combine this with transform_index and add arg as an optional argument? They seem similar in spirit which is that they both modify the index (for permute it comes from the semantics, and for read/write from the dynamic vals).

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -948,7 +956,7 @@ class Read(CustomOp):
     memory: fx.Proxy
     elements_per_thread: Optional[Any] = None
     mapping: Optional[IndexMapping] = None
-    mapping_dynamic_vals: tuple["Register", ...] = ()
+    mapping_dynamic_vals: tuple[fx.Node, ...] = ()
```
harsh-nod, line 959: Shouldn't this be a Register?

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -265,10 +265,48 @@ def has_gpr_offsets(node: fx.Node) -> bool:
...
 
 
+def combine_derived_index(
+    src_index: Optional[dict[IndexSymbol, IndexSequence]],
+    dst_index: dict[IndexSymbol, IndexSequence],
+) -> dict[IndexSymbol, IndexSequence]:
+    if src_index is None:
+        return dst_index
+
+    new_index = copy(src_index)
```
harsh-nod, line 275: Why shallow copy instead of deep copy here?

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -265,10 +265,48 @@ def has_gpr_offsets(node: fx.Node) -> bool:
...
+        assert (
+            old_idx.start == 0 or old_idx.start == old_idx.start
+        ), f"Index conflict: {old_idx} and {new_idx}"
+        new_index[dim] = new_idx
+
+    return new_index
+
+
+def set_derived_index(trace):
+    sources = trace.walk(lambda node: isinstance(get_custom(node), (Read, Write)))
```
harsh-nod, line 291: This is fine, but just wondering, did you try adding this to propagate_index()? I think these can be combined together.

harsh-nod (changes requested):
Mostly looks good, just a few comments. I think the big thing missing is some lit tests, but should be good to land after that.

iree/turbine/kernel/_support/indexing.py
Hardcode84: I wanted to have this method as straightforward as possible. Any pre/post conditions should be enforced on caller side (and `subs` doesn't do any checks either).

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -156,7 +156,7 @@ def has_gpr_offsets(node: fx.Node) -> bool:
         if not isinstance(custom, (Read, Write)):
             return False
         num_dims_with_gpr = sum(
-            1 for v in custom.index.values() if v.start.has(GPR_NUM)
+            1 for v in custom.index.values() if sympy.sympify(v.start).has(GPR_NUM)
```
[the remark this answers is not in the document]
harsh-nod: makes sense, thanks

iree/turbine/kernel/wave/index_sequence_analysis.py
harsh-nod: makes sense

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: I think in general we would load from memory to register and then use that as the dynamic value. So I think this should be register.

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: Actually, that's something I didn't quite understand - the difference between forward and backward. What is a forward transformation and what is a backward transformation? Aren't they all just transformations of the index sequence?

iree/turbine/kernel/ops/wave_ops.py
harsh-amd: This should be an fx.Node.

iree/turbine/kernel/ops/wave_ops.py
harsh-nod: So based on our discussion, the forward and backward transformations relate to apply a function f() and its inverse f_inv(). Could you add a comment describing this ?

harsh-nod (approved):
lgtm! thanks!

PR #321, opened 2024-12-06
[TKW][NFC] Change IndexSequence print format

Hardcode84:
Always print full form with sizes/strides to avoid ambiguity/confusion with sympy expr printing.


harsh-nod (approved):
lgtm! thanks!

PR #326, opened 2024-12-10
Run unit tests across Python versions 3.10-3.12.

ScottTodd:
Also updated enough of the project to pass these new test configurations:

* Avoid starred expressions in indexes for Python < 3.11 (https://stackoverflow.com/a/77080166, https://stackoverflow.com/q/77580216)
* Import `Self` from `typing_extensions` for Python < 3.11 (https://stackoverflow.com/a/77247460)
* Limit use of `logger.getLevelNamesMapping` to Python >= 3.11 (https://docs.python.org/3/library/logging.html#logging.getLevelNamesMapping, see also https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks)
* Stop using `unittest` in files that don't need it, fixing exit code 5 in Python >= 3.12 for "no tests were run" (https://docs.python.org/3/library/unittest.html#unittest.main)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ["3.10", "3.11", "3.12", "3.13"]
+        version:
+          - "3.10"
+          - "3.11"
+          - "3.12"
+
+          # Support for Python 3.13 depends on https://github.com/pytorch/pytorch/issues/130249
+          # - "3.13"
         os: [ubuntu-22.04]
     runs-on: ${{matrix.os}}
     env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,9 +68,8 @@ jobs:
         run: |
           pytest -n 4 --capture=tee-sys -vv .
 
-      # TODO: enable lit tests on 3.12 when they work (failing with "Exit Code: 5")
       - name: Run LIT tests
-        if: ${{ !cancelled() && matrix.version != '3.12' }}
+        if: ${{ !cancelled() }}
         run: |
           lit lit_tests/ -v
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,3 +72,23 @@ jobs:
         if: ${{ !cancelled() }}
         run: |
           mypy
+
+  # Depends on all other jobs to provide an aggregate job status.
+  ci_summary:
+    if: always()
+    runs-on: ubuntu-20.04
+    needs:
+      - test
+    steps:
+      - name: Getting failed jobs
+        run: |
+          echo '${{ toJson(needs) }}'
+          FAILED_JOBS="$(echo '${{ toJson(needs) }}' \
+            | jq --raw-output \
+            'map_values(select(.result!="success" and .result!="skipped")) | keys | join(",")' \
+          )"
+          echo "failed-jobs=${FAILED_JOBS}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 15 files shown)


lit_tests/kernel/wave/barriers.py
```diff
@@ -262,4 +262,14 @@ def test_gemm():
 
 if __name__ == "__main__":
     logging.basicConfig(level=logging.DEBUG)
-    unittest.main()
```
Hardcode84, line 265: None of the lit tests are actually using unittest (or pytest), I think, so you can just remove related code and imports entirely.

ScottTodd: Ah... That would certainly be simpler. Let me try that...

ScottTodd: That works. Thanks!

.github/workflows/ci.yaml
```diff
@@ -22,11 +22,17 @@ concurrency:
 
 jobs:
   test:
-    name: "Unit Tests and Type Checking"
+    name: "${{ matrix.os }} :: ${{ matrix.version }} :: Unit Tests and Type Checking"
```
ScottTodd, line 25: @marbre you may have opinions on the job naming. I'm not sure about this.

BTW, type checking does make sense to run on multiple python versions. The use of `logging.getLevelNamesMapping()  # Added in 3.11` was caught by running mypy on 3.10. I originally wanted to move type checking to pre-commit since it's similar to other lint checks, but I now think it makes sense to keep it here

.github/workflows/ci.yaml
```diff
@@ -22,11 +22,17 @@ concurrency:
...
       fail-fast: false
       matrix:
-        version: [3.11]
+        version:
+          - "3.10"
+          - "3.11"
+          - "3.12"
+
+          # Support for Python 3.13 depends on https://github.com/pytorch/pytorch/issues/130249
+          # - "3.13"
```
marbre, line 35: A more compact version could be
```suggestion
        # Support for Python 3.13 depends on https://github.com/pytorch/pytorch/issues/130249
        version: ["3.10", "3.11", "3.12"]
```

.github/workflows/ci.yaml
marbre: Following the naming in other repos we would have `"Unit Tests and Type Checking :: ${{ matrix.os }} :: ${{ matrix.version }}"` but to be honest I don't really care and if we start to dislike we can change it in the future. Thus I don't want to be picky here :wink:

marbre (approved):
Looks good and is a nice improvement!

.github/workflows/ci.yaml
ScottTodd: Yeah I'd like to put important information at the start of the name so it doesn't get cut off

* image
* image

marbre: Works for me :)

.github/workflows/ci.yaml
```diff
@@ -22,11 +22,12 @@ concurrency:
 
 jobs:
   test:
-    name: "Unit Tests and Type Checking"
+    name: "${{ matrix.os }} :: ${{ matrix.version }} :: Unit Tests and Type Checking"
```
ScottTodd, line 25: Argh GitHub. This was also being used for a required check matching `Unit Tests and Type Checking (3.11, ubuntu-22.04)`, but that check is actually picking up `TK CI / Unit Tests and Type Checking (3.11, ubuntu-22.04)` also.

I might just copy this snippet from IREE to have a "ci-summary" job with a unique name we can set as required: https://github.com/iree-org/iree/blob/7177c29f9b2d9e255b63987f5dfff174ec2afc2f/.github/workflows/ci.yml#L201-L236

PR #327, opened 2024-12-10
Add device queues to device affinities

rsuderman:


iree/turbine/aot/tensor_traits.py
```diff
@@ -20,16 +20,19 @@
 class DeviceAffinity:
     """This is used to provide device affinities to exported function arguments."""
 
-    def __init__(self, ordinal: int):
+    def __init__(self, ordinal: int, queues: Optional[list] = None):
```
sogartar, line 23: It would be helpful to know what type is to be expected for a queue.

sogartar (approved):

PR #329, opened 2024-12-11
[TKW] Kernel Cacher

raikonenfnu:
For eager mode to be viable, we'd need to implement Kernel Cacher S.T we do not need to re-compile kernels every time. Here are the main changes:

1. Refactor wave.py's compile_and_invoke to two separate functions `compile_to_vmfb`, and `invoke_vmfb` this is S.T we can intercept the compiled vmfb cleanly and store it to the caches.
2. Implement kernel cache dataclass which is a struct necessary to reconstruct kernels S.T it is invokable as original state
3. Implement fn to invoke kernel cache
4. Implement kernel cache manager that can hash, load/store kernel to RAM, load/store kernel to files.
5. Tests and helper fn for cache manager.

Let's discuss a little more about the newly developed cache manager, the Wave  cache manager has two main components/cache:

1. Session/Online cache - This is the main cache that our compiler and runtime will load from and store to. It is essentially a dict that uses the kernel hash as keys and the WaveCache as values. We added LRU functionality with limits for number of kernel cached here, because this lives on RAM, and we wouldn't want to run OOM.

2. File/Offline cache - This cache is essential for loading saved/compiled cache between sessions/runs. This is done by storing vital kernel information (vmfb, kernel_sig,and mlir) to CACHE_BASE_DIR/kernel_hash directory. If said kernel is queried during a new run and does not exist on session/online cache yet, we'd load files from the kernel_hash directory and reconstruct the WaveCache from it.


iree/turbine/kernel/wave/cache.py
```diff
@@ -0,0 +1,233 @@
...
+        config: dict[str, str],
+        use_scheduling: bool,
+        use_scheduling_barriers: bool,
+        run_bench: bool,
+    ):
+        """
+        Get a unique identifier for a given kernel.
+        """
+        key = [
+            inspect.getsource(kernel_body),
```
Hardcode84, line 78: Function source code is not always possible to retrieve and `inspect.getsource` will raise an error. We can live with this limitation for now, but in this case we need a way to bypass caching.

iree/turbine/kernel/wave/cache.py
```diff
@@ -0,0 +1,233 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Callable
+
+from .constraints import Constraint
+from ..compiler.kernel_codegen import KernelBufferUsage
+from .._support.indexing import IndexExpr
+from .utils import invoke_vmfb, _read_file, _write_file
+
+default_cache_base_dir = f"{str(Path.home())}/.wave"
```
Hardcode84, line 17: `Path.home() / ".wave"`

iree/turbine/kernel/wave/cache.py
raikonenfnu: do you know how/which cases it will fail to do inspect.getsource? I can add some checker there

iree/turbine/kernel/wave/cache.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+
+    cache_id: str
+    kernel_sig: tuple[KernelBufferUsage]
+    vmfb: bytes
+
+    @property
+    def module_op(self):
+        filepath = f"{CACHE_BASE_DIR}/{self.cache_id}/{self.cache_id}.mlir"
+        if not os.path.exists(filepath):
```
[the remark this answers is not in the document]
raikonenfnu: sounds good!

iree/turbine/kernel/wave/cache.py
raikonenfnu: sg and done!

iree/turbine/kernel/wave/cache.py
raikonenfnu: For now I have it on try and except not the nicest aesthetically but should do the job!

iree/turbine/kernel/wave/expansion.py
```diff
@@ -455,7 +455,10 @@ def _expand_mma_reduction(
     for dim in mma.indexing_dims:
         if dim not in dim_scaling and mma.vector_shapes[dim] > 0:
             tile_size = idxc.get_static_value(dim)
-            dim_scaling[dim] = max(tile_size // mma.vector_shapes[dim], 1)
+            try:
```
harsh-nod, line 458: debug artifact?

raikonenfnu: removed, good catch! :)

harsh-nod (approved):
lgtm! thanks for putting this together and will be a big boost to users. Just had some stylistic comments but looks good overall!

iree/turbine/kernel/wave/cache.py
```diff
@@ -0,0 +1,286 @@
...
+        self,
+        vmfb: bytes,
+        kernel_sig: tuple[KernelBufferUsage],
+        module_str: str,
+        kernel_hash: str,
+    ):
+        """
+        Save given kernel(vmfb, kernel_sig, and MLIR) into session_cache and file/offline cache.
+        """
+        if not WAVE_CACHE_ON or not kernel_hash:
```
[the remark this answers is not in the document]
raikonenfnu: I think we can store it to file cache, S.T we can use it to debug stuff (Triton has the exact same flag), which I had found useful. But I can make sure it doesn't store to the online/session cache.

iree/turbine/kernel/wave/wave.py
```diff
@@ -364,6 +366,41 @@ def _trace_and_get_kernel_signature(
...
+
+        # Get cached kernel when available.
+        cache_manager = get_cache_manager()
+        kernel_hash = cache_manager.get_hash(
+            self.constraints,
+            self._f,
+            IndexingContext.current().subs,
+            dynamic_symbols,
+            config,
+            use_scheduling=use_scheduling,
```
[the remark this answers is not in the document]
raikonenfnu: added a todo!

iree/turbine/kernel/wave/cache.py
```diff
@@ -0,0 +1,286 @@
...
+        module_str: str,
+    ):
+        """
+        Stores/save compiled kernels into CACHE_BASE_DIR/kernel_hash
+        including it's MLIR, VMFB, and kernel signature.
+        """
+        cur_cache_dir = f"{CACHE_BASE_DIR}/{kernel_hash}"
+        if not os.path.exists(cur_cache_dir):
+            os.makedirs(cur_cache_dir)
+        cur_vmfb_path = f"{cur_cache_dir}/{kernel_hash}.vmfb"
```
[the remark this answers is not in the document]
raikonenfnu: done!

iree/turbine/kernel/wave/cache.py
Hardcode84: The main case is when function constructed dynamically via `eval`/`exec`. It is also possible to distribute python programs as bytecode (`.pyc`) without source code.

PR #333, opened 2024-12-13
[TKW] Avoid nested `multiprocessing.Pool` usage

Hardcode84:
Processes created by `multiprocessing.Pool` have `daemon` flag which forbids to create child processes (e.g. from nested `multiprocessing.Pool`). Fallback to single-process compilation in this case. This fixes a crash when compilation itself is called from inside `multiprocessing.Pool` processes.

### DIFF
--- a/iree/turbine/kernel/wave/scheduling/graph_utils.py
+++ b/iree/turbine/kernel/wave/scheduling/graph_utils.py
@@ -15,6 +15,7 @@ import math
 from functools import partial
 from ..utils import safe_subs
 import multiprocessing as mp
+from typing import Optional
 
 T = index_symbol("$INITIATION_INTERVAL")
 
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/wave/scheduling/graph_utils.py
```diff
@@ -229,7 +230,11 @@ def all_pairs_longest_paths(
     # Parallel implementation
     for k in range(N):
         func = partial(all_pairs_longest_path_parallel, N, D, k)
-        results = pool.map(func, range(N))
+        if pool is not None:
```
harsh-nod, line 233: Can you add a comment explaining what would be the right way to do this?

harsh-nod (approved):
Thanks for fixing this! Please add a comment explaining how we would do nested multiprocessing loops so someone else can take that on.

PR #342, opened 2024-12-16
[TKW][NFC] Split gemm and igemm lit tests

Hardcode84:


harsh-nod (approved):
thanks for the cleanup!

PR #343, opened 2024-12-16
Remove 'torch' from deps and warn instead

marbre:
Instead of enforcing the installation for 'torch' as a dependency, error if 'torch' cannot be imported and point the user to how to install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ Discord server).
     pip install torch --index-url https://download.pytorch.org/whl/test/cpu
     ```
 
+    Then install iree-turbine:
+
     ```bash
     # Stable releases
     pip install iree-turbine
--- a/README.md
+++ b/README.md
@@ -32,21 +32,20 @@ Discord server).
 
 1. Install from PyPI:
 
+    Install a torch version that fulfills your needs:
+
+    ```bash
+    # Fast installation of torch with just CPU support.
+    # See other options at https://pytorch.org/get-started/locally/
+    pip install torch --index-url https://download.pytorch.org/whl/test/cpu
+    ```
+
     ```bash
     # Stable releases
     pip install iree-turbine
 
     # Nightly releases
     pip install --find-links https://iree.dev/pip-release-links.html --upgrade --pre iree-turbine
-    ```
-
-    The above does install some CUDA/cuDNN packages which are unnecessary for most
-    usage. To avoid this you can
-    [install just pytorch-cpu](https://pytorch.org/get-started/locally/) via:
-
-    ```bash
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -107,7 +107,6 @@ def initialize_options(self):
         f"numpy{get_version_spec('numpy')}",
         f"iree-base-compiler{get_version_spec('iree-base-compiler')}",
         f"iree-base-runtime{get_version_spec('iree-base-runtime')}",
-        "torch>=2.3.0",
```
ScottTodd, line 110: Let's update the docs too.

https://github.com/iree-org/iree-turbine?tab=readme-ov-file#quick-start-for-users
https://github.com/iree-org/iree-turbine?tab=readme-ov-file#install-pytorch-for-your-system

Specifically mention that installing `torch` is the user's responsibility and drop these comments:

> The above does install some CUDA/cuDNN packages which are unnecessary for most usage.

> If no explicit action is taken, the default PyTorch version will be installed.

ScottTodd: In the IREE docs we can add a Linux/macOS/Windows branch in the docs here too: https://iree.dev/guides/ml-frameworks/pytorch/#prerequisites. I'll send a PR for that.

marbre: > In the IREE docs we can add a Linux/macOS/Windows branch in the docs here too: https://iree.dev/guides/ml-frameworks/pytorch/#prerequisites. I'll send a PR for that.

Thanks! Updated the docs here, excuse me initialy missing this.

README.md
```diff
@@ -32,6 +32,14 @@ Discord server).
 
 1. Install from PyPI:
 
+    Install a torch version that fulfills your needs:
+
+    ```bash
+    # Fast installation of torch with just CPU support.
+    # See other options at https://pytorch.org/get-started/locally/
+    pip install -r pytorch-cpu-requirements.txt
+    ```
```
ScottTodd, line 41: These user instructions don't assume that you have the repository checked out yet, so I don't think we should depend on this file (and it only works on Linux right now)

setup.py
ScottTodd: Sent https://github.com/iree-org/iree/pull/19493 for https://iree.dev/guides/ml-frameworks/pytorch/#prerequisites

README.md
marbre: Good catch. Updated again.

ScottTodd: Oh, you updated the section below but not this first user section? The text down there LGTM, but this user section right at the top of the docs shouldn't assume you have the git repo cloned.

ScottTodd (approved):
One more nit

PR #345, opened 2024-12-18
Fix typo in rocdl import

GMNGeoffrey:
At least I assume this a typo. The actual imported symbol doesn't appear to ever get used, but seems better to have this right.


ScottTodd:
Yeah looks like a typo. Missed in https://github.com/iree-org/iree-turbine/pull/330, oops!

iree/turbine/kernel/compiler/ir.py
```diff
@@ -46,7 +46,7 @@
     llvm as llvm_d,
     math as math_d,
     memref as memref_d,
-    rocdl as rocld_d,
+    rocdl as rocdl_d,
```
ScottTodd, line 49: > The actual imported symbol doesn't appear to ever get used, but seems better to have this right.

BTW for this, I'm considering wrapping this import in a `try: except:` to maintain some compatibility with prior versions of the IREE packages... at least the import is used.

ScottTodd (commented):
For the DCO check, see https://iree.dev/developers/general/contributing/#developer-certificate-of-origin (that replaced the CLA check)

CI failures are strange... let me check if those are preexisting.

iree/turbine/kernel/compiler/ir.py
GMNGeoffrey: Yeah I ran into the same failure as someone was discussing on the discord, which is how I noticed this typo :-)

ScottTodd (approved):

PR #346, opened 2024-12-18
Add workflow_dispatch triggers to ci.yml and ci-tk.yml.

ScottTodd:
This will let us test the workflows without needing to push changes.


marbre (approved):
Makes sense, thanks.

PR #347, opened 2024-12-18
Switch torch requirements back to stable channel.

ScottTodd:
We have some test failures with 2.6.0 prereleases: https://github.com/iree-org/iree-turbine/actions/runs/12399117695?pr=346. Sticking to stable versions until we can triage.

### DIFF
--- a/pytorch-cpu-requirements.txt
+++ b/pytorch-cpu-requirements.txt
@@ -1,4 +1,3 @@
---pre
 --index-url https://download.pytorch.org/whl/test/cpu
 torch>=2.3.0
 torchaudio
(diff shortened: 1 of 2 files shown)


pytorch-cpu-requirements.txt
```diff
@@ -1,5 +1,4 @@
---pre
---index-url https://download.pytorch.org/whl/test/cpu
+--index-url https://download.pytorch.org/whl/cpu
```
ScottTodd, line 1: Might want to be careful with this `whl/test/cpu` index downstream too, @marbre : https://github.com/nod-ai/shark-ai/blob/c4a592ac8bcb2202a554ab1a4d311fdf5ddf28eb/.github/workflows/ci-sharktank.yml#L77

marbre (approved):

PR #350, opened 2024-12-19
[TKW] Fix chained dynamic vals

Hardcode84:
* Properly propagate index through chained read.write ops using `transform_index_backwards`
* Use proper symbols for dynamic vals
* Do not propagate 0 index

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -1012,10 +1012,14 @@ class Read(CustomOp):
             i = self.mapping_dynamic_vals.index(arg)
             iters = self.mapping.iters
             mapping = self.mapping.dynamic_val_mappings[i]
-            subs = {v: k for k, v in zip(iters, mapping.keys())}
+
+            # This logic relies on fact out mapping is identity.
+            subs = {
+                k: index[v] for k, v in zip(iters, self.mapping.output_mapping.keys())
+            }
             return {
-                k: v.apply_expr(subs[k], mapping[k])
-                for k, v in zip(arg.type.symbolic_shape, index.values())
+                k: IndexSequence.from_expr(mapping[k], subs)
+                for k in arg.type.symbolic_shape
             }
 
         return index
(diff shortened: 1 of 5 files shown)


iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1013,16 +1013,22 @@ def transform_index_backwards(
             iters = self.mapping.iters
             mapping = self.mapping.dynamic_val_mappings[i]
             subs = {v: k for k, v in zip(iters, mapping.keys())}
-            return {k: v.apply_expr(subs[k], mapping[k]) for k, v in index.items()}
+            return {
+                k: v.apply_expr(subs[k], mapping[k])
+                for k, v in zip(arg.type.symbolic_shape, index.values())
```
harsh-nod, line 1018: Why not use index.items() instead of splitting this into a type.symbolic_shape and index.values()?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1013,16 +1013,22 @@ def transform_index_backwards(
...
     def get_derived_indices(
         self,
     ) -> list[tuple[dict[IndexSymbol, IndexSequence], fx.Node]]:
         def transform_idx(arg):
-            new_index = self.transform_index_backwards(self.index, arg)
-            return {k: v for k, v in zip(arg.type.symbolic_shape, new_index.values())}
+            return {
+                k: v
+                for k, v in self.transform_index_backwards(self.index, arg).items()
+                if v.start != 0
```
harsh-nod, line 1030: Just wondering, what happens if you propagate the 0?

lit_tests/kernel/wave/codegen.py
```diff
@@ -498,7 +500,7 @@ def test(
         )
         tkw.write(res, b, elements_per_thread=16)
 
-    with codegen_test_context(canonicalize=True):
+    with codegen_test_context(canonicalize=True, additional_symbols={ONE: 1}):
```
harsh-nod, line 503: why add ONE as an additional symbol? Is it to avoid adding it to the hyperparams every time?

harsh-nod (approved):
lgtm! thanks!

PR #351, opened 2024-12-19
[TKW] Kernel cache changes

Hardcode84:
* Disable filesystem-based kernel caching by default as it too human error prone
* Do not access `cache_manager` and do not compute kernel hash if caching is disabled.
* Only run cache tests is caching is enabled


harsh-nod (approved):
lgtm!

PR #353, opened 2024-12-20
[TKW] Distribute gpu tests

Hardcode84:


iree/turbine/kernel/wave/utils.py
```diff
@@ -1052,8 +1052,18 @@ def all_equal(input_list: list[Any]) -> bool:
     return all(elem == input_list[0] for elem in input_list)
 
 
+DEFAULT_GPU_DEVICE = None
```
harsh-nod, line 1055: Rather than have this as a global variable, can we instead pass this as an argument to the get_gpu_device_name(int) function? I am trying to see if we can reduce the number of command line arguments we need to export and this will also help when we get to distributed kernels (across multiple GPUs).

harsh-nod (changes requested):

.github/workflows/ci-tk.yaml
```diff
@@ -76,7 +76,8 @@ jobs:
         run: |
           pip install --no-compile -r pytorch-rocm-requirements.txt
           export WAVE_RUN_E2E_TESTS=1
-          WAVE_CACHE_ON=0 pytest -n 4 --capture=tee-sys -vv ./tests/kernel/wave/
+          export WAVE_DISTRIBUTE_GPU_TESTS=8
+          WAVE_CACHE_ON=0 pytest -n 8 --capture=tee-sys -vv ./tests/kernel/wave/
```
Hardcode84, line 80: Switch to pytest command line param

iree/turbine/kernel/wave/utils.py
harsh-nod: After a discussion, it makes sense to keep this for now to minimize changes across tests. But we will remove the WAVE_DISTRIBUTE_GPU_TESTS parameter and instead make it a pytest argument.

harsh-nod (approved):
lgtm!

Hardcode84:
tests are still failing occasionally, btw

PR #356, opened 2024-12-24
[TKW] Add `arange` op

Hardcode84:
`arange` allows to create evenly spaced constant across a symbolic dimension.


Hardcode84:
mostly superseded by `self_index`

PR #357, opened 2024-12-26
Redo expansion

harsh-nod:
This PR is a complete overhaul of the expansion process in wave incorporating learnings from the past several months. Below, we list some of the major changes - 

1. Use an iterative rather than recursive approach which should allow for better debugging
2. Allow expansion to move through reductions and no special case expansions for reductions (previously, expansion was bfs in the reduction and dfs elsewhere)
3. Updated printing of nodes to specify dimensions
4. Split expansion into expansion and fixup phase (reductions and MMAs are adjusted in the fixup)
5. Only special handling for reshape & reduce ops that require concatenation of inputs 
6. Ordering of nodes is now 00, 01, 10, 11, unlike previously
7. Nodes are always cloned in expansion where previously the first wasn't while the rest were

and many more. While updating the lit tests, I noticed that we were doing some unnecessary reductions which are not happening anymore.


iree/turbine/kernel/wave/expansion/expansion.py
```diff
@@ -0,0 +1,698 @@
...
+        metadata.dim_query, node.indexing_dims
+    )
+
+    # Check if the node has already been expanded, if so return early.
+    key = ExpansionInfo(node, get_indexed_dims(expanded_dims, node))
+    if key in expansion_context:
+        update_users(node, expansion_context[key], metadata, expansion_context)
+        return nodes_to_expand
+
+    if metadata.do_not_expand:
```
harsh-nod, line 530: Remove

iree/turbine/kernel/wave/expansion/expansion.py
```diff
@@ -0,0 +1,698 @@
...
+        dim_scaling,
+        metadata,
+        expansion_context,
+    )
+
+    # Check for any expanded users and update their arguments.
+    update_users(node, new_node, metadata, expansion_context)
+
+    # Check if node should not be expanded.
+    if metadata.do_not_expand:
```
harsh-nod, line 555: remove

iree/turbine/kernel/wave/expansion/expansion_utils.py
```diff
@@ -0,0 +1,253 @@
...
+    stack = leaf_nodes
+    while stack:
+        node = stack.pop(0)
+        inputs, _ = get_inputs(node.fx_node, None)
+        for input in inputs:
+            stack.append(get_custom(input))
+        if not node.users:
+            if node.fx_node._erased:
+                continue
+            node.graph.erase_node(node.fx_node)
```
GMNGeoffrey, line 244: FYI, I tried patching this in and now get errors in a previously functioning kernel where a GetResult node has no value during `remove_original_nodes`. It looks like the node gets revisited even if it's already been erased, so this would happen any time it has multiple users, so I think you want to move the check if the node has already been erased to the beginning of the loop (also, nit: this is a fifo queue, not a stack, and it's got custom ops, not fx nodes)


```suggestion
def remove_original_nodes(leaf_nodes: list[CustomOp]):
    """
    Remove the original nodes from the graph.
    """
    queue = leaf_nodes
    while queue:
        custom = queue.pop(0)
        if custom.fx_node._erased:
            continue
        inputs, _ = get_inputs(custom.fx_node, None)
        for input in inputs:
            queue.append(get_custom(input))
        if not custom.users:
            custom.graph.erase_node(custom.fx_node)
```

harsh-nod: Thanks! Good catch, I will incorporate these changes.

harsh-nod:
@GMNGeoffrey - I have incorporated your changes. Any other issues with this PR or is it good to land?

GMNGeoffrey:
> @GMNGeoffrey - I have incorporated your changes. Any other issues with this PR or is it good to land?

That's the only one I hit so far, but I didn't review the code otherwise

Hardcode84 (approved):

PR #359, opened 2025-01-02
[TKW] Add support for tkw.round_even

egebeysel:
Adds support for the `round_even` op. Also added some missing `CHECK` statements for the tests.


raikonenfnu:
Sorry @egebeysel, just got back from holidays,  the PR looks overall good but do need a rebasing though, was wondering if you can help with that :)

PR #360, opened 2025-01-02
[TKW] Add layout to maximize shared reads

raikonenfnu:
To improve performance on attention, we add virtual MMA layout to maximize read/loads from shared memory to register. We also add decomposition for these virtual MMA layout that converts virtual MMA + coalesced reads -> unrolled MMA + unrolled broken apart reads.  The decomposition for VMMA is done in three steps:

      1. Collect all MMA ops and only process the ones with virtual intrinsics.
      2. For each VirtualMMAOp, we generate check it's unroll factor
      3. Generate slices of lhs and rhs from coaleseced reads based on native and virtual MMA sizes
      4. Generate native MMA ops based on unroll factor that takes in slices of coaleseced lhs and rhs
      5. Modify index expression of new mma op by replacing the expression based on virtual layout
         with it's equivalent native layout expression.

### DIFF
--- a/iree/turbine/kernel/wave/constraints.py
+++ b/iree/turbine/kernel/wave/constraints.py
@@ -46,6 +46,8 @@ class MMAType(Enum):
     # Intrinsics introduced in CDNA1
     F32_16x16x16_F16 = 0x1020
     F32_32x32x8_F16 = 0x1021
+    F32_16x16x32_K8_F16 = 0x1022
+    F32_32x32x16_K8_F16 = 0x1023
     I32_16x16x16_I8 = 0x10C0
     I32_32x32x8_I8 = 0x10C1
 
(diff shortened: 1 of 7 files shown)


iree/turbine/kernel/wave/constraints.py
```diff
@@ -132,19 +134,112 @@ def mma_matrix_shapes(self, mma_type: Optional[MMAType]) -> tuple[int]:
...
+                | MMAType.F32_32x32x16_K8_F16
                 | MMAType.F32_32x32x16_K4_F8
                 | MMAType.I32_32x32x16_I8
             ):
                 return (32, 32, 16)
             case _:
                 return ()
 
+    def mma_index_offset(self, mma_type: Optional[MMAType]):
+        lane = self.linearized_thread_id % self.threads_per_wave
```
harsh-nod, line 153: minor: use lane as input argument instead of recomputing it? Also seems like the mma_type check is also being done outside this function so not needed inside?

lit_tests/kernel/wave/gemm.py
```diff
@@ -1235,3 +1235,173 @@ def repeat(
         # CHECK: %[[BCAST_SRC:.+]] = vector.extract %[[MAX_EXT]][0] : f32 from vector<1xf32>
         # CHECK: %[[BROADCAST:.+]] = vector.splat %19 : vector<4xf32>
         # CHECK: arith.divf %[[LOOP]]#1, %[[BROADCAST]] : vector<4xf32>
+
+
+# This test that our stack is able to handle VMMA layout with maximized width read in the K-dimension.
```
harsh-nod, line 1240: I love this! Tells me exactly what I need to look for!

iree/turbine/kernel/wave/decompose_vmma_ops.py
```diff
@@ -0,0 +1,166 @@
...
+    # Get MMA nodes.
+    mma_nodes = trace.walk(lambda node: isinstance(get_custom(node), MMA))
+    if not mma_nodes:
+        return
+    hardware_constraint = next(
+        c for c in constraints if isinstance(c, HardwareConstraint)
+    )
+    for node in mma_nodes:
+        mma_op = get_custom(node)
+        mma_type = mma_op.mma_type
```
harsh-nod, line 92: how about
```
mma_type = mma_op.mma_type or hardware_constraint.mma_type
```
which gives you what you want because of python short circuiting?

iree/turbine/kernel/wave/decompose_vmma_ops.py
```diff
@@ -0,0 +1,166 @@
...
+        virtual_vector_shapes = mma_op.vector_shapes
+        native_vector_shapes = {
+            k: v
+            for k, v in zip(
+                virtual_vector_shapes,
+                hardware_constraint.mma_matrix_shapes(native_mma_type),
+            )
+        }
+
+        innermost_lhs_dim = mma_op.lhs.type.symbolic_shape[-1]
```
harsh-nod, line 108: If you are looking for the reduction dim, you can just do ```mma_op.reduction_dim ```

iree/turbine/kernel/wave/decompose_vmma_ops.py
```diff
@@ -0,0 +1,166 @@
...
+        innermost_rhs_dim = mma_op.rhs.type.symbolic_shape[-1]
+        assert (
```
harsh-nod, line 110: just wondering, when do you run into these situations where the innermost dims don't match up?

iree/turbine/kernel/wave/decompose_vmma_ops.py
```diff
@@ -0,0 +1,166 @@
...
+            # Emitting ReshapeOp for slicing
+            with mma_op.graph.inserting_before(mma_op.fx_node):
+                slice_lhs = Reshape([mma_op.lhs], virtual_vector_shapes).add_to_graph(
+                    mma_op.graph
+                )
+                slice_rhs = Reshape([mma_op.rhs], virtual_vector_shapes).add_to_graph(
+                    mma_op.graph
+                )
+
+            # Setting vector_shapes for num_paritions
```
harsh-nod, line 131: minor typo

iree/turbine/kernel/wave/decompose_vmma_ops.py
```diff
@@ -0,0 +1,166 @@
...
+def get_n_dim(mma_op):
+    n_dims = set(mma_op.rhs_type.symbolic_shape).difference(
+        mma_op.lhs_type.symbolic_shape
+    )
+    assert len(n_dims) == 1, "Expect to have single N-dim."
+    return n_dims.pop()
+
+
+def get_k_dim(mma_op):
+    k_dims = set(mma_op.lhs_type.symbolic_shape).intersection(
```
harsh-nod, line 50: Could use the mma_op's reduction dim here ?

harsh-nod (changes requested):
Overall looks good and clean implementation! Thanks! Just some minor comments but should be good to land.

iree/turbine/kernel/wave/constraints.py
raikonenfnu: I am using it from within this because we are using this functionality from the decompose_vmma_ops pass as well to be able to search and replace the Index Expression of the virtual mma with the origina native one during decomposition

iree/turbine/kernel/wave/decompose_vmma_ops.py
raikonenfnu: make sense, done, thanks!

iree/turbine/kernel/wave/decompose_vmma_ops.py
raikonenfnu: same reply as above, but for now replaced this with mma_op.reduciton_dim so should be the same. :)
> Done! but the main reason I was initially unsure on whether to do this or not is this is only valid because tkw.mma is always mmt_transpose_b. But if in the case it's not we may need to bring these checks back.

iree/turbine/kernel/wave/decompose_vmma_ops.py
raikonenfnu: Good catch, done!

iree/turbine/kernel/wave/decompose_vmma_ops.py
raikonenfnu: makes sense, done, thanks!

iree/turbine/kernel/wave/constraints.py
harsh-nod: makes sense, thanks

iree/turbine/kernel/wave/decompose_vmma_ops.py
harsh-nod: Good point. You could add a comment to that effect here if you would like saying that if we ever consider changing the semantics of the mma op this would need to be revisited.

harsh-nod (approved):
lgtm!

PR #372, opened 2025-01-09
Cleanup requirements and packaging files.

ScottTodd:
This is preparation for adding automation to handle updating dependency versions.

* Renamed `iree-requirements.txt` and `iree-requirements-ci.txt` to `requirements-iree-[un]pinned.txt` so they sort next to other `requirements-*.txt` files and match the names used downstream in https://github.com/nod-ai/shark-ai.
* Deleted `MANIFEST.in` that was unused since `package_data` in `setup.py` does what those `include` lines used to: https://github.com/iree-org/iree-turbine/blob/e4550f37dcd8b9b691db93c30b478c1d67eee83b/setup.py#L97-L100
* Sorted dependencies in `requirements.txt`.
* Deleted unused `mypy-requirements.txt`.

### DIFF
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -55,7 +55,7 @@ jobs:
           # from non default locations first. Installing the PyTorch CPU
           # wheels saves multiple minutes and a lot of bandwidth on runner setup.
           pip install --no-compile -r pytorch-cpu-requirements.txt
-          pip install --no-cache-dir -r iree-requirements-ci.txt --upgrade
+          pip install --no-cache-dir -r requirements-iree-pinned.txt --upgrade
           pip install -r requirements.txt -e .
 
       - name: Run unit tests
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -46,7 +46,7 @@ jobs:
         id: cache-pip
         with:
           path: ${{ env.PIP_CACHE_DIR }}
-          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements.txt') }}
+          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements*.txt') }}
 
       - name: Install pip deps
         run: |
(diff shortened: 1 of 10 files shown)


build_tools/build_release.py
```diff
@@ -117,7 +117,7 @@ def _download_iree_binaries_for_platform_args(
                 # across projects and new stable versions of the IREE packages
                 # haven't yet been pushed.
                 "-r",
-                REPO_ROOT / "iree-requirements-ci.txt",
+                REPO_ROOT / "requirements-iree-pinned.txt",
```
marbre, line 120: Without looking to deep into the script, I am wondering why it is needed to install those requirements when building a release. Anyway the change is fine.

marbre (approved):

build_tools/build_release.py
ScottTodd: The downloading is mostly leftover from a previous iteration of the release process where IREE wheels would be pushed _from iree-turbine_:

1. From iree-turbine, run this script
2. The script downloads candidate wheels from IREE, builds iree-turbine, and runs iree-turbine tests
3. Push all IREE and iree-turbine wheels to PyPI _from iree-turbine_

Lately I've been doing this instead:

1. Choose a release candidate in IREE
2. Switch to iree-turbine then set `iree-requirements-ci.txt` / `requirements-iree-pinned.txt` to that release candidate and send a PR that runs CI tests then merge it
3. Switch back to IREE and publish the IREE wheels
4. Switch back to iree-turbine, build wheels, and publish

.github/workflows/ci-tk.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
         id: cache-pip
         with:
           path: ${{ env.PIP_CACHE_DIR }}
-          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements.txt') }}
+          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements*.txt') }}
```
ScottTodd, line 49: These `hashFiles` sections are so error prone while we have many different requirement files x_x

build_tools/build_release.py
marbre: Thanks for enlightening me :)

.github/workflows/ci-tk.yaml
marbre: Oh yeah, I could have fetched that but did not :/

PR #377, opened 2025-01-10
[TKW] CI: Install GPU deps in separate step

Hardcode84:


.github/workflows/ci-tk.yaml
```diff
@@ -63,24 +63,26 @@ jobs:
         run: |
           pytest -n 4 --capture=tee-sys -vv ./tests/kernel/wave/
 
-      - name: Test TKW runtime related stack on amdgpu
+      - name: Install GPU deps
         if: "contains(matrix.os, 'amdgpu') && !cancelled()"
         run: |
           pip install --no-compile -r pytorch-rocm-requirements.txt
```
ScottTodd, line 69: We should probably be working within a venv, especially on persistent self-hosted runners 🤔

ScottTodd (approved):

.github/workflows/ci-tk.yaml
Hardcode84: I thought we already do, but yeah, I can  look into it

ScottTodd: Sample PR: https://github.com/nod-ai/shark-ai/pull/640

Hardcode84: BTW, as an potential alternative I was using `setup-miniconda` action in the past https://github.com/numba/numba-mlir/blob/main/.github/workflows/build.yml#L52 and was quite happy with it

PR #368, opened 2025-01-07
Add reference for paged decode attention

harsh-nod:
This PR adds a reference kernel for paged decode
attention that can be used for validation.

PR #378, opened 2025-01-10
[TKW] Properly handle dynamic vals in minimize_global_loads

Hardcode84:
Extracted from https://github.com/iree-org/iree-turbine/pull/368

### DIFF
--- a/iree/turbine/kernel/wave/minimize_global_loads.py
+++ b/iree/turbine/kernel/wave/minimize_global_loads.py
@@ -89,6 +89,14 @@ def identify_optimizable_loads(
         custom = get_custom(read_node)
         if custom.memory in processed_memories:
             continue
+
+        # Bail out if we have dynamic values dependensies with elements_per_thread != 1
+        if any(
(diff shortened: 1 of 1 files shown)


harsh-nod (approved):

iree/turbine/kernel/wave/minimize_global_loads.py
```diff
@@ -89,6 +89,14 @@ def identify_optimizable_loads(
         custom = get_custom(read_node)
         if custom.memory in processed_memories:
             continue
+
+        # Bail out if we have dynamic values dependensies with elements_per_thread != 1
+        if any(
+            (getattr(get_custom(v), "elements_per_thread", None) != 1)
+            for v in custom.mapping_dynamic_vals
+        ):
```
harsh-nod, line 97: Actually, I think we should skip this pass if we have dynamic vals regardless of the elements per thread because that violates the assumptions of the pass.

Hardcode84: https://github.com/iree-org/iree-turbine/pull/383

PR #382, opened 2025-01-11
[TKW] `set_symbol` and `apply_expr` ops

Hardcode84:
* `set_symbol` op to set specific symbol to some dynamic value (i.e. read from memory) during codegen, this is mostly a hack until we have a proper support for slices/reductions with dynamic sizes/tilesizes.
* `apply_expr` op to allow nontrivial index calculations.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -1395,12 +1395,8 @@ class SetSymbol(CustomOp):
         return get_custom(self.register_).type
 
     @property
-    def index(self) -> Optional[dict[IndexSymbol, IndexSequence]]:
-        return get_custom(self.register_).index
-
-    @index.setter
-    def index(self, value: dict[IndexSymbol, IndexSequence]):
-        pass  # noop
+    def indexing_dims(self) -> list[IndexSymbol]:
+        return get_custom(self.register_).indexing_dims
 
 
 @define_py_op(operator.getitem)
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -1390,7 +1390,7 @@ class Write(CustomOp):
 
 @define_op("apply_expr")
 @dataclass
-class AppplyExpr(CustomOp):
+class ApplyExpr(CustomOp):
     register_: fx.Proxy
     expr: Callable
 
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -98,7 +98,7 @@ def write(
     ...
 
 
-def apply_expr(value: "Register", expr: IndexExpr) -> "Register":
+def apply_expr(value: "Register", expr: Callable) -> "Register":
     ...
 
 
@@ -1392,7 +1392,7 @@ class Write(CustomOp):
 @dataclass
 class AppplyExpr(CustomOp):
     register_: fx.Proxy
-    expr: IndexExpr
+    expr: Callable
 
     @property
     def type(self) -> "Register":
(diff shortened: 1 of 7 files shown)


harsh-nod:
> @harsh-nod codegen still fails as I'm not sure how to construct dispatch op

I think you can leave the dispatch codegen as is for now and just pass in a dummy value since it will not be used in the kernel.

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -517,6 +517,181 @@ def test(
...
+
+    i = tkw.IndexMapping.iterator(0)
+    j = tkw.IndexMapping.iterator(1)
+    mapping = tkw.IndexMapping(
+        num_iterators=2,
+        inputs={S: S, N: j},
+        outputs={S: i, N: j},
+    )
+
+    A = tkw.APPLY_EXPR_ARG
```
harsh-nod, line 648: why do we need A? Can't we use a lambda?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1380,6 +1388,36 @@ def is_contiguous_vec(self) -> bool:
         )
 
 
+@define_op("apply_expr")
+@dataclass
+class AppplyExpr(CustomOp):
```
harsh-nod, line 1393: typo: ApplyExpr

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1380,6 +1388,36 @@ def is_contiguous_vec(self) -> bool:
...
+    @property
+    def indexing_dims(self) -> list[IndexSymbol]:
+        return get_custom(self.register_).indexing_dims
+
+
+@define_op("set_symbol")
+@dataclass
+class SetSymbol(CustomOp):
+    symbol: IndexExpr
+    register_: fx.Proxy
```
harsh-nod, line 1410: Just wondering why mark these as type fx.Proxy instead of fx.Node?

harsh-nod (approved):
looks good! Thanks!

iree/turbine/kernel/ops/wave_ops.py
Hardcode84: Half of the ops are using `fx.Proxy` intead `fx.Node`, we should probably align all of them.

iree/turbine/kernel/wave/codegen.py
```diff
@@ -907,6 +909,40 @@ def handle_write(emitter: WaveEmitter, node: fx.Node):
...
 
+@handle_op(apply_expr)
+def handle_apply_expr(emitter: WaveEmitter, node: fx.Node):
+    try:
+        register, expr = node.args
+    except ValueError as e:
+        raise ValidationError("Malformed arguments") from e
+
+    APPLY_EXPR_ARG = index_symbol("$APPLY_EXPR_ARG")
+    expr = expr(APPLY_EXPR_ARG)
```
nicolasvasilache, line 920: @Hardcode84 could you explain the rationale here?
I was hoping this could compose expressions in a quite general manner but I must be using it improperly as the eval always returns False in my use cases.

lit_tests/kernel/wave/codegen.py
```diff
@@ -593,6 +593,133 @@ def test(
...
+        tkw.set_symbol(S, offset)
+        res = tkw.read(
+            a,
+            mapping=mapping,
+            elements_per_thread=1,
+        )
+        tkw.write(res, b, elements_per_thread=1)
+
+    with codegen_test_context(
+        canonicalize=True,
```
nicolasvasilache, line 699: It is generally a good idea to not canonicalize in lit tests.

In this case I think what is happening is that M is distributed to 1 by the transformation and then M-1 is canonicalized/folded to give `%[[C0:.*]] = arith.constant 0 : index`

Issue #374, opened 2025-01-10
[TKW] Bug: gemm write inside of unrelated reduction writes incorrect value


GMNGeoffrey:
Putting a gemm inside of a reduction that uses another variable causes writes of its output to be incorrect if it expands to multiple mfma operations. Specifically, the result of the first mfma is written rather than the result of the final accumulation. Note that there is no actual reduction happening here (no for loop) as the block size is the same as the dimension size, but the reduction is necessary for the repro

https://github.com/gmngeoffrey/iree-turbine/commit/94451bc4f46daa2406f8653fd02e7cc1903803c7 reproduces this using the test case copied below, which was reduced from the flash attention forward pass (with writing of intermediate values for debugging). Note that further operations on the gemm output correctly use the fully accumulated value and so the final result is correct. It's easiest to see in the output values when the two inputs values are constant because then the output is just half what it's supposed to be, but it also happens with random inputs. Switching the K1 dimension to half the mfma size instead, runs into a different problem that doubles all the outputs (including for the max). I think half the threads should be doing no work in that case, but they're not.

FYI @harsh-nod

<details>

```python
def testReproWriteInLoopBug():
    mfma_variant = MMAType.F32_16x16x16_F16
    shape = (16, 32, 16)
    q_seq_len, qk_head_dim, kv_seq_len = shape
    M = tkl.sym.M
    K1 = tkl.sym.K1
    K2 = tkl.sym.K2

    BLOCK_K2 = tkl.sym.BLOCK_K2
    ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
    LOAD_ELEMS_PER_THREAD = tkl.sym.LOAD_ELEMS_PER_THREAD
    STORE_ELEMS_PER_THREAD = tkl.sym.STORE_ELEMS_PER_THREAD

    # Expose user-constraints
    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, M, 0)]
    constraints += [tkw.TilingConstraint(K2, BLOCK_K2)]

    constraints += [
        tkw.HardwareConstraint(
            threads_per_wave=64,
            waves_per_block=(1, 1, 1),
            mma_type=mfma_variant,
        )
    ]

    @tkw.wave(constraints)
    def attention_fwd(
        q: tkl.Memory[M, K1, GLOBAL_ADDRESS_SPACE, tkl.f16],
[...]

PR #395, opened 2025-01-18
Expand from write nodes inside reductions

GMNGeoffrey:
Right now, it only looks at write nodes in the root graph and misses ones in reductions. Instead we walk all the regions for `Write` nodes and `GetResult` nodes without users.

`get_leaf_node_types` also appeared to always return `[Write]` as it sets `leaf_nodes` to that and then immediately checks if it's empty, which it never will be, so I deleted it.

Also includes some .gitignore additions that I can drop if you don't like them.

Fixes https://github.com/iree-org/iree-turbine/issues/374

### DIFF
--- a/iree/turbine/kernel/wave/expansion/expansion.py
+++ b/iree/turbine/kernel/wave/expansion/expansion.py
@@ -42,8 +42,11 @@ from ..utils import (
     get_users,
     get_inputs,
 )
+from ....support.logging import get_logger
 from copy import deepcopy
 import math
+
+logger = get_logger("turbine.wave.expansion")
 
 
 @dataclass(frozen=True)
@@ -88,25 +91,6 @@ class ExpansionContext:
 
     def __setitem__(self, key: ExpansionInfo, value: CustomOp):
         self.expansion_context[key] = value
-
-
-def get_leaf_node_types(
-    trace: CapturedTrace, all_nodes_reversed: list[fx.Node]
-) -> list[Type[CustomOp]]:
-    """
-    Get the types of leaf nodes of the trace. These are of type write.
-    If there are no write nodes, then the type of the last node in the trace is
-    used.
-    """
-    leaf_nodes: list[Type[CustomOp]] = [Write]
-    if not leaf_nodes:
-        for node in (get_custom(node) for node in all_nodes_reversed):
-            if isinstance(node, Output):
-                continue
-            leaf_nodes.append(node.__class__)
-            break
-
-    return leaf_nodes
 
 
 def get_dim_combinations(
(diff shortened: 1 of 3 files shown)


GMNGeoffrey:
I actually signed the commits with my ssh key, but it looks like it's not accepting that.

GMNGeoffrey:
@harsh-nod PTAL

ScottTodd:
> I actually signed the commits with my ssh key, but it looks like it's not accepting that.

DCO allows signing (instead of sign-off) for organization members. I've sent you an invite to iree-org + the iree-triage team. That will also let checks run without approval.

IREE docs: https://iree.dev/developers/general/contributing/#developer-certificate-of-origin
DCO docs: https://github.com/dcoapp/app?tab=readme-ov-file#skipping-sign-off-for-organization-members

iree/turbine/kernel/wave/expansion/expansion.py
```diff
@@ -650,6 +634,13 @@ def fixup_reduction_nodes(trace: CapturedTrace, expansion_context: ExpansionCont
         remove_original_nodes(return_vals)
 
 
+def is_leaf_node(node):
```
harsh-nod, line 637: nice!

iree/turbine/kernel/wave/expansion/expansion.py
```diff
@@ -663,17 +654,15 @@ def expand_graph(
 
     add_get_results(trace)
 
-    all_nodes_reversed = list(reversed(trace.get_root_graph().nodes))
-    leaf_node_types = get_leaf_node_types(trace, all_nodes_reversed)
-    leaf_nodes = []
+    leaf_ops = [get_custom(node) for node in reversed(trace.walk(is_leaf_node))]
+    if not leaf_ops:
+        final_op = get_custom(next(reversed(trace.get_root_graph().nodes)))
```
harsh-nod, line 659: Just wondering could you get the last op by doing ```trace.get_root_graph()._root.prev``` ?

harsh-nod (changes requested):
thanks, this looks good! Could you add a lit test with a write inside a reduction in expansion to make sure that this works properly?

iree/turbine/kernel/wave/expansion/expansion.py
GMNGeoffrey: Done. I'm not thrilled about using the private member here, but what I was doing wasn't working anyways (added a test for that also)

GMNGeoffrey:
Added tests. PTAL

GMNGeoffrey:
The failing test doesn't seem related to my change and passes on my local mi210. It's also a numerical diff on a check that already has rtol set to 0.8, which suggests to me that it's already broken and should probably be xfailed or skipped...

harsh-nod:
> The failing test doesn't seem related to my change and passes on my local mi210. It's also a numerical diff on a check that already has rtol set to 0.8, which suggests to me that it's already broken and should probably be xfailed or skipped...

Yes we have flaky attention tests that need to be debugged, but you can just retrigger for now.

harsh-nod (approved):
lgtm! thanks!

PR #396, opened 2025-01-19
Migrate ci-tk workflow to run on Conductor for MI300

yamiyysu:
This PR updated ci-tk's MI300 workflow to switch to using Conductor by updating the run label to linux-mi300-gpu-1. The runner had already been created offline.


Hardcode84:
FYI, we are using runner name to detect if it's a gpu runner: https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/ci-tk.yaml#L52 so you will have to update these as well

PR #397, opened 2025-01-20
Rework dependency updates to avoid force pushing.

ScottTodd:
This changes how we use https://github.com/peter-evans/create-pull-request in the recently added [`update_iree_requirement_pins.yml`](https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/update_iree_requirement_pins.yml) action to avoid force pushing when changes are detected.

Now, the action will only check for changes and create a PR if the branch does not already exist. I read through the action's documentation (https://github.com/peter-evans/create-pull-request/blob/main/docs/common-issues.md#disable-force-updates-to-existing-pr-branches), issues, and sample usage across GitHub before settling on this implementation strategy. I figure that our dependency updates are either

* (A) uneventful and can be merged quickly 
* (B) tricky and will require local patches, rounds of CI runs, and possibly multiple days of debugging

For (B), I don't want the automation to force push over the existing branch and I also don't want a new pull request to be created that is missing all of that debugging state. We could have the automation continue to push update commits (without force push) to the branch, but that might further destabilize integration work. Some middle ground might be to only push update commits if there has been no activity on the PR yet (like labels added, commits pushed, or a comment made). At the point where we have complicated rules controlling automated behavior, we might as well take over the process manually ¯\\\_(ツ)_/¯

## Other misc changes

* Allow the branch name to be overridden when invoked via workflow_dispatch.
* Skip `actions/create-github-app-token` when run in repositories without the matching secrets.
  * This will still create PRs, but those PRs won't run workflow checks automatically.
* Set the commit author to `iree-pr-automator[bot] <iree-github-actions-bot@google.com>` rather than the GitHub actor (which will be me for scheduled runs until someone else edits the workflow, lol).
  * That email is used for gh-pages branches updates in the iree-org/iree repository.
  * We could also set the author to `github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>` or another bot/email.

## Testing

Tested on my fork using workflow_dispatch:

1. Run with no existing branch: https://github.com/ScottTodd/iree-turbine/actions/runs/12876405273 created https://github.com/ScottTodd/iree-turbine/pull/15
2. Run with an existing branch: https://github.com/ScottTodd/iree-turbine/actions/runs/12876416861 skipped update checks and PR creation


.github/workflows/update_iree_requirement_pins.yml
```diff
@@ -16,47 +21,73 @@ permissions:
...
+        id: cpr
         uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f # v7.0.6
         with:
-          token: ${{ steps.generate-token.outputs.token }}
+          token: ${{ steps.generate-token.outputs.token || secrets.GITHUB_TOKEN }}
           base: main
-          branch: integrates/iree
-          delete-branch: true
+          branch: ${{ env.GIT_BRANCH_NAME }}
+          author: iree-pr-automator[bot] <iree-github-actions-bot@google.com>
```
marbre, line 81: `iree-github-actions-bot@google.com`? `@google.com`? :thinking:

marbre: I was just wondering. Go with what you think makes most sense here.

marbre (approved):

Issue #375, opened 2025-01-10
[TKW] Bug: multiple issues with input/output determination


GMNGeoffrey:
The current logic for inputs and outputs has a number of issues:

https://github.com/iree-org/iree-turbine/blob/e4550f37dcd8b9b691db93c30b478c1d67eee83b/iree/turbine/kernel/compiler/kernel_codegen.py#L250-L289

(amongst other places)

1. If something has any dependency *other* than a write, it gets labeled an input, which then marks it as readonly and it comes out unmodified.
2. Writes inside of reductions are not considered for classifying an output.
3. Everything designated an output gets moved to the end of the signature, so when the compiled kernel is called the arguments are all scrambled.
4. Arguments that are unused somehow get stripped out during compilation and then when the compiled kernel is called the arguments are all scrambled.

So far, I've found that the easiest thing was to just mark everything as an output. Not sure what other implications that has besides passing up optimizations dependent on things being read-only. I think this also still doesn't fix problem 4.

nithinsubbiah:
closing as stale, please reopen issue at https://github.com/iree-org/wave if this isn't solved yet

PR #400, opened 2025-01-21
[TKW] `tkw.conditional` op

Hardcode84:
Simple conditional op which is lowered to `scf.if` (without else for now). Expected argument of integer type and treats it as:
```
scf.if(%val != 0) {
  ...
}
```

### DIFF
--- a/iree/turbine/kernel/compiler/kernel_codegen.py
+++ b/iree/turbine/kernel/compiler/kernel_codegen.py
@@ -34,7 +34,7 @@ from ..lang.kernel_buffer import (
 )
 from ..lang.wave_types import Memory
 from ..lang.grid import Grid
-from ..ops.wave_ops import get_custom, Placeholder, Read, Write
+from ..ops.wave_ops import get_custom, Placeholder, NestedRegionOp, Read, Write
 
 from .base import (
     CodegenError,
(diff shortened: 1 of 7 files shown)


iree/turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -251,13 +251,34 @@ def determine_input_output_buffers(self, graph: fx.Graph):
         # Extract all placeholder nodes.
         placeholder_nodes = filter_fx_graph(graph, is_placeholder)
 
+        def get_users_recursive(node):
```
harsh-nod, line 254: @GMNGeoffrey - this could explain why you had to mess with the input/output bindings. (We didn't look recursively to identify output nodes)

harsh-nod (approved):
looks good! I think we will need some additional changes to integrate this with paged attention but this is a good start! thanks!

iree/turbine/kernel/compiler/kernel_codegen.py
GMNGeoffrey: Maybe. I think the logic is still wrong though as noted in https://github.com/iree-org/iree-turbine/issues/375. If a node has both read and write dependencies it can't be marked as read-only, which is still what happens here IIUC.

PR #402, opened 2025-01-22
Remove windows newlines

GMNGeoffrey:
These appear to have made their way into the repository and were causing weird diffs when I edited things.


GMNGeoffrey:
`git log -S '\r'` was not able to tell me when these got introduced but based on just the commits that created those files, I'd guess that you might want to look at your git `autocrlf` or editor settings @Hardcode84

ScottTodd:
Looks like pre-commit has a hook that could help guard against this in the future:
https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#mixed-line-ending

ScottTodd:
Will let @Hardcode84 weigh in and then merge.

Hardcode84 (approved):
Probably from my unconventional sublime+wsl setup ), thanks

PR #407, opened 2025-01-23
Update grid size for PDA

harsh-nod:
This PR updates the grid distribution for paged decode attention. Previously, we were launching only as many workgroups as there were kv heads. However, if instead we can exploit much more parallelism if instead we launch ```shape.num_query_heads // min(HEAD_BLOCK_SIZE, head_ratio)``` workgroups and do some arithmetic on the workgroup index to recover the kv head index.

### DIFF
--- a/iree/turbine/kernel/wave/codegen.py
+++ b/iree/turbine/kernel/wave/codegen.py
@@ -396,6 +396,23 @@ def gen_sympy_index(dynamics: dict[IndexSymbol, Value], expr: sympy.Expr) -> OpR
         if isinstance(val, _Rational):
             raise CodegenError(f"Rational is not supported yet in '{type(term)}'")
 
+    def _remove_denominators(lhs, rhs):
+        """
+        Converts     (z/x) < y -> z < x*y
+              or     z < (y/x) -> z*x < y
+              or (a/b) < (c/d) -> a*d < b*c
+        """
+        if isinstance(lhs, _Rational) and not isinstance(rhs, _Rational):
+            rhs = _mul(lhs.denominator, rhs)
+            lhs = lhs.numerator
+        if isinstance(rhs, _Rational) and not isinstance(lhs, _Rational):
+            lhs = _mul(rhs.denominator, lhs)
+            rhs = rhs.numerator
+        if isinstance(lhs, _Rational) and isinstance(rhs, _Rational):
+            rhs = _mul(lhs.denominator, rhs.numerator)
+            lhs = _mul(rhs.denominator, lhs.numerator)
+        return lhs, rhs
+
     def _get_const(val):
         if isinstance(val, int):
             return arith_d.constant(IndexType.get(), val)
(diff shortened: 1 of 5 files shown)


iree/turbine/kernel/wave/templates/paged_decode_attention.py
```diff
@@ -88,20 +91,19 @@ def phase_0_constraints():
...
 
         # BH is the kv-head index and is distributed across workgroups.
         # B is the query index and is distributed like BH but with a different
         # workgroup and wave tile size.
-        # TODO: We will want to add a function to the workgroup constraint to
-        # allow for using WG / ceil(kv_group_num, BLOCK_B) instead of just WG.
-        # This can be done by adding an optional additional argument to the WorkgroupConstraint.
 
-        constraints += [tkw.WorkgroupConstraint(BH, BLOCK_BH, 1)]
+        wg_func_2 = lambda wg: wg // math.ceil(head_ratio / HEAD_BLOCK_SIZE)
```
raikonenfnu, line 102: Can we get a little bit of doc on what is going on here? :)

raikonenfnu (approved):
Some NITs/questions, but overall code looks good so not going to block landing if needed

iree/turbine/kernel/wave/codegen.py
```diff
@@ -395,6 +395,23 @@ def _enforce_non_rational(val, term):
...
+        """
+        if isinstance(lhs, _Rational) and not isinstance(rhs, _Rational):
+            rhs = _mul(lhs.denominator, rhs)
+            lhs = lhs.numerator
+        if isinstance(rhs, _Rational) and not isinstance(lhs, _Rational):
+            lhs = _mul(rhs.denominator, lhs)
+            rhs = rhs.numerator
+        if isinstance(lhs, _Rational) and isinstance(rhs, _Rational):
+            rhs = _mul(lhs.denominator, rhs.numerator)
+            lhs = _mul(rhs.denominator, lhs.numerator)
```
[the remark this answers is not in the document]
harsh-nod: This doesn't work right? Because if rhs is rational for example, then we need to multiply the rhs numerator with lhs denominator not rhs with lhs denominator?

iree/turbine/kernel/wave/templates/paged_decode_attention.py
harsh-nod: sure, but short summary is that this is GQA.

iree/turbine/kernel/wave/codegen.py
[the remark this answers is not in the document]
raikonenfnu: if rhs is rational, we should indeed multiply the rhs numerator since it is a `if-if` not an `if-elif` right?

say we have `a/b > c/d`, after the first if statement, we'll get:
`a > (c*b) / d`

Then after second if statement, we'll get:
`a * d > c * b`

Which is what we want no?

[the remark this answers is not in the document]
harsh-nod: makes sense will change thanks

iree/turbine/kernel/wave/templates/paged_decode_attention.py
raikonenfnu: make sense! :)

iree/turbine/kernel/wave/codegen.py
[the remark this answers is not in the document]
harsh-nod: actually lets do this in a separate PR and update _Rational to a dataclass or maybe even use some sympy datatypes.

iree/turbine/kernel/wave/templates/paged_decode_attention.py
harsh-nod: added comment

raikonenfnu (approved):
thanks for the comments/docs!

tests/kernel/wave/attention/paged_attention_test.py
```diff
@@ -251,7 +251,7 @@ def testPagedFlashDecoding(
         num_kv_splits, shape.num_seqs, shape.num_query_heads, dtype=torch.float32
     )
     output = device_zeros(
-        shape.num_seqs, shape.num_query_heads, shape.head_size_kv, dtype=torch.float32
+        shape.num_seqs, shape.num_query_heads, shape.head_size_kv, dtype=torch.float16
```
Hardcode84, line 254: Was this intended?

GMNGeoffrey: Looks like it conflicted with https://github.com/iree-org/iree-turbine/pull/421 and now tests are failing: https://github.com/iree-org/iree-turbine/actions/runs/13004914350/job/36269993961

PR #409, opened 2025-01-23
Revert "Replace deprecated `assert_allclose` (#405)"

harsh-nod:
This reverts commit e226606631c442aa9170309722d19b170ecbb494.

### DIFF
--- a/tests/kernel/wave/attention/decode_attention_test.py
+++ b/tests/kernel/wave/attention/decode_attention_test.py
@@ -20,7 +20,7 @@ from iree.turbine.kernel.wave.templates.decode_attention import (
     get_decode_attention_kernels,
 )
 import os
-from torch.testing import assert_close
+from torch.testing import assert_allclose
 from ..common.utils import (
     require_e2e,
     enable_scheduling_barriers,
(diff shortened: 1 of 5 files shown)


harsh-nod:
A simple fix to the dtypes is not enough, I think we need to also determine the tolerances when switching to the non deprecated API.

harsh-nod:
I think this PR was landed prematurely. Let's take another crack at it with dtype and tolerance fixes and keep the builders green on main for now.

ScottTodd:
Ah, new failures after the change:
```
FAILED tests/kernel/wave/runtime/cache_test.py::testSameConfig - AssertionError: The values for attribute 'dtype' do not match: torch.float32 != torch.float16.
FAILED tests/kernel/wave/runtime/cache_test.py::testDifferentDynamicSameBlock - AssertionError: The values for attribute 'dtype' do not match: torch.float32 != torch.float16.
FAILED tests/kernel/wave/runtime/cache_test.py::testSameSizeDifferentBlock - AssertionError: The values for attribute 'dtype' do not match: torch.float32 != torch.float16.
```
https://github.com/iree-org/iree-turbine/actions/runs/12915847540/job/36018656310#step:8:330

ScottTodd (approved):

GMNGeoffrey:
Wait why did this land with failing tests? Everything passed for me locally, but I assumed the CI would check it on other platforms

GMNGeoffrey (approved):

GMNGeoffrey:
Ah and the reason it passed for me locally is that I have the cache turned off (I've run into a couple bugs with it)

PR #412, opened 2025-01-24
Prefill Attention

harsh-nod:
This PR adds an implementation of prefill attention. In prefill attention, we have variable length sequences and so we need to pass in the start offsets as well as the lengths of the sequences. The setting of the offsets and variable sequence lengths is accomplished using set_symbol.

Also, this PR adds the ability to specify the total count for a work group constraint and adds attention cleanups.

### DIFF
--- a/iree/turbine/kernel/wave/constraints.py
+++ b/iree/turbine/kernel/wave/constraints.py
@@ -376,7 +376,7 @@ class WorkgroupConstraint(Constraint):
     workgroup_dim: int
     apply_fn: Optional[Callable] = None
     primary: Optional[bool] = True
-    iters: Optional[IndexExpr] = None
+    iters: Optional[IndexExpr | int] = None
 
     def __post_init__(self):
         self.wg_dim = None
--- a/iree/turbine/kernel/wave/templates/attention_common.py
+++ b/iree/turbine/kernel/wave/templates/attention_common.py
@@ -1,0 +1,45 @@
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from dataclasses import dataclass
+from typing import Optional
+
+import iree.turbine.kernel.lang as tkl
+
+
+@dataclass
+class AttentionShape:
+    num_query_heads: int
+    num_kv_heads: int
+    head_size: int
+    head_size_kv: int
+    # -----------------------
+    # Prefill specific
+    num_seqs: Optional[int] = None
+    max_seq_len: Optional[int] = None
+    total_seq_len: Optional[int] = None
+    # -----------------------
+    # Vanilla attention
+    query_seq_len: Optional[int] = None
+    kv_seq_len: Optional[int] = None
+
+
+# Commonly-used attention symbols
+H = tkl.sym.H
+H_Q = tkl.sym.H
+H_KV = tkl.sym.H
+N_Q = tkl.sym.N_D
+N_KV = tkl.sym.N_KV
+D_Q = tkl.sym.D_Q
+D_KV = tkl.sym.D_KV
+
+BLOCK_H = tkl.sym.BLOCK_H
+BLOCK_H_Q = tkl.sym.BLOCK_H
+BLOCK_H_KV = tkl.sym.BLOCK_H
+BLOCK_N_Q = tkl.sym.BLOCK_N_D
+BLOCK_N_KV = tkl.sym.BLOCK_N_KV
+BLOCK_D_Q = tkl.sym.BLOCK_D_Q
+BLOCK_D_KV = tkl.sym.BLOCK_D_KV
--- a/iree/turbine/kernel/wave/templates/attention_common.py
+++ b/iree/turbine/kernel/wave/templates/attention_common.py
@@ -27,15 +27,16 @@ class AttentionShape:
     kv_seq_len: Optional[int] = None
 
 
-# Commonly-used attention symbols
-H = tkl.sym.H
-H_Q = tkl.sym.H
-H_KV = tkl.sym.H
-N_Q = tkl.sym.N_D
-N_KV = tkl.sym.N_KV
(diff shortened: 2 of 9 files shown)


iree/turbine/kernel/wave/constraints.py
```diff
@@ -392,6 +393,8 @@ def count(self) -> IndexExpr:
         """
         Returns an expression for the total number of workgroups for the specific workgroup_dim.
         """
+        if self.iters:
```
raikonenfnu, line 396: Can you explain a little when we would want something other than `ceiling(self.dim / self.tile_size)`? I thought we'd always want `grid_size/block_size`? 

Unless we are somehow splitting same data dim across mutliple thread grid dim? 🤔

iree/turbine/kernel/wave/constraints.py
```diff
@@ -376,6 +376,7 @@ class WorkgroupConstraint(Constraint):
     workgroup_dim: int
     apply_fn: Optional[Callable] = None
     primary: Optional[bool] = True
+    iters: Optional[IndexExpr | int] = None
```
nicolasvasilache, line 379: Could you plz explain the intention here?

My, yet uninformed, reading is that you want to override how many WG you will have along that dimension but I am unclear how this resolves with the potentially conflicting `dim/tile_size` expectation.

I see below you happen to set values that happen to agree with `dim/tile_size`, but I would love to understand the intent here (if there is something deeper that will go on).

i.e.
```
tkw.set_symbol(N_Q, seq_len)
+
  hyperparams = {
        ...
        BLOCK_N_Q: SEQ_TILE_SIZE,
```
happen to agree with:
```
tkw.WorkgroupConstraint(
  N_Q, BLOCK_N_Q, 0, iters=math.ceil(shape.max_seq_len / SEQ_TILE_SIZE)
)
```

but I assume that inserting a new `iters` arg is to support cases where they don't agree?

iree/turbine/kernel/wave/constraints.py
harsh-nod: Good question. In this case, the sequence length is dynamic and so in order to determine how many workgroups to launch, we use the maximum sequence length.

iree/turbine/kernel/wave/constraints.py
harsh-nod: Good question. Yes, in this case the sequence length is dynamic but we need to know the grid size at kernel launch. So in order to do that, we determine the number of workgroups using the maximum sequence length.

iree/turbine/kernel/wave/templates/attention_common.py
```diff
@@ -0,0 +1,45 @@
...
+    total_seq_len: Optional[int] = None
+    # -----------------------
+    # Vanilla attention
+    query_seq_len: Optional[int] = None
+    kv_seq_len: Optional[int] = None
+
+
+# Commonly-used attention symbols
+H = tkl.sym.H
+H_Q = tkl.sym.H
```
raikonenfnu, line 32: Are we using H_Q or H_KV or is it for future use?

harsh-nod: I was going to have all the attention kernels use some standard notation so its easier to compare the different kernels. In this PR, I only modified prefill attention, but will modify the others in another PR.

iree/turbine/kernel/wave/templates/attention_common.py
```diff
@@ -0,0 +1,45 @@
...
+H_KV = tkl.sym.H
```
raikonenfnu, line 33: Can we add some docs on what H_KV, N_Q, N_KV, D_Q, D_KV stand for?

raikonenfnu (approved):
Overall code looks great! slight/minor comment but not going to block landing

iree/turbine/kernel/wave/templates/attention_common.py
harsh-nod: sure

raikonenfnu (approved):

lit_tests/kernel/wave/attention.py
```diff
@@ -1181,3 +1125,59 @@ def test_paged_flash_decoding():
...
+        torch.manual_seed(0)
+        q = torch.randn(q_shape, dtype=torch.float16)
+        k = torch.randn(k_shape, dtype=torch.float16)
+        v = torch.randn(v_shape, dtype=torch.float16)
+        output = torch.zeros(o_shape, dtype=torch.float32)
+        offsets = torch.ones(shape.num_seqs, dtype=torch.int32)
+        seq_lens = torch.ones(shape.num_seqs, dtype=torch.int32)
+        print(prefill_attention(q, k, v, offsets, seq_lens, output).module_op)
+
+        # CHECK-LABEL:       func.func @prefill_attention
```
raikonenfnu, line 1167: Ahhh I think the lit test here needs to be re-worked, IIRC there shouldn't be vector.maskedloads after the reordering patch that we pushed up

harsh-nod: yeah they have been reordered. I updated the lit test so it should pass now.

raikonenfnu (approved):

raikonenfnu:
Hey @harsh-nod we may need to do the torch.cuda.cache clear thing, seems like it is consistently breaking in `torch.OutOfMemoryError`

```
FAILED tests/kernel/wave/wave_gemm_test.py::testBatchedGemm[True-shape1] - torch.OutOfMemoryError: HIP out of memory. Tried to allocate 48.00 MiB. GPU 5 has a total capacity of 191.98 GiB of which 0 bytes is free. Of the allocated memory 5.34 MiB is allocated by PyTorch, and 672.00 KiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_HIP_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

raikonenfnu:
or maybe we can split the tests

harsh-nod:
> or maybe we can split the tests

Based on our discussion, reducing number of processes for pytest now, but lets revisit after the sprint.

PR #416, opened 2025-01-25
[TKW] Fixes for `@conditional` and paged attention

Hardcode84:
* Fix `proxy.create_arg` as `sympy.StrictlyLessThan` and friends are not subclass of sympy.Expr.
* It is now possible to directly use sympy expressions in `@conditional` arg.
* Fix `DCE` to not remove ops if they contain nested ops with side-effects (e.g. writes inside conditional).
* Update `get_users`/`get_inputs`/`node_args` ops to properly handle vars captured through placeholders.
* Update `CustomOp.erase` to cleanup dead captures when deleting placeholders.
* Update paged attention template.

### DIFF
--- a/iree/turbine/kernel/wave/templates/paged_decode_attention.py
+++ b/iree/turbine/kernel/wave/templates/paged_decode_attention.py
@@ -74,6 +74,13 @@ def get_paged_decode_attention_kernels(
     head_ratio = shape.num_query_heads // shape.num_kv_heads
     B_WAVES = 1
 
+    # T represents the indices of the sequence tokens.
+    # T is dynamic and is distributed across workgroups and is tiled.
+    count = sympy.Piecewise(
+        (sympy.ceiling(T / U), WORKGROUP_0 < sympy.Mod(T, U)),
+        (sympy.floor(T / U), True),
+    )
+
     def phase_0_constraints():
         # K1, K2 are reduction dimensions that are fixed (not distributed) so
         # they are not part of the constraints.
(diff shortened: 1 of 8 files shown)


iree/turbine/kernel/wave/templates/paged_decode_attention.py
```diff
@@ -272,8 +273,10 @@ def loop(
         res = res_mm * reciprocal_sum
         res_max_log_sum = res_max + tkw.log2(res_sum)
 
-        tkw.write(res_max_log_sum, output_max, elements_per_thread=1)
-        tkw.write(res, output, elements_per_thread=STORE_ELEMS_PER_THREAD)
```
harsh-nod, line 276: Could you also move the operations before the write and after the for loop into the conditional?

iree/turbine/kernel/wave/utils.py
```diff
@@ -17,19 +17,22 @@
...
+    ApplyExpr,
+    Conditional,
     CustomOp,
-    Reduction,
-    GetResult,
     ExtractSlice,
+    GetResult,
     IterArg,
-    Reshape,
+    MMA,
```
harsh-nod, line 26: Thanks for the cleanup :)

harsh-nod (approved):
overall lgtm!

iree/turbine/kernel/wave/templates/paged_decode_attention.py
harsh-nod: Just want to make sure that also works, since we wouldn't want to do that work if count is 0.

iree/turbine/kernel/wave/utils.py
```diff
@@ -182,24 +185,43 @@ def DCE(trace: CapturedTrace):
...
+            return False
+
+        subgraph = custom.graph.subgraphs[custom.subgraph_name]
+        for node in subgraph.nodes:
+            if is_global_write(node) or has_nested_writes(node):
+                return True
+
+        return False
+
+    def is_removable_operator(node: fx.Node) -> bool:
```
raikonenfnu, line 207: Shouldn't we also check for no users?

Hardcode84: `if custom.users` below

iree/turbine/kernel/wave/templates/paged_decode_attention.py
Hardcode84: done

PR #418, opened 2025-01-27
Add functions that prepare tensors as input args for IREE tools

sogartar:
One drawback of using npy files is that they don't support some datatypes.

This change adds functionality to prepare arguments from Torch tensors in the form
--input=1x2xbf16=@arg0.bin
These can then be passed to tools like iree-run-module and iree-benchmark-module.

### DIFF
--- a/iree/turbine/support/conversions.py
+++ b/iree/turbine/support/conversions.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from typing import Any, Callable
+from collections.abc import Sequence
 
 import numpy as np
 import torch
@@ -160,3 +161,28 @@ def torch_dtype_to_numpy(torch_dtype: torch.dtype) -> Any:
         return TORCH_DTYPE_TO_NUMPY[torch_dtype]
     except KeyError:
         raise UnknownDTypeError(torch_dtype)
+
+
+def torch_dtyped_shape_to_iree_format(
+    shape_or_tensor: Sequence[int] | torch.Tensor,
+    /,
+    dtype: torch.dtype | None = None,
+) -> str:
+    """Example:
+    shape = [1, 2, 3]
+    dtype = torch.bfloat16
+    Returns
+    "1x2x3xbf16"
+    """
+    if isinstance(shape_or_tensor, torch.Tensor):
+        dtype = dtype or shape_or_tensor.dtype
+        return torch_dtyped_shape_to_iree_format(shape_or_tensor.shape, dtype)
+    else:
+        if dtype is None:
+            raise ValueError(
+                "dtype must be provided when shape is given instead of a tensor."
+            )
+    shape_str = "x".join([str(d) for d in shape_or_tensor])
+    shape_dtype_delimiter = "x" if len(shape_or_tensor) > 0 else ""
+    dtype_str = TORCH_DTYPE_TO_IREE_TYPE_ASM[dtype]
+    return f"{shape_str}{shape_dtype_delimiter}{dtype_str}"
--- a/iree/turbine/support/tools.py
+++ b/iree/turbine/support/tools.py
@@ -1,0 +1,75 @@
+# Copyright 2025 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import torch
+from typing import Collection, Iterable
+from os import PathLike
+
+from .conversions import torch_dtyped_shape_to_iree_format
+
+
+def iree_tool_format_cli_input_arg(arg: torch.Tensor, file_path: str | PathLike) -> str:
+    """Format the CLI value for an input argument.
+    Example:
+    iree_tool_format_cli_input_arg(torch.empty([1,2], dtype=torch.float32), "arg0.bin")
+    Returns:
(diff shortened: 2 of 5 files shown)


tests/support/conversions_test.py
```diff
@@ -0,0 +1,24 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import torch
+from iree.turbine.support.conversions import torch_dtyped_shape_to_iree_format
+
+
+def test_torch_dtyped_shape_to_iree_format():
+    iree_format = torch_dtyped_shape_to_iree_format([1, 2, 3], dtype=torch.bfloat16)
+    assert iree_format == "1x2x3xbf16"
```
ScottTodd, line 13: Thanks for the unit test! Can you please also add cases for other formats? The list of test cases doesn't need to be exhaustive, but it should cover a few other representative cases and ideally error conditions

Should have a few other shapes and dtypes. Could parameterize the test case too https://docs.pytest.org/en/stable/how-to/parametrize.html

something like
```python
@pytest.mark.parametrize(
    "shape,dtype,expected",
    [
      ([1,2,3], torch.bfloat16, "1x2x3xbf16"),
      ([1,2,3], torch.float32, "1x2x3xf32"),
    ]
)
```

tests/support/tools_test.py
```diff
@@ -0,0 +1,35 @@
...
+from iree.turbine.support.tools import iree_tool_prepare_input_args
+import tempfile
+from pathlib import Path
+
+
+def test_iree_tool_prepare_input_args():
+    arg0 = torch.tensor([1.1, 2.2, 3.3, 4.4], dtype=torch.bfloat16)
+    arg1 = torch.tensor([[4, 5], [6, 7]], dtype=torch.int8)
+    args = [arg0, arg1]
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
ScottTodd, line 17: Could use the `tmp_path` fixture here instead: https://docs.pytest.org/en/stable/how-to/tmp_path.html

ScottTodd (commented):
Cool, having more helper functions in the installable tools will be great. I'd love to have enough of these utils available for test suites too, so we don't constantly rewrite them downstream:

* https://github.com/iree-org/iree-test-suites/blob/main/onnx_ops/import_onnx_tests_utils.py
* https://github.com/iree-org/iree-test-suites/blob/main/onnx_models/utils.py
* etc.

iree/turbine/support/tools.py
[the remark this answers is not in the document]
sogartar: I added documentation.

iree/turbine/support/conversions.py
```diff
@@ -156,3 +156,28 @@ def torch_dtype_to_numpy(torch_dtype: torch.dtype) -> Any:
         return TORCH_DTYPE_TO_NUMPY[torch_dtype]
     except KeyError:
         raise UnknownDTypeError(torch_dtype)
+
+
+def torch_dtyped_shape_to_iree_format(
```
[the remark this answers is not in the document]
sogartar: Done.

iree/turbine/support/tools.py
```diff
@@ -0,0 +1,65 @@
...
+def iree_tool_prepare_input_args(
+    args: Collection[torch.Tensor],
+    /,
+    *,
+    file_paths: Iterable[str] | None = None,
+    file_path_prefix: str | None = None,
+) -> list[str]:
+    """Write the raw contents of tensors to files without any metadata.
+    Returns the CLI input args description.
+    If file_path_prefix will chose a default naming for argument files.
```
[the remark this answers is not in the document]
sogartar: I improved the docstring.

iree/turbine/support/tools.py
```diff
@@ -0,0 +1,65 @@
...
+    """Write the contents of the tensor as they are in memory without any metadata."""
+    with open(file_path, "wb") as f:
+        f.write(tensor.cpu().view(dtype=torch.int8).numpy().data)
+
+
+def iree_tool_prepare_input_args(
+    args: Collection[torch.Tensor],
+    /,
+    *,
+    file_paths: Iterable[str] | None = None,
```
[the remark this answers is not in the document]
sogartar: Added `PathLike` to type hints.

tests/support/conversions_test.py
sogartar: I refactored the tests as you suggested.

tests/support/tools_test.py
sogartar: Done.
I did not know about that built-in fixture.

ScottTodd (approved):
LGTM! Please resolve the DCO check

iree/turbine/support/tools.py
```diff
@@ -0,0 +1,75 @@
...
+        "4x5xi8=@/some/path/arg1.bin"
+    ]
+    ```
+
+    This results can be prefixed with "--input=" to arrive at the final CLI flags
+    expected by IREE tools.
+
+    Exactly one of file_paths and file_path_prefix must be provided.
+    Does not create parent directory(s).
+    """
```
ScottTodd, line 60: FYI this did not format correctly on the website: https://iree-turbine.readthedocs.io/en/latest/core/support.html#iree.turbine.support.tools.iree_tool_prepare_input_args

image

* Docs: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html
* Example: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

PR #422, opened 2025-01-27
Mark args as outputs if *any* user is a write

GMNGeoffrey:
Input arguments are marked as read-only. Therefore, if there's any write the argument has to be marked as an output.

This fixes most of the issues noted in https://github.com/iree-org/iree-turbine/issues/375 (reordering and removing arguments is still a problem).


GMNGeoffrey:
Ugh sorry for the newb formatting mistakes. Will fix in a sec

Hardcode84 (approved):

GMNGeoffrey:
Could someone merge this for me please? (if not waiting for Harsh's review as well)

PR #423, opened 2025-01-28
Add print options to Wave pass pipeline

GMNGeoffrey:
I found these very helpful while debugging. I actually didn't realize there was a `print_trace` function before, so I had modified the `__str__` method on `CapturedTrace`. I've adapted it to use `print_trace` instead. I ported over some of the things I put in the `__str__` method though, like labels for the regions. I think those might be useful for lit tests as well (I had to update the lit tests for it).

Note that I removed a number of comments that mostly repeated the name of the function. I kept ones that seemed like they were elaborating though.

### DIFF
--- a/iree/turbine/kernel/compiler/kernel_codegen.py
+++ b/iree/turbine/kernel/compiler/kernel_codegen.py
@@ -320,11 +320,16 @@ class KernelSignature:
     def __repr__(self):
         parts = []
         for b in self.bindings:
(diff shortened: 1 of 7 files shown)


GMNGeoffrey:
Ok, weird that the lit tests are failing on CI. I swear I ran them locally...

harsh-nod (approved):
thank you! this is very nice and will be a helpful debugging tool for all :)

PR #424, opened 2025-01-28
[TKW] Reordering of shared load-write to minimize lds barriers

raikonenfnu:
In order to improve performance (4.3 ms -> 3.9 ms on SDXL unet attention shape of B0: 2, B1: 20, (M, K2): 1024: K1: 64). We introduce reordering of read_from_global -> write_to_shared -> read_from_shared ordering. Previously, the generated instruction ordering used to look like:
```
read_from_global lhs
write_to_shared lhs
shared_barrier
read_from_shared lhs
read_from_global rhs
write_to_shared rhs
shared_barrier
read_from_shared rhs
```
For this simple example, we have 2 shared barriers. 
Ideally, we can do it with 1 barrier and should look something like:
```
read_from_global lhs
write_to_shared lhs
read_from_global lhs
write_to_shared lhs
shared_barrier
read_from_shared lhs
read_from_shared rhs
```

In order to get the above optimal form, we need to implement these changes:

1. Implement a `move_node_after` helper function that can move src node to a location after a target node.
2. In `promotion.py`, we keep track of the last/most recent `last_write_to_shared` and insert the new `read_from_global` + `write_to_shared` right after the `last_write_to_shared`, this essentially will isolate all the `read_from_shared` to be at the end.
3.  For correctness and numerical stability, in `barriers.py` we are adding an extra barrier to the start of a tiled-reduction loop if there exist any shared reads.
4. For correctness, in `codegen.py` and `hoisting.py`, we update the match between `captured_vars` inside a subgraph and it's corresponding `implicit_capture` values in the root graph to use `.meta["lifted"]` instead of ordering of `implicit_capture`


iree/turbine/kernel/wave/hoisting.py
```diff
@@ -94,8 +99,8 @@ def hoist_loop_invariant_ops(trace: CapturedTrace, constraints: list[Constraint]
                         hoistable_op.replace_all_uses_with(new_op)
                         hoistable_op.erase()
                         if isinstance(hoistable_op, Read):
-                            capture_arg = captured_vars.index(hoistable_op.memory)
-                            new_op.update_arg("memory", implicit_captures[capture_arg])
+                            root_var = hoistable_op.memory.meta["lifted"]
```
harsh-nod, line 102: For a separate PR, can we add a get_lifted or get_captured_var function?

iree/turbine/kernel/wave/promotion.py
```diff
@@ -34,11 +34,21 @@ def apply_padding(
     )
 
 
-def apply_promotion_pattern(custom_node: Read | Write, allocate_node: Allocate):
+def apply_promotion_pattern(
```
harsh-nod, line 37: Could you add your comment from the PR in here? Would be helpful to understand what is going on.

iree/turbine/kernel/wave/hoisting.py
raikonenfnu: SG, will put in separate PR :)

iree/turbine/kernel/wave/promotion.py
raikonenfnu: done!

tests/kernel/wave/attention/vanilla_attention_test.py
```diff
@@ -133,6 +133,7 @@ def repeat(
             q_reg = tkw.read(q, elements_per_thread=LOAD_ELEMS_PER_THREAD_QK)
             # b_reg: tkw.Register[B, N, K, tkl.f16]
             k_reg = tkw.read(k, elements_per_thread=LOAD_ELEMS_PER_THREAD_QK)
```
harsh-nod, line 135: Can we move v back?

harsh-nod (approved):
lgtm!

PR #425, opened 2025-01-28
Add extend attention

harsh-nod:
This PR adds support for extend attention (also known as chunked prefill).  A new kernel template is added along with an end to end test with a reference kernel.

### DIFF
--- a/iree/turbine/kernel/wave/expansion/expansion.py
+++ b/iree/turbine/kernel/wave/expansion/expansion.py
@@ -632,7 +632,8 @@ def fixup_reduction_nodes(
     expansion_context: ExpansionContext,
 ):
     reduction_context = expansion_context.reduction_context
-    for reduction in trace.walk(lambda x: isinstance(get_custom(x), Reduction)):
+    reduction_nodes = trace.walk(lambda x: isinstance(get_custom(x), Reduction))
+    for reduction in reversed(reduction_nodes):
         reduction = get_custom(reduction)
         reduction_subgraph = trace.get_subgraph(reduction.subgraph_name)
         output = get_custom(get_last(reduction_subgraph.nodes))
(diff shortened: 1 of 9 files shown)


iree/turbine/kernel/wave/expansion/expansion.py
```diff
@@ -632,7 +632,8 @@ def fixup_reduction_nodes(
     expansion_context: ExpansionContext,
 ):
     reduction_context = expansion_context.reduction_context
-    for reduction in trace.walk(lambda x: isinstance(get_custom(x), Reduction)):
+    reduction_nodes = trace.walk(lambda x: isinstance(get_custom(x), Reduction))
+    for reduction in reversed(reduction_nodes):
```
raikonenfnu, line 636: For education/documentation on github, can you give explain why we are reversing the order?

iree/turbine/kernel/wave/promotion.py
```diff
@@ -77,7 +77,7 @@ def apply_promotion_pattern(
             # We move CustomOp/Read up to the last write_to_shared_mem S.T
             # all reads from shared mem happens only after all read from globals
             # and write to shared mem happen. Which will minimize lds_barrier count.
-            if last_write_to_shared:
+            if last_write_to_shared and last_write_to_shared.graph == custom_node.graph:
```
raikonenfnu, line 80: For education purposes, can you explain a little why we need this for extend attention?

raikonenfnu (approved):
LGTM! just some questions (but no code action needed)

iree/turbine/kernel/wave/promotion.py
harsh-nod: Sure, since we have multiple nested region ops in the graph, then the last_write_to_shared
is in a different graph as custom.node. So I added this check to make sure they are always in the same graph. Also added a comment.

iree/turbine/kernel/wave/expansion/expansion.py
harsh-nod: In situations where we have multiple reductions, and the outputs of a
    reduction are used as inputs to another reduction, we need to ensure
    the fixup is done in the correct order, specifically from the last
    reduction to the first reduction since that is the order in
    which expansion proceeds. Also added a comment.

PR #432, opened 2025-01-29
Collection of QoL changes

GMNGeoffrey:
This is a bad PR in the sense that it does a bunch of different things. This is a collection of changes I made while developing kernels, mostly around improving debugging with more verbose and specific error messages. But there's also some other stuff in here. Most of these don't really seem sufficiently important to break out into their own PR, but it seemed a shame to just throw away things that seemed like improvements. I did try to clean things up a bit to make them actually upstreamable.

One note is that I'm not really sure what the convention is for Exception types in the project, so a lot of these are just RuntimeError. That's not awesome, but I think it's still a lot more helpful than nothing. I mostly tried to avoid `raise ... from` as in my experience these usually result in unhelpfully verbose stacks.

Let me know if you'd like me to do this differently, there are any of these changes that you think warrant more discussion, or you don't think they're worth this grab bag PR. If this doesn't land before I head off, feel free to modify into something uncontroversial and merge.

### DIFF
--- a/iree/turbine/kernel/_support/indexing.py
+++ b/iree/turbine/kernel/_support/indexing.py
@@ -117,6 +117,18 @@ class IndexingContext:
         self.frozen_subs: list[tuple[IndexSymbol, int]] = []
         self.unbacked_symbols: list[IndexSymbol] = []
 
+    def __str__(self):
+        return (
+            f"IndexingContext("
+            f"subs: {self.subs}\n"
+            f"special_subs: {self.special_subs}\n"
+            f"shaped_bindings: {self.shaped_bindings}\n"
+            f"dyn_dims: {self.dyn_dims}\n"
+            f"frozen_subs: {self.frozen_subs}\n"
+            f"unbacked_symbols: {self.unbacked_symbols}\n"
+            ")"
+        )
+
     def next_dyn_dim(self) -> IndexSymbol:
         s = index_symbol(f"D{len(self.dyn_dims)}")
         self.dyn_dims.append(s)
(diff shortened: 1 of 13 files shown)


GMNGeoffrey:
Sorry, should've run tests after doing cleanup, obviously....

raikonenfnu:
I know it can get quite long, but would be lovely if we have the list of fixes on the PR message

GMNGeoffrey:
I removed some of the more substantive changes, fixed the test failures, and wrote down what the PR does.

GMNGeoffrey:
Ping on this?

harsh-nod:
Can you break this into individual commits and we can start landing those

GMNGeoffrey:
Sure, it was mostly one PR so that I could send it all before I went on vacation. I'll look into breaking it up a bit

GMNGeoffrey:
I've extracted a few PRs. I think what will remain here actually has a coherent single purpose of adding missing validation and error messages.

GMNGeoffrey:
I think this is now sufficiently narrow to be a reasonable PR. It still has a few extraneous things like adding some missing return types to some functions, but that doesn't seem like something that's worth breaking off into its own PR and it also seems a bit silly to just delete them.

GMNGeoffrey:
The big diff in iree/turbine/kernel/wave/codegen/emitter.py is due to windows newlines, which I'm removing in #544

GMNGeoffrey:
Whoops, I accidentally merged #548 into all my PRs. Will revert once the CI runs finish

GMNGeoffrey:
CI failure is the single test already failing at head (#620). Would appreciate review here: it's annoying to keep having to resolve merge conflicts for this

GMNGeoffrey:
@harsh-nod @raikonenfnu @Hardcode84 could I get review here please?

Hardcode84 (approved):

PR #437, opened 2025-01-30
[TKW] Keep disambiguating number in expanded node name

GMNGeoffrey:
Nodes get named like `add`, `add_1`, etc to disambiguate ones with the same name. But the node naming in expansion throws away this suffix. I think it's helpful to leave it in, when viewing the IR. Nodes don't have to have unique names, but it's nice if they do.

In updating the lit tests, I also noticed that the node names in scheduling.py had a different convention for dimension expansion than everything else, writing it `K_1` instead of `K:1`. To further confuse things, it looked like torch was actually incrementing that trailing `_1` if node names conflicted. Since my change made things not conflict it appeared to change the expansion dimension number! I changed it so that scheduling instead preserves the original node name. Given that we are copying these nodes into a new graph, I think it is fine to repeat node names (and in fact it's better because then you can relate them). Node names also shouldn't be load bearing for anything and conflicts shouldn't cause problems, so this shouldn't have the potential to create any bugs or failures (famous last words).


GMNGeoffrey:
Huh, it doesn't seem like the `skip-ci` trailer works in this repo. Note that the turbine contributing guidelines point to the IREE contributing guidelines, which document the [ci control options](https://iree.dev/developers/general/contributing/#build-systems). You should perhaps consider adding them to turbine. Whoever came up with them created a really cool and useful feature 😜  <-- @ScottTodd

GMNGeoffrey:
Ping on this @harsh-nod? Wanted to get your thoughts on this before fixing all the lit tests

harsh-nod:
How about we add a more descriptive name such as including the placeholder name as well (and maybe other attributes) and then add a number after if necessary?

GMNGeoffrey:
> How about we add a more descriptive name such as including the placeholder name as well (and maybe other attributes) and then add a number after if necessary?

Sure, although I think that's somewhat orthogonal to this change. Right now expansion deletes all the disambiguating information. I'm just making it not do that. We could also add additional disambiguating information. It might make sense to do that in one PR though because both involve updating all the lit tests. Related to that, I think it might be worth considering having the human-readable and lit-test printed output be different so that these sorts of changes wouldn't require updating the lit tests. IDK

GMNGeoffrey:
Whoops, I accidentally merged #548 into all my PRs. Will revert once the CI runs finish

lit_tests/kernel/wave/scheduling.py
```diff
@@ -120,45 +120,46 @@ def test_gemm_pipelined():
         # CHECK-NEXT: %rotating_reg_1
         # CHECK-NEXT: %rotating_reg_2
         # CHECK-NEXT: %rotating_reg_3
-        # CHECK-NEXT: %read_shared_M_0_N_0_K_0
-        # CHECK-NEXT: %read_shared_M_0_N_0_K_1
+        # CHECK-NEXT: %read_4_shared_M_0_N_0_K_0
+        # TODO: Why did this change from K_1 to K_0???
+        # CHECK-NEXT: %read_2_shared_M_0_N_0_K_0
```
GMNGeoffrey, line 125: This change is weird. The second read here went from K_1 to K_0. The place where its used also change. There are a couple places lower down where something similar happens. K_2 and K_3 both become K_1. I don't understand how my change could have caused this though: it doesn't touch how expansion name suffixes are generated. Unless the logic adding the trailing number suffix was somehow getting invoked for these after expansion? Actually, that might explain it.

lit_tests/kernel/wave/scheduling.py
```diff
@@ -120,45 +120,46 @@ def test_gemm_pipelined():
...
         # CHECK-NEXT: %read_21
         # CHECK-NEXT: %read_22
         # CHECK-NEXT: %scheduling_group_barrier
         # CHECK-SAME: ({Operation.READ_SHARED: 2, Operation.READ_GLOBAL: 2}, 0)
-        # CHECK-NEXT: %read_shared_M_0_N_1_K_0
-        # CHECK-NEXT: %read_shared_M_0_N_1_K_1
+        # CHECK-NEXT: %read_4_shared_M_0_N_1_K_0
```
GMNGeoffrey, line 130: Also these are f"{dim_name}_{dim_val}" instead of f"{dim_name}:{dim_val}". It kind of looks like these are getting process by expansion, which was stripping the numeric disambiguator and adding "_shared" and then later something else was doing its own expansion and then adding a numeric disambiguator, but actually treating the last dim value as the thing to increment.

lit_tests/kernel/wave/scheduling.py
GMNGeoffrey: I have diagnosed this, which I think was the same reason that the dimensions were being written as `K_1` instead of `K:1`. My best guess is that the fx node naming convention by default replaces `:` with `:` and that then it further disambiguates by incrementing a trailing `_\d` if node names conflict. I've instead made scheduling preserve the original node names.

iree/turbine/kernel/wave/utils.py
```diff
@@ -818,6 +818,7 @@ def graph_copy(graph: fx.Graph) -> tuple[fx.Graph, dict[fx.Node, fx.Node]]:
     for node in graph.nodes:
         custom = get_custom(node)
         new_node = custom.copy(
+            new_name=node.name,
```
GMNGeoffrey, line 821: This is a utility function, so it could potentially be used outside of scheduling (although currently it appears not to be), but I think preserving the existing node names is reasonable behavior given that they are being copied into a new graph.

GMNGeoffrey:
Harsh could you please take another look and make sure you're ok with my changes to the node naming during the scheduling pass? See details in the PR description and my comments.

GMNGeoffrey:
MI250 failure is an infrastructure failure. MI300X failure looks like rpe attention is flaky :-/

harsh-nod:
Sorry I didn't get to this earlier, but looks good, thank you!

PR #443, opened 2025-02-01
[Wave] Improve support for non-primary wg constraint to fix GQA decode

raikonenfnu:
This PR implements missing feature to enabled GQA for non-transposedV on prefill attention. In order to do that there are two main things this PR introduces:

1. GQA by definition dictates that Q_head >> KV_head. However, previously we set the same symbolic dims `H` on `Q_head` and `KV_head`, this ends up giving incorrect result during GQA non transposedV extend attention. For example, in the case of `seq_len=64, q_head=4, kv_head=1, h_dim=128` with same symbolic dim, it still expect v_shape of (tensor<64x4x128xf16>) which will produce transposed-v read/gather with `stride=128*4=512` but in reality it should be `tensor<64x1x128xf16>` with `stride=128*1=128`.

2. Once we have the proper/correct "non-primary" workgroup constraint set, kernel starts giving numerically wrong results. Upon investigation this is because once we set a "non-primary" workgroup constraint, the original/primary workgroup constraint does not get delinearized since it is overwritten in `get_workgroup_dims` by the new/"non-primary" workgroupConstraint. To fix this, we refactored `get_workgroup_dims` to return all the dims as a lit, and refactored `initialize_workgroup_constraint` to be able to delinearize both "primary" and "non-primary" workgroupConstraints and dims.

### DIFF
--- a/tests/kernel/wave/attention/prefill_attention_test.py
+++ b/tests/kernel/wave/attention/prefill_attention_test.py
@@ -124,7 +124,7 @@ def testPrefillAttention(
     (query, key, value, start_offsets, seq_lens) = create_inputs(shape, seq_lens, dtype)
 
     output_shape = (shape.total_seq_len, shape.num_query_heads, shape.head_size_kv)
-    permuted_value = value.permute(1, 2, 0)
+    permuted_value = value
     # Run the wave kernel.
     (prefill, hyperparams) = get_prefill_attention_kernel(
         shape,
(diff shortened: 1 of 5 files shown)


tests/kernel/wave/attention/prefill_attention_test.py
```diff
@@ -124,7 +124,7 @@ def testPrefillAttention(
     (query, key, value, start_offsets, seq_lens) = create_inputs(shape, seq_lens, dtype)
 
     output_shape = (shape.total_seq_len, shape.num_query_heads, shape.head_size_kv)
-    permuted_value = value.permute(1, 2, 0)
+    permuted_value = value
```
harsh-nod, line 127: minor: remove this, we can just use value directly?

harsh-nod (approved):
lgtm!

tests/kernel/wave/attention/prefill_attention_test.py
raikonenfnu: done!

PR #445, opened 2025-02-01
[Wave] Add API to pass in dtypes for prefill

raikonenfnu:


iree/turbine/kernel/wave/templates/prefill_attention.py
```diff
@@ -23,7 +27,25 @@ def get_prefill_attention_kernel(
     k_shape: tuple[int],
     v_shape: tuple[int],
     o_shape: tuple[int],
+    input_dtype: Optional[torch.dtype] = torch.float16,
+    output_dtype: Optional[torch.dtype] = torch.float32,
+    seq_dtype: Optional[torch.dtype] = torch.int32,
```
harsh-nod, line 32: minor: can we rename this to size_type because it is the type of the sequence lengths / indices?

harsh-nod (approved):
lgtm! I will add something like this to the extend attention template as well.

iree/turbine/kernel/wave/templates/prefill_attention.py
raikonenfnu: done!

PR #450, opened 2025-02-04
Remove transpose v in extend attention

harsh-nod:
This PR removes the need for the value tensors to be permuted for the attention kernel.


raikonenfnu (approved):
LGTM!

PR #456, opened 2025-02-05
Sglang integrarion: Fix dtype mismatch

yichiche:
- Cast all int64 values to int32 to fix compatibility issue of sglang interface.
- Add support for tkl.bf16 to expand data type handling.

### DIFF
--- a/iree/turbine/kernel/wave/templates/extend_attention.py
+++ b/iree/turbine/kernel/wave/templates/extend_attention.py
@@ -40,6 +40,7 @@ def get_extend_attention_kernel(
 
     assert wave_input_dtype in [
         tkl.f16,
+        tkl.bf16,
     ], f"Unsupported input datatype: {wave_input_dtype}"
     assert (
         wave_output_dtype.is_float_asm()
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/wave/templates/extend_attention.py
```diff
@@ -202,11 +204,15 @@ def first_loop(
                 elements_per_thread=LOAD_ELEMS_PER_THREAD_QK,
                 mapping=q_mapping,
             )
+            if wave_input_dtype == tkl.bf16:
+                q_reg = tkw.cast(tkw.cast(q_reg, tkl.f32), tkl.f16)
```
harsh-nod, line 208: I realize now that this is not necessary, so we can remove this here and elsewhere.

iree/turbine/kernel/wave/templates/extend_attention.py
```diff
@@ -215,12 +221,14 @@ def first_loop(
             e_delta = tkw.exp2(x_j - m_j)
             e_init = partial_sum * e_delta_max
             d_j = tkw.sum(e_delta, e_init, dim=N_KV)
-            imm_f16 = tkw.cast(e_delta, wave_input_dtype)
```
raikonenfnu, line 218: I think we should also keep this as `imm_f16 = tkw.cast(e_delta, wave_input_dtype)` just because we may have bf16 inputs

harsh-nod:
Closing this out as not needed.

PR #457, opened 2025-02-05
[Wave] Make extend sequence dimension dynamic

harsh-nod:
This PR modifies the extend attention kernel so that the query/key/value/output extend sequence length is dynamic. This is required to avoid recompilation of the kernel for different values of the sequence length.


raikonenfnu (approved):
LGTM!

PR #458, opened 2025-02-05
Include numbered version for release candidates

sogartar:
We can't make multiple different versions for the same day.

With this change the release candidates will have a version of the form

3.2.0rc20250205.0
3.2.0rc20250205.1
...

instead of just
3.2.0rc20250205

### DIFF
--- a/.github/workflows/build_packages.yml
+++ b/.github/workflows/build_packages.yml
@@ -25,8 +25,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-        with:
-          fetch-tags: true
 
       - name: "Setting up Python"
         uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
@@ -38,13 +36,10 @@ jobs:
 
       - name: Build iree-turbine release candidate
         run: |
-          version_name="$(./build_tools/compute_local_version.py -rc --write-json)"
+          ./build_tools/compute_local_version.py \
+            -rc --write-json \
+            --version-suffix ".${{ github.run_number }}"
           ./build_tools/build_release.py --no-download
-
-      - name: Tag version
-        run: |
-          git tag "$version_name"
-          git push origin "$version_name"
 
       - name: Upload python wheels
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
--- a/.github/workflows/build_packages.yml
+++ b/.github/workflows/build_packages.yml
@@ -25,6 +25,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-tags: true
 
       - name: "Setting up Python"
         uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
@@ -36,8 +38,13 @@ jobs:
 
       - name: Build iree-turbine release candidate
         run: |
-          ./build_tools/compute_local_version.py -rc --write-json
+          version_name="$(./build_tools/compute_local_version.py -rc --write-json)"
           ./build_tools/build_release.py --no-download
+
+      - name: Tag version
+        run: |
+          git tag "$version_name"
+          git push origin "$version_name"
 
       - name: Upload python wheels
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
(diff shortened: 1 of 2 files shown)


sogartar:
This change relies on having the github workflow fetch all tags, not just the branch that it is building. I assumed this is the case.

sogartar:
I changed it to fetch the tags.

.github/workflows/build_packages.yml
```diff
@@ -36,9 +38,14 @@ jobs:
 
       - name: Build iree-turbine release candidate
         run: |
-          ./build_tools/compute_local_version.py -rc --write-json
+          version_name="$(./build_tools/compute_local_version.py -rc --write-json)"
```
ScottTodd, line 41: Did you test this (on your fork)? Variables set like this are not persisted between workflow steps automatically.

See
* https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable
* https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-output-parameter


```suggestion
          version_name="$(./build_tools/compute_local_version.py -rc --write-json)"
          echo "version_name=${version_name}" >> "$GITHUB_ENV"
```

build_tools/compute_local_version.py
```diff
@@ -44,12 +44,41 @@ def write_version_to_file(version_file, version):
...
+    """
+    Get the next numbered tag that does not exists.
+    E.g. iree-turbine-3.1.0rc20250205. -> iree-turbine-3.1.0rc20250205.3
+    """
+    tags = (
+        subprocess.check_output(["git", "tag", "--list", f"{tag_prefix}*"])
+        .decode()
+        .strip()
+        .splitlines()
+    )
```
ScottTodd, line 66: I'm conflicted about pushing a tag for each nightly release. We do that in iree-org/iree, but it generates a lot of tags.

You could instead use the workflow run number, as we used to do there: https://github.com/iree-org/iree/blob/92ac1b2deea48c018ecb1f66a05d454b099ae294/.github/workflows/schedule_candidate_release.yml#L39-L44

Those values are unique: 
image

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs
>	A unique number for each run of a particular workflow in a repository. This number begins at 1 for the workflow's first run, and increments with each new run. This number does not change if you re-run the workflow run.

WDYT?

.github/workflows/build_packages.yml
```diff
@@ -36,9 +38,14 @@ jobs:
 
       - name: Build iree-turbine release candidate
```
ScottTodd, line 39: > We also can't make multiple different versions for the same day.

Yep. I thought about making a change like this one but figured that nightly releases are "best effort" anyways, and overwriting a nightly release if we want to pick up some commits before the next day isn't a major issue. I'm not opposed to adding a number suffix and allowing multiple releases per day, if you put the work in to make that possible. It's just more code to write and maintain.

.github/workflows/build_packages.yml
sogartar: You are right.
I removed the tagging.

build_tools/compute_local_version.py
sogartar: I removed the git tagging. If someone is interested in the exact commit they could go and retrieve it manually from the workflow.

sogartar: I made it use the `github.run_number` istead.

ScottTodd (approved):

.github/workflows/build_packages.yml
sogartar: I wanted this to improve reproducibility as we are using pinned versions for the projects that depend on IREE Turbine.

ScottTodd (approved):

.github/workflows/build_packages.yml
ScottTodd: Ah, still not tested though? This failed last night:

https://github.com/iree-org/iree-turbine/actions/runs/13172140306/job/36764292832
```
usage: compute_local_version.py [-h] [--write-json]
                                (-stable | -rc | -dev | --version-suffix VERSION_SUFFIX)
compute_local_version.py: error: argument --version-suffix: not allowed with argument -rc/--nightly-release
```

I'd recommend testing on your fork using workflow dispatch, on https://github.com/sogartar/iree-turbine/actions/workflows/build_packages.yml

Hmm... looking through history, https://github.com/iree-org/iree/pull/19456 and https://github.com/iree-org/iree-turbine/pull/328 made the version suffix mutually exclusive with the other options. I had some discussions with @marbre about that... should have put more context in the PRs. Pasting here:

> when running the scripts locally, I certainly like the convenience of just doing -rc or -dev and having the script figure it out for me
> I don't think I'd ever want to do -rc --version-suffix
> if I'm specifying an explicit suffix, I just want that, no other stuff added for me

(oops, this is a good reason to want that combination of options 😅)

PR #465, opened 2025-02-06
[Wave] Teach compiler to handle unaligned attention

raikonenfnu:
Most time on LLM, we'd be facing unaligned shapes since sequence length is not always aligned/padded. In order to support that, we implement unaligned attention kernel along with a couple compiler features S.T we can handle unaligned shapes. Features are:

1. Implement `&` operator to combine multiple masks (causal + unaligned shape)
2. Teach broadcast to handle symbolic broadcasts on unit dims as no-op (important for broadcasting K2 -> [M, K2] for masks)
3. Let ApplyExpr go through expansion, initially caused a numeric bug if not expanded properly.
4. Modify necessary LIT to check for unaligned mask, and e2e to handle unaligned shape

### DIFF
--- a/iree/turbine/kernel/wave/codegen/handlers.py
+++ b/iree/turbine/kernel/wave/codegen/handlers.py
@@ -455,6 +455,18 @@ def handle_div(lhs: Value, rhs: Value) -> OpResult:
     return result
 
 
+@handle_binary_op(operator.and_)
+def handle_and(lhs: Value, rhs: Value) -> OpResult:
+    element_type = get_type_or_element_type(lhs.type)
+    if _is_integer_like_type(element_type) and (
+        element_type.is_signed or element_type.is_signless
+    ):
+        result = arith_d.andi(lhs, rhs)
+    else:
+        raise ValidationError(f"Found unhandled operand type for div: {element_type}")
+    return result
+
+
 @handle_binary_op([operator.gt, gt])
 def handle_gt(lhs: Value, rhs: Value) -> OpResult:
     element_type = get_type_or_element_type(lhs.type)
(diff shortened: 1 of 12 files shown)


tests/kernel/wave/common/shapes.py
```diff
@@ -12,9 +12,18 @@
 _e2e_test_shapes["attention"] = [
     (8, 128, 128, 64, 256),
     (40, 1024, 64, 64, 1024),
+    (32, 1024, 128, 128, 1357),
```
harsh-nod, line 15: remove shape ?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -830,13 +842,14 @@ def handle_extract_slice(emitter: WaveEmitter, node: fx.Node):
...
-        register, _ = node.args
+        register, target_shape = node.args
     except ValueError as e:
         raise ValidationError("Malformed arguments") from e
 
     # Get thread_shape/size for broadcast.
     get_thread_shape = lambda index: max(subs_idxc(x.size) for x in index.values())
 
+    src_thread_size = get_thread_shape(register.index) if register.index else None
     bcast_dim_lane_dim_size = get_thread_shape(node.index)
```
harsh-nod, line 853: rename variable to target thread size

harsh-nod (approved):
lgtm! nice one :)

tests/kernel/wave/common/shapes.py
raikonenfnu: done!

iree/turbine/kernel/wave/codegen/handlers.py
raikonenfnu: done!

PR #467, opened 2025-02-06
Attention with Linear Biases

ftynse:
This implements (causal) ALiBi, attention with linear biases, following
the paper "Train Short, Test Long: Attention with Linear Biases Enables
Input Length Extrapolation" by Press et.al in ICLR 2022.

### DIFF
--- a/tests/kernel/wave/attention/alibi_attention_test.py
+++ b/tests/kernel/wave/attention/alibi_attention_test.py
@@ -1,0 +1,173 @@
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+import torch
+import math
+import iree.turbine.kernel as tk
+from iree.turbine.kernel.lang.global_symbols import *
+from iree.turbine.kernel.wave.utils import (
+    get_default_run_config,
+    get_default_scheduling_params,
+    device_arange,
+    device_full,
+    device_randn,
+    device_zeros,
+    to_default_device,
+)
+from iree.turbine.kernel.wave.constraints import MMAType
+from iree.turbine.kernel.wave.templates.alibi_attention import (
+    get_alibi_attention_kernel,
+)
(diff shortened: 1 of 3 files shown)


playground/attention_alibi_template.py
```diff
@@ -0,0 +1,168 @@
+# Copyright 2025 The IREE Authors
```
raikonenfnu, line 1: Can we move this out of playground? I was thinking it may be good to have it in 
`iree/turbine/kernel/wave/template/attention/flexattention/`  wdyt @harsh-nod ?

ftynse: Moved to a file next to other attentions. FlexAttention is slightly broader and will need some config on top to select a combination of the attention variations (alibi, rpe, causal, softcap, etc.) that may require some changes in core to, e.g., gracefully handle lack of uses for certain arguments.

harsh-nod: Sure that's fine for now. When RPE comes in, we can refactor this into a flexattention folder and generalize this.

tests/kernel/wave/attention/alibi_attention_test.py
```diff
@@ -0,0 +1,173 @@
...
+def validate_accuracy(
+    query: torch.Tensor, key: torch.Tensor, value: torch.Tensor, output: torch.Tensor
+) -> torch.Tensor:
+    # Precompute values.
+    dk_sqrt = math.sqrt(1.0 / query.shape[-1])
+    alibi_slopes = precompute_alibi_slopes(query.shape[0])
+
+    # Straightforward implementation of attention with bias.
+    scores = torch.matmul(query, key.transpose(-1, -2)) * dk_sqrt
+    bias = alibi_slopes.unsqueeze(-1).unsqueeze(-1) * get_relative_positions(
```
raikonenfnu, line 83: We don't need to change for this PR, but for future PRs on `flexattention` related stuff, we can also just import flex_attention and use it directly, as seen also in https://github.com/iree-org/iree-turbine/commit/176d443b0256eac484606bf7fc8cebbea29e2e0b#diff-1f375ba26b8000b24b7d3a662f186833ab26f2a9782dbe099b931fc3263445aeR83

harsh-nod (approved):
lgtm!

PR #480, opened 2025-02-08
[Wave] Fixes to extend attention 

harsh-nod:
This PR adds the following fixes to extend attention:
1. Fixes the way the block indices were being read for the v matrix in the first loop. Because of the layout differences between the first and second mma, we need to read block indices with a different layout and correspondingly elements per thread when loading from the kv cache.
2. Moves the qk-transpose scaling inside the kernel for accuracy reasons.
3. Moves reuse shared allocs pass after minimize global loads.
4. Minor compiler fix regarding use of Sequence vs list

### DIFF
--- a/iree/turbine/kernel/wave/templates/extend_attention.py
+++ b/iree/turbine/kernel/wave/templates/extend_attention.py
@@ -71,7 +71,7 @@ def get_extend_attention_kernel(
     LOG2E = 1.44269504089
     logit_cap *= LOG2E
     dk_sqrt = math.sqrt(1.0 / shape.head_size)
-    layer_scaling = layer_scaling or dk_sqrt * LOG2E
+    layer_scaling = (layer_scaling or dk_sqrt) * LOG2E
 
     constraints: list[tkw.Constraint] = []
     constraints += [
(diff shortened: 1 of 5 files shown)


iree/turbine/kernel/wave/templates/extend_attention.py
```diff
@@ -69,6 +70,8 @@ def get_extend_attention_kernel(
     N_WAVES = 1
     LOG2E = 1.44269504089
     logit_cap *= LOG2E
+    dk_sqrt = math.sqrt(1.0 / shape.head_size)
+    layer_scaling = layer_scaling or dk_sqrt * LOG2E
```
raikonenfnu, line 74: would this be same as `(layer_scaling or dk_sqrt) * LOG2E?

raikonenfnu (approved):
Overall looks good, just that one quick question

iree/turbine/kernel/wave/templates/extend_attention.py
harsh-nod: good catch, that's a bug. will fix

raikonenfnu (approved):

PR #483, opened 2025-02-09
[TKW] Reuse allocs fix

Hardcode84:
* Move `reuse_shared_allocs` after `minimize_global_loads`, as `minimize_global_loads` gets confused by the new alloc structure.
* Insert barrier when replacing alloc to make sure all threads finished their reads/writes to would-be-reused alloc.
* Do not kill `SharedMemoryBarrier` by `DCE`

### DIFF
--- a/iree/turbine/kernel/wave/reuse_shared_allocs.py
+++ b/iree/turbine/kernel/wave/reuse_shared_allocs.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from iree.turbine.kernel._support.tracing import CapturedTrace
 import torch.fx as fx
 
-from ..ops.wave_ops import get_custom, Allocate
+from ..ops.wave_ops import get_custom, Allocate, SharedMemoryBarrier
 from .utils import get_users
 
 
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/wave/reuse_shared_allocs.py
```diff
@@ -55,6 +55,11 @@ def _try_replace(current_node: fx.Node, candidates: list[fx.Node]) -> bool:
     """
     for candidate in candidates:
         if _is_dead(current_node, candidate):
+            if not isinstance(get_custom(current_node.prev), SharedMemoryBarrier):
+                graph = current_node.graph
+                with graph.inserting_before(current_node):
+                    SharedMemoryBarrier().add_to_graph(graph)
```
harsh-nod, line 61: why do we need a shared memory barrier?

harsh-nod: Because we always add a barrier at the beginning of the for loop?

harsh-nod: sounds good thanks!

PR #484, opened 2025-02-09
Bring back maskedload and maskedstore

harsh-nod:
In a previous PR, we lost the ability to generate maskedload and maskedstore (buffer load and store were the only option). This PR enables folks to use maskedloads and maskedstores if they dont want buffer ops.


raikonenfnu (approved):

harsh-nod:
Not required.

PR #485, opened 2025-02-10
[Wave] Add free variable as key to WaveCacher

raikonenfnu:
Previously free variables such as mfma variants, logit_caps, or is_causal flags are not used as keys for caching. This may cause numerics issues in real life use case.


harsh-nod (approved):
lgtm! thanks!

PR #487, opened 2025-02-10
Include Github run numbered in release candidate version

sogartar:
We can't make multiple different versions for the same day.

With this change the release candidates will have a version of the form
3.2.0rc20250205-0
3.2.0rc20250205-1
...
instead of just 3.2.0rc20250205.

### DIFF
--- a/.github/workflows/build_packages.yml
+++ b/.github/workflows/build_packages.yml
@@ -36,7 +36,9 @@ jobs:
 
       - name: Build iree-turbine release candidate
         run: |
-          ./build_tools/compute_local_version.py -rc --write-json
+          ./build_tools/compute_local_version.py \
+            -rc --write-json \
+            --version-suffix "-${{ github.run_number }}"
           ./build_tools/build_release.py --no-download
 
       - name: Upload python wheels
--- a/build_tools/compute_local_version.py
+++ b/build_tools/compute_local_version.py
@@ -19,12 +19,12 @@ import subprocess
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--write-json", action="store_true")
+parser.add_argument("--version-suffix", action="store", type=str)
 
 release_type = parser.add_mutually_exclusive_group(required=True)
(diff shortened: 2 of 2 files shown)


sogartar:
This is the same of a reverted change #458.

I have verified that the [workflow](https://github.com/iree-org/iree-turbine/actions/runs/13248018093) passes fine.
I don't see the upload at https://iree.dev/pip-release-links.html.

ScottTodd:
> I have verified that the [workflow](https://github.com/iree-org/iree-turbine/actions/runs/13248018093) passes fine.

Thanks! The version shows up as `iree_turbine-3.2.0rc20250210.post73` though. Was that intentional?

See the documentation at https://packaging.python.org/en/latest/discussions/versioning/.

I'm also still not sure we really need this extra suffix given the extra complexity.
* If each package has its own extra suffix then manually updating the versions in a file like https://github.com/nod-ai/shark-ai/blob/main/requirements-iree-pinned.txt will be harder.
* iree-turbine has low commit volume (< 10 commits per day), so I would expect us to only very rarely be publishing multiple builds per day

> I don't see the upload at https://iree.dev/pip-release-links.html.

That index is updated by an action in iree-org/iree, that runs when the IREE docs are updated or an IREE release is completed. The workflow can also be triggered manually.
https://github.com/iree-org/iree/blob/624a9fae4f8dd03becfaae69110411c84c4ed6e8/.github/workflows/publish_website.yml#L67-L82

ScottTodd:
You can also test in your fork while still iterating. The test run in this repository pushed the `iree_turbine-3.2.0rc20250210.post73-py3-none-any.whl` package to https://github.com/iree-org/iree-turbine/releases/tag/dev-wheels. We can either keep that or delete it since "dev wheels" and nightly releases are best-effort and may occasionally do weird things.

sogartar:
@ScottTodd, sorry for not closing this in time. Your argument is solid that we can just rerun the pipeline if we need to rebuild and would not be a problem due to a low commit frequency.

PR #490, opened 2025-02-10
[Wave] Make maximum extend sequence length dynamic

harsh-nod:
This PR makes the maximum extend sequence a dynamic kernel argument that is used in the grid calculation. 
Tested with gsm8k to 0.816.

### DIFF
--- a/iree/turbine/kernel/compiler/host_codegen.py
+++ b/iree/turbine/kernel/compiler/host_codegen.py
@@ -60,6 +60,10 @@ def isolated_test_call(
         argument_dims = get_dynamic_dims(sig.kernel_buffer_bindings, dynamic_symbols)
         # Adding unique dynamic dims as inputs.
         input_tensors += [IndexType.get() for _ in list(dict.fromkeys(argument_dims))]
+        # Add additional dynamic symbols as inputs.
+        input_tensors += [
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -60,6 +60,10 @@ def isolated_test_call(
         argument_dims = get_dynamic_dims(sig.kernel_buffer_bindings, dynamic_symbols)
         # Adding unique dynamic dims as inputs.
         input_tensors += [IndexType.get() for _ in list(dict.fromkeys(argument_dims))]
+        # Add additional dynamic symbols as inputs.
+        input_tensors += [
+            IndexType.get() for _ in set(dynamic_symbols).difference(argument_dims)
```
raikonenfnu, line 65: Can we get this captured in the LIT test? I reckon current extend attention LIT test will fail anyways without it

raikonenfnu (commented):
Overall looks good, just need a bit of test as mentioned in comment

iree/turbine/kernel/compiler/host_codegen.py
harsh-nod: Sure, updated the lit test.

raikonenfnu (approved):

PR #491, opened 2025-02-11
[Wave] Get extend working with more MFMA intrinsics

harsh-nod:
This PR adds support for the 32 x 32 x 8 MFMA intrinsic in extend attention.

### DIFF
--- a/iree/turbine/kernel/wave/utils.py
+++ b/iree/turbine/kernel/wave/utils.py
@@ -389,11 +389,11 @@ def get_mma_dimensional_mapping(
             arg = get_custom(arg)
             if is_reshape_needed(arg, mma.vector_shapes, prev_mma.vector_shapes):
                 reshape = Reshape(arg.fx_node, prev_mma.vector_shapes).add_to_graph(
-                    custom.graph
+                    mma.graph
                 )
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/wave/utils.py
```diff
@@ -389,11 +389,11 @@ def add_reshape_if_needed(mma: MMA, prev_mma: MMA, arg_index: int):
             arg = get_custom(arg)
             if is_reshape_needed(arg, mma.vector_shapes, prev_mma.vector_shapes):
                 reshape = Reshape(arg.fx_node, prev_mma.vector_shapes).add_to_graph(
-                    custom.graph
+                    mma.graph
                 )
                 custom_reshape = get_custom(reshape)
-                custom_reshape.vector_shapes = custom.vector_shapes
-                custom.update_arg(arg_index, reshape)
```
raikonenfnu, line 396: was this a bug?

harsh-nod: In this context, yes. It was using custom which is defined in the outside function which was not the same as mma.

raikonenfnu (approved):

PR #486, opened 2025-02-10
Reland Attention with Linear Biases

ftynse:
This implements (causal) ALiBi, attention with linear biases, following
the paper "Train Short, Test Long: Attention with Linear Biases Enables
Input Length Extrapolation" by Press et.al in ICLR 2022.

Incidentally fixes API misuse in code generation for broadcast.

Reland #467

PR #497, opened 2025-02-12
Add RPE variant

nicolasvasilache:
As suggested by @hardcode84, when indirect indexings bottom out into load/store, we can more easily use IndexMapping with complex sympy expressions.

Signed-off-by: Nicolas Vasilache <nicolasvasilache@users.noreply.github.com>

### DIFF
--- a/iree/turbine/kernel/wave/codegen/emitter.py
+++ b/iree/turbine/kernel/wave/codegen/emitter.py
@@ -197,8 +197,7 @@ _emulate_ceildiv = True
 _Rational = namedtuple("_Rational", ["numerator", "denominator"])
 
 
-def gen_sympy_index(dynamics: dict[IndexSymbol, Value],
-                    expr: sympy.Expr) -> OpResult:
+def gen_sympy_index(dynamics: dict[IndexSymbol, Value], expr: sympy.Expr) -> OpResult:
     stack: list[OpResult] = []
 
     def _get_ir_value(arg):
@@ -260,9 +259,9 @@ def gen_sympy_index(dynamics: dict[IndexSymbol, Value],
             # If rhs_val is negative, since it is of index type and we have
             # overflow flags, we need to express it as a sub to get the
             # correct result.
-            return arith_d.subi(lhs,
-                                _get_const(-rhs_val),
-                                overflow_flags=overflow_flags)
+            return arith_d.subi(
+                lhs, _get_const(-rhs_val), overflow_flags=overflow_flags
+            )
 
         return arith_d.addi(lhs, rhs, overflow_flags=overflow_flags)
 
@@ -306,8 +305,7 @@ def gen_sympy_index(dynamics: dict[IndexSymbol, Value],
 
     def _floor(value):
         if isinstance(value, _Rational):
-            value = arith_d.divsi(
-                *_broadcast(value.numerator, value.denominator))
+            value = arith_d.divsi(*_broadcast(value.numerator, value.denominator))
 
         return value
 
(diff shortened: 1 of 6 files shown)


iree/turbine/kernel/wave/codegen/emitter.py
```diff
@@ -518,6 +525,9 @@ def _get_const(val):
                 cond = select_stack.pop()
                 last_expr = select_stack.pop()
                 last_cond = select_stack.pop()
+                # TODO: this is a hack, need a proper expansion of constant
+                # literal to vector.
+                expr = vector_d.splat(last_expr.type, expr)
                 res = arith_d.select(last_cond, last_expr, expr)
```
Hardcode84, line 531: `res = arith_d.select(last_cond, *_broadcast(last_expr, expr))`

harsh-nod:
Could you move to templates and add an e2e test like Alex's PR here https://github.com/iree-org/iree-turbine/pull/486/files ?

playground/attention_with_rpe_template.py
```diff
@@ -0,0 +1,221 @@
...
+    else:
+        rpe_layout = tkl.MemoryLayout(shape=[1, ])
+    # yapf: enable
+
+    @tkw.wave(constraints)
+    def base_attention(
+        q: tkl.Memory[B, M, K1, GLOBAL_ADDRESS_SPACE, tkl.f16],
+        k: tkl.Memory[B, K2, K1, ADDRESS_SPACE, tkl.f16],
+        v: tkl.Memory[B, N, K2, ADDRESS_SPACE, tkl.f16],
+        # TODO: if not use_t5_rpe, this will DCE; atm DCE on blockargs crashes.
```
harsh-nod, line 118: Since this is going to be t5_rpe template can you remove the use_t5_rpe in the entire template?

playground/attention_with_rpe_template.py
```diff
@@ -0,0 +1,221 @@
...
+
+                # yapf: disable
+                rpe_reg = tkw.read(
+                    rpe,
+                    mapping=offset_mapping,
+                    mapping_dynamic_vals=( i, j, ),
+                    elements_per_thread=LOAD_ELEMS_PER_THREAD_QK,
+                )
+                # yapf: enable
+                tkw.write(rpe_reg,
```
harsh-nod, line 161: Is this write only for debugging purposes? If so, can you remove?

playground/attention_with_rpe_test.py
```diff
@@ -0,0 +1,218 @@
...
+#     )
+# )
+
+
+#################################################################################
+# INIT VALS
+#################################################################################
+# num_query_heads, num_kv_heads, head_size, head_size_kv
+# WARNING: for now 128^n is the minimal size because attention_with_rpe_template
+# does not support padding and as a consequence we have
```
harsh-nod, line 67: Do you mean padding on the RPE vector? Should we add a mask on RPE ?

playground/attention_with_rpe_template.py
nicolasvasilache: ok, but note that it was by using this template and setting use_t5_rpe to false that I realized the vanilla version was incorrect on 32^n (due to no padding/masking support).
This is what prompted me to go to 128^n after looking at the WaveConstraints.

playground/attention_with_rpe_template.py
nicolasvasilache: ok, but note I used it to test the RPE cond matrix itself is consistent, which let me find the missing log2e factor.

playground/attention_with_rpe_test.py
nicolasvasilache: the vanilla attention part (without RPE) did not agree with torch on sizes 32^n; I suspect this is due to the workgroup / wave constraints that use 128 and 64.
This was what generated those masked stores we talked about earlier; I think (not sure though) this could be a miscompile?

nicolasvasilache:
@harsh-nod I see there was some failure that I am not able to repro locally.
Can we rerun to see if transient? 
If so that could suggest some potentially tricky synchronization issues.

Issue #419, opened 2025-01-27
[TKW] Bug: different return type when cached


GMNGeoffrey:
The combination of `WAVE_CACHE_ON=1` (the default) and `WAVE_DUMP_MLIR=1` fails (AttributeError: 'str' object has no attribute 'get_asm') when there's a cache hit because the cache `module_op` is just a string instead of an actual module op:

https://github.com/iree-org/iree-turbine/blob/bca982a7776ec13be73a2f6627fb129f47b6d2a4/iree/turbine/kernel/wave/cache.py#L50-L55

The return type for cached and uncached results should be the same

PR #500, opened 2025-02-13
WaveCache.module_op already returns the MLIR module str

tyb0807:


GMNGeoffrey:
I believe this is only true when the compilation is cached. See https://github.com/iree-org/iree-turbine/issues/419 (unless the non-cache path has changed since then)

GMNGeoffrey:
#419 was fixed by #574, so I think this is obsolete.

PR #504, opened 2025-02-14
Extend attention + RPE

nicolasvasilache:
This PR starts from an Extend attention + RPE by @raikonenfnu.

Then, avoid flex_attention for RPE as it is unclear there is a correct implementation possible given the
limitations of the create_block_mask re conditionals.

Instead we use the manual torch implementation that is known to be correct.

After debugging and further changes, the RPE implementation works as expected.

Signed-off-by: Nicolas Vasilache <nicolasvasilache@users.noreply.github.com>

### DIFF
--- a/iree/turbine/kernel/wave/templates/extend_attention_rpe.py
+++ b/iree/turbine/kernel/wave/templates/extend_attention_rpe.py
@@ -207,6 +207,7 @@ def get_extend_attention_rpe_kernel(
         ],
         rpe: tkl.Memory[N_KV, GLOBAL_ADDRESS_SPACE, tkl.f32, rpe_layout],
         c: tkl.Memory[N_Q, H, D_KV, GLOBAL_ADDRESS_SPACE, wave_output_dtype, o_layout],
+        rpe_debug: tkl.Memory[H, N_Q, N_KV, GLOBAL_ADDRESS_SPACE, tkl.f32],
     ):
         c_reg = tkl.Register[H, D_KV, N_Q, tkl.f32](0.0)
         init_sum = tkl.Register[H, N_Q, tkl.f32](0.0)
@@ -273,6 +274,11 @@ def get_extend_attention_rpe_kernel(
                 rpe,
                 mapping=rpe_mapping,
                 mapping_dynamic_vals=(i, j),
+                elements_per_thread=LOAD_ELEMS_PER_THREAD_QK,
+            )
+            tkw.write(
+                rpe_reg,
+                rpe_debug,
                 elements_per_thread=LOAD_ELEMS_PER_THREAD_QK,
             )
             # Layer and RPE scaling since we use log2 instead of log2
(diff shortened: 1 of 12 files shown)


tests/kernel/wave/attention/extend_attention_test.py
```diff
@@ -364,11 +409,153 @@ def testExtendAttention(
...
         is_causal=is_causal,
         logit_cap=logit_cap,
     )
 
     assert_allclose(output, ref_output, rtol=1e-3, atol=1e-3)
+
+
+# TODO: Investigate errors on MI250.
+@require_e2e
+# @require_cdna3
```
Hardcode84, line 424: uncomment

Hardcode84 (commented):
Seems non-rpe extend is broken `FAILED tests/kernel/wave/attention/extend_attention_test.py::testExtendAttention[mfma_variant0-False-False-False-dtype0-16x1x128x128x2x1024x64] - AttributeError: 'NoneType' object has no attribute 'device'`

Regarding 0.5% mismatched elements, I think we can increase tolerance for now for rpe to land it and investigate later.

iree/turbine/kernel/wave/templates/extend_attention_rpe.py
```diff
@@ -0,0 +1,387 @@
...
+            j = tkw.self_index(
+                N_KV, tkl.i64, elements_per_thread=LOAD_ELEMS_PER_THREAD_QK
+            )
+            rpe_reg = tkw.read(
+                rpe,
+                mapping=rpe_mapping,
+                mapping_dynamic_vals=(i, j),
+                elements_per_thread=LOAD_ELEMS_PER_THREAD_QK,
+            )
+            # Layer and RPE scaling since we use log2 instead of log2
```
raikonenfnu, line 326: Do you mean `instead of loge` here?

tests/kernel/wave/attention/extend_attention_test.py
```diff
@@ -229,12 +270,14 @@ def create_inputs(
...
         logit_cap,
+        rpe_bias,
+        max_rpe_context_length,
     )
 
 
 # TODO: Investigate errors on MI250.
 @require_e2e
-@require_cdna3
+# @require_cdna3
```
raikonenfnu, line 280: uncomment or delete if it is indeed working on mi250?

Hardcode84 (approved):

qcolombet:
@nicolasvasilache What was the issue in the end with the RPE tensor?
I don't see any substantial differences between what you sent here and what you sent me a couple of days for debugging.
Hence, I'm curious what was wrong and what fixed it.

qcolombet:
> @nicolasvasilache What was the issue in the end with the RPE tensor? I don't see any substantial differences between what you sent here and what you sent me a couple of days for debugging. Hence, I'm curious what was wrong and what fixed it.

@nicolasvasilache ping?

Issue #278, opened 2024-11-19
[Wave] Fun projects for beginnners


raikonenfnu:
Are you interested in learning more about GPU programming and developing cool optimizations? Do you want to help build next generation and state-of-the-art machine learning models and layers? Do you want to define the future programming paradigm of machine learning and GPU layers? Look no further, come join us in building "Wave"!

Here are some fun starter tasks to look at:

- [ ] Implement backwards attention kernel on TKW
- [ ] Cache kernels
- [ ] Scalar support
- [ ] Auto detection of hardware architecture in runtime
- [ ] Support more architecture in codegen aside from CDNA.

NoumanAmir657:
Would this be open for someone who has almost no knowledge of GPU programming except for some basic understanding.

raikonenfnu:
Hi @NoumanAmir657 thanks for reaching out! There are plenty of tasks to go around, depending on your experience level, would probably assign easier tasks to start with.

NoumanAmir657:
> Hi @NoumanAmir657 thanks for reaching out! There are plenty of tasks to go around, depending on your experience level, would probably assign easier tasks to start with.

Hi, yes would love any pointers to which task would help me getting started. You can assign me whichever you see fit and I can work towards it

raikonenfnu:
@NoumanAmir657 I think a good starter task is to implement `tkw.abs` and it's lowering into `math.absfOp` or `math.absIOp` in math dialect of MLIR. Here is a sample of a similar task on reciprocal (https://github.com/iree-org/iree-turbine/commit/82852d13198be9f18f6a070b2762bdaf6e78c5f3).

LMK what you think. :)

Disclaimer: if this(or any task assigned to external collaborator) becomes high priority, someone internal may take on this task to get it out asap.

NoumanAmir657:
> @NoumanAmir657 I think a good starter task is to implement `tkw.abs` and it's lowering into `math.absfOp` or `math.absIOp` in math dialect of MLIR. Here is a sample of a similar task on reciprocal (https://github.com/iree-org/iree-turbine/commit/82852d13198be9f18f6a070b2762bdaf6e78c5f3).
> 
> LMK what you think. :)
> 
> Disclaimer: if this(or any task assigned to external collaborator) becomes high priority, someone internal may take on this task to get it out asap.

Yes this seems a good starting point. I shall make a PR soon. Thanks!

raikonenfnu:
[...]

PR #507, opened 2025-02-15
[TKW] Add MinOp as an Op for Quantized LLM and GenAI workload

NoumanAmir657:
The TKW::MinOp is lowered using TKW::MinimumOp into arith::MinimumFOp, arith::MinisIOp, and arith::MiniuIOp for floats, signed and unsigned integers respectively

For #278


tests/kernel/wave/test_param.json
```diff
@@ -40,6 +40,12 @@
         [256, 256],
         [256, 1024]
     ],
+    "test_tiled_reduce_min": [
```
raikonenfnu, line 43: I think this is only required for e2e test. We can forgo this part for now

raikonenfnu (commented):
Thanks for this, overall looks good, aside from the redundant size params

tests/kernel/wave/test_param.json
NoumanAmir657: Fixed.

raikonenfnu (approved):
Thanks for the awesome contribution! :)

PR #511, opened 2025-02-17
[TKW] Fix RPE 32x32x8 MMA

Hardcode84:
* Add missing `subs_idxc` in `has_gpr_offsets`
* Handle `mapping_dynamic_vals` for read/write ops in `has_gpr_offsets`
* Propagate index to generated `Reshape` op, which needed for proper detection dynamic vals handling
* Use `STORE_ELEMS_PER_THREAD` for RPE read in RPE kernel as the result is combined with mma result
* Remove `elements_per_thread` from `self_index` as they can be inferred

### DIFF
--- a/iree/turbine/kernel/wave/index_sequence_analysis.py
+++ b/iree/turbine/kernel/wave/index_sequence_analysis.py
@@ -261,6 +261,11 @@ def partition_ops_with_gpr_offsets(trace: CapturedTrace, constraints: list[Const
                 ] = updated_dim_with_gpr_offset
 
                 if hasattr(custom, "mapping_dynamic_vals"):
+                    # If we are partitioning read/write ops, dynamic_vals can be
+                    # potentially partitioned as well. Partitioned dyn vals are
+                    # are merged into single value using Reshape op which still
+                    # holds the original index containing `GPR_NUM`.
+                    # Extract corresponding partitioned chunk from such ops.
                     new_dynamic_vals = []
                     for dyn_val in custom.mapping_dynamic_vals:
                         if any(
@@ -319,6 +324,9 @@ def partition_ops_with_gpr_offsets(trace: CapturedTrace, constraints: list[Const
                 )
                 reshape.expanded_dims = custom.expanded_dims
                 reshape.vector_shapes = custom.vector_shapes
+
+                # Save the original index on the reshape op so later we can
+                # detect if op was part of `gpr_offset` partition.
                 reshape.index = custom.index
                 custom.replace_all_uses_with(reshape)
 
--- a/iree/turbine/kernel/wave/index_sequence_analysis.py
+++ b/iree/turbine/kernel/wave/index_sequence_analysis.py
@@ -284,6 +284,7 @@ def partition_ops_with_gpr_offsets(trace: CapturedTrace, constraints: list[Const
                         extract,
                         custom.memory,
                         mapping=custom.mapping,
+                        mapping_dynamic_vals=new_dynamic_vals,
                         elements_per_thread=gpr_size,
                     ).add_to_graph(custom.graph)
                 elif isinstance(custom, Read):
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -260,6 +260,20 @@ def has_gpr_offsets(node: fx.Node) -> bool:
...
                 ] = updated_dim_with_gpr_offset
 
+                if hasattr(custom, "mapping_dynamic_vals"):
+                    new_dynamic_vals = []
+                    for dyn_val in custom.mapping_dynamic_vals:
+                        if any(
+                            sympy.sympify(v.start).has(GPR_NUM)
+                            for v in get_custom(dyn_val).index.values()
+                        ):
+                            extract = ExtractSlice(
```
raikonenfnu, line 270: Just curious, do we need to actually extract_slice on the dyn_vals? what example would it fail on if we just pass in dynamic vals directly without extract?

iree/turbine/kernel/wave/index_sequence_analysis.py
```diff
@@ -303,7 +319,9 @@ def has_gpr_offsets(node: fx.Node) -> bool:
                 )
                 reshape.expanded_dims = custom.expanded_dims
                 reshape.vector_shapes = custom.vector_shapes
+                reshape.index = custom.index
```
raikonenfnu, line 322: Just curious, in what case would we need this?

iree/turbine/kernel/wave/index_sequence_analysis.py
Hardcode84: `self_index` with `GPR_NUM` gets partitioned 4x4 ops which are then reshaped into 16 vec. Read op which is using it is also partitioned into 4x4, so we need to extract 4-sized chunks back.

iree/turbine/kernel/wave/index_sequence_analysis.py
Hardcode84: I'm using index to determine if dynamic val was partitioned or not: https://github.com/iree-org/iree-turbine/pull/511/files#diff-42e6ce65d001220b0098fcfb15f4c93f181a3f53d461b2ce97c2d27d254b5cb4R267

This is ugly, but I didn't found a better way.

iree/turbine/kernel/wave/index_sequence_analysis.py
raikonenfnu: the ReadOp being the Read in charge of reading the rpe bias?

raikonenfnu: Can we add some comments just explaining this on the code actually? :)

Hardcode84: added some comments

raikonenfnu: thanks!

iree/turbine/kernel/wave/index_sequence_analysis.py
raikonenfnu: > detect if op was part of `gpr_offset` partition
Just curious, how/where are we doing this actual check

Hardcode84:
```
                        if any(
                            sympy.sympify(v.start).has(GPR_NUM)
                            for v in get_custom(dyn_val).index.values()
                        ):
```

raikonenfnu: Doesn't this check come before the `reshape.index = custom.index` ?

Hardcode84: This check is for there read op, but `reshape.index = custom.index` will happen for the `self_index` ops before the read.

raikonenfnu: Ahh I see makes sense, thanks!

raikonenfnu (approved):

PR #515, opened 2025-02-18
[Wave] minimize global load for dynamics

raikonenfnu:
Add support for minimizing global loads for cases that has dynamic vals offsets in it's mapping. This is especially useful to optimize for cases such as paged attention, where we use dynamic vals as an offset for k_cache reads.

Summary of changes:

-  Move `_get_fastest_index` to utils for reuse in other stack of the code
- Add some util functions to check if contiguous reads in `is_transposed_read` and another util function to copy/set node's index to a desired/target index expression in `update_index_dims`
- Modify `minimize_global_loads.py`'s identifying optimizable loads to also let in **non**-transposed Read who has dynamic offset while keeping track of the set of vals said memory use as dynamic offset.
- Add code to handle **non**-transposed Read who has dynamic offset by updating the dynamic offset val's index to match minimum_global_access_pattern and by pointing the newly generated Read to use said value as an dynamic offset value.
- Keep track of load_elems_per_thread, max_elements_per_load per custom.memory instead of always using a global optimized one. This is most useful for cases where we cannot use widest read (i.e 8xf16) but can still eliminate redundant reads with (two 4xf16). For example K_cache with workgroup/shared memory tile size of (32x64xf16) expects single read, but in reality has two offsets => we need to break apart the reads.


iree/turbine/kernel/wave/utils.py
```diff
@@ -1235,6 +1235,20 @@ def _get_start_indices(
     return start_indices
 
 
+def _get_fastest_index(indices: dict[IndexExpr, IndexSequence]):
```
Hardcode84, line 1238: nit: as this is now a public function, we can remove underscore from name.

raikonenfnu: sg!

raikonenfnu: thanks for the catch, done! :)

Hardcode84 (approved):

PR #519, opened 2025-02-19
[Wave] Remove polynomial approx with iree compile flag

raikonenfnu:
Polynomial approximation causes major slowdown, so we add a IREE compile flag to remove it. The changes wouldn't be visible at Wave level but will give good perf boost in generated LLVMIR

### DIFF
--- a/iree/turbine/kernel/wave/utils.py
+++ b/iree/turbine/kernel/wave/utils.py
@@ -609,6 +609,9 @@ def compile_to_vmfb(
     if backend == "rocm":
         target = config["target"]
         flags.append(f"--iree-hip-target={target}")
+        # Polynomial approximation passes in MLIR/IREE often generate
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/wave/utils.py
```diff
@@ -609,6 +609,7 @@ def compile_to_vmfb(
     if backend == "rocm":
         target = config["target"]
         flags.append(f"--iree-hip-target={target}")
+        flags.append("--iree-codegen-gpu-native-math-precision=true")
```
Hardcode84, line 612: Some comment would be nice )

Hardcode84 (approved):

iree/turbine/kernel/wave/utils.py
raikonenfnu: makes sense, done! :)

PR #522, opened 2025-02-20
[TKW] Fix memory leak

Hardcode84:
We are converting kernel args to dlpack capsule to feed them into IREE runtime, unfortunately, IREE runtime renames all incoming capsules from "dltensor" to "dltensor_used" for some reason and also will never delete capsule if its name not a "dltensor". This causes incoming tensors lifetime to be prolonged indefinitely and to be never released.

Rename capsule back after `device.from_dlpack_capsule`


raikonenfnu (approved):
This is very awesome! :)

PR #526, opened 2025-02-21
[TKW] Use buffer ops for splatted masked load/stores

Hardcode84:
Add a special case when masked load/store mask is result of `vector.splat`. In this case we don't need to unroll the op and instead can use vector buffer ops directly.

### DIFF
--- a/iree/turbine/kernel/wave/codegen/read_write.py
+++ b/iree/turbine/kernel/wave/codegen/read_write.py
@@ -445,7 +445,12 @@ def _create_vec_read(
             )
 
         if splatted_masked:
-            offset_th = arith_d.index_cast(IntegerType.get_signless(32), offset_th)
+            i32 = IntegerType.get_signless(32)
+            offset_th = arith_d.index_cast(i32, offset_th)
+            oob_idx = _get_max_buffer_size(element_type)
+            oob_idx = arith_d.constant(i32, oob_idx)
+            offset_th = arith_d.select(mask_splat, offset_th, oob_idx)
+
             load_type = vector_type
             if elements_per_thread == 1:
                 load_type = element_type
@@ -629,7 +634,12 @@ def _create_vec_write(
         )
 
         if splatted_masked:
-            offset_th = arith_d.index_cast(IntegerType.get_signless(32), offset_th)
+            i32 = IntegerType.get_signless(32)
+            offset_th = arith_d.index_cast(i32, offset_th)
+            oob_idx = _get_max_buffer_size(element_type)
+            oob_idx = arith_d.constant(i32, oob_idx)
+            offset_th = arith_d.select(mask_splat, offset_th, oob_idx)
+
             if elements_per_thread == 1:
                 value = vector_d.extract(
                     value, static_position=[0], dynamic_position=[]
--- a/iree/turbine/kernel/wave/codegen/read_write.py
+++ b/iree/turbine/kernel/wave/codegen/read_write.py
@@ -385,10 +385,11 @@ def _get_splat_input(src: Optional[Value]) -> Optional[Value]:
     if src is None:
         return None
 
-    try:
-        op = src.owner.opview
-    except:
+    owner = getattr(src, "owner", None)
+    if owner is None:
         return None
+
+    op = src.owner.opview
     if isinstance(op, vector_d.SplatOp):
         return op.input
 
--- a/iree/turbine/kernel/wave/codegen/read_write.py
+++ b/iree/turbine/kernel/wave/codegen/read_write.py
@@ -417,11 +417,12 @@ def _create_buffer_read_write(
         return None
 
 
-def _create_vec_read(
+def _create_vec_read_write(
     emitter: WaveEmitter,
     symbolic_shape: tuple[IndexExpr, ...],
     mem: Value,
-    vector_type: IrType,
+    value: Optional[Value],
+    vector_type: Optional[IrType],
     start_indices: tuple[Value],
     start_indices_wg: tuple[Value],
     start_indices_th: tuple[Value],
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -77,20 +77,21 @@ def _split_index(src: IndexExpr | int) -> tuple[IndexExpr, IndexExpr]:
...
-    # All thread symbols should cancel-out in the result, but to be sure
-    # replace all thread symbols by 0 in the result.
-    # We cannot just replace all thread symbols without the subtraction as
-    # any constant or dynamic values will end up in both expressions.
-    thread_indepdndent_index = sympy.simplify(
-        safe_subs(src - thread_dependend_index, subs_th)
-    )
-    return thread_indepdndent_index, thread_dependend_index
+    # All thread symbols and dynamic should cancel-out in the result.
+    thread_independent_index = sympy.simplify(src - thread_dependend_index)
```
harsh-nod, line 86: minor typo: thread_dependent_index here and below

iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -77,20 +77,21 @@ def _split_index(src: IndexExpr | int) -> tuple[IndexExpr, IndexExpr]:
...
+    if thread_independent_index.free_symbols - set(subs_wg.keys()):
+        # If we have any symbols besides wg symbols, means some thread or
+        # dynamic symbols were not canceled out, use the entire index as
+        # thread dependent index.
+        thread_independent_index = sympy.sympify(0)
```
harsh-nod, line 91: Why sympy.sympify(0) instead of sympy.Integer(0)?

iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -377,6 +378,23 @@ def _linearize_memref(
...
 
 
+def _get_splat_input(src: Optional[Value]) -> Optional[Value]:
+    """
+    If `src` is vector.splat result, return splat input, otherwise return None.
+    """
+    if src is None:
+        return None
+
+    try:
```
harsh-nod, line 388: Since this is not doing much in the except branch, how about replacing with getattr?

iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -421,33 +438,75 @@ def _create_vec_read(
...
+            res = amdgpu_d.raw_buffer_load(load_type, data, indices=[offset_th])
+            if elements_per_thread == 1:
+                res = vector_d.splat(vector_type, res)
 
-        return result
+            return res
+        else:
+            # If mask value is different for each element, unroll op to
+            # individual values.
+            offset_th = vector_d.splat(offsets_vec.type, offset_th)
```
harsh-nod, line 473: These functions are almost identical for read/write, can we refactor them into a common function?

iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -583,28 +641,67 @@ def _create_vec_write(
         data, offset_th = _linearize_memref(
             mem, start_indices_wg, start_indices_th, strides
         )
-        offset_th = vector_d.splat(offsets_vec.type, offset_th)
-        offsets_vec = arith_d.addi(offsets_vec, offset_th)
-        if mask is not None:
+
+        if splatted_masked:
```
harsh-nod, line 645: The comment below says we can use vector buffer ops but I don't see any vector buffer ops. What did you mean by vector buffer ops?

iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -583,28 +641,67 @@ def _create_vec_write(
...
-        for i in range(elements_per_thread):
-            offset = vector_d.extract(
-                offsets_vec, static_position=[i], dynamic_position=[]
-            )
-            elem = vector_d.extract(value, static_position=[i], dynamic_position=[])
-            if mask is None:
-                memref_d.store(elem, data, indices=[offset])
-            else:
-                amdgpu_d.raw_buffer_store(elem, data, indices=[offset])
+            # Buffer ops doesn't support 1-element vectors, convert to scalar.
```
harsh-nod, line 654: Is this an LLVM/MLIR bug? If so can you mark it as a TODO for someone to upstream the fix?

harsh-nod (changes requested):
Overall looks good, mostly requests for refactoring changes and some explanation of comments.

iree/turbine/kernel/wave/codegen/read_write.py
Hardcode84: Upstream fIx is merged https://github.com/llvm/llvm-project/pull/128266, we need to wait until IREE update

iree/turbine/kernel/wave/codegen/read_write.py
Hardcode84: No particular reason, they are equivalent

iree/turbine/kernel/wave/codegen/read_write.py
Hardcode84: done

iree/turbine/kernel/wave/codegen/read_write.py
Hardcode84: done

iree/turbine/kernel/wave/codegen/read_write.py
Hardcode84: That means we can pass vector values directly to `raw_buffer_store`/ use vector `load_type` for `raw_buffer_load`

iree/turbine/kernel/wave/codegen/read_write.py
harsh-nod: thanks

iree/turbine/kernel/wave/codegen/read_write.py
harsh-nod: nice

iree/turbine/kernel/wave/codegen/read_write.py
harsh-nod: makes sense, thanks

PR #530, opened 2025-02-25
[TKW] Extend attention RPE `F32_32x32x8_F16`

Hardcode84:
Need to use `STORE_ELEMS_PER_THREAD` for RPE as we are applying RPE to mma accumulator.


harsh-nod (approved):
nice!

PR #531, opened 2025-02-26
Improvements to caching

harsh-nod:
This PR improves the way caching was done. Rather than use functools.lru_cache, users can now pass in a kernel_hash argument which will return the kernel hash for the kernel. This can then be reused in subsequent calls to avoid the hash computation.

Furthermore, we add a runtime cache to cache the system context and function to lower overheads.

### DIFF
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -59,7 +59,6 @@ from .utils import (
     canonicalize_module,
     compile_to_vmfb,
     invoke_vmfb,
-    safe_dict_to_tuple,
     safe_subs,
     remove_chained_getresult,
     remove_chained_extractslice,
@@ -74,7 +73,6 @@ from .cache import (
     is_cache_enabled,
     get_cache_manager,
     invoke_cached_kernel,
-    anonymize_constraints,
 )
 
 # Others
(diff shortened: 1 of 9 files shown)


iree/turbine/kernel/wave/wave.py
```diff
@@ -521,23 +519,30 @@ def test_execute(self, args, kwargs):
...
-                self._f,
-                safe_dict_to_tuple(IndexingContext.current().subs),
-                tuple(dynamic_symbols),
-                safe_dict_to_tuple(config),
-                use_scheduling=use_scheduling,
-                use_scheduling_barriers=use_scheduling_barriers,
-                run_bench=run_bench,
-            )
+            if cached_kernel_hash:
+                kernel_hash = cached_kernel_hash[0]
```
raikonenfnu, line 532: I don't understand why are we always reutrning the first index in cached_kernel_hash? Shouldn't we do some index check or something?

tests/kernel/wave/attention/extend_attention_test.py
```diff
@@ -464,7 +465,7 @@ def testExtendRpeAttention(
         rpe_bias,
         max_rpe_context_length,
     ) = create_inputs(shape, dtype)
-    shape.max_seq_len = max_len_extend
```
raikonenfnu, line 467: What does this change do?

harsh-nod: I wanted to make attentionshape hashable and so now that it is hashable, you can only update the value by making a copy and dataclass has a convenient function replace that lets you copy a dataclass and only replace specified fields.

iree/turbine/kernel/wave/wave.py
harsh-nod: So the idea is that when you use this flag, you are confident that you have only one kernel it should be cached and thats why I only load the first element. But open to suggestions

harsh-nod: I have modified this so that the length of cached_kernel_hash will always be 1.

iree/turbine/kernel/wave/wave.py
```diff
@@ -521,23 +519,33 @@ def test_execute(self, args, kwargs):
...
-                safe_dict_to_tuple(config),
-                use_scheduling=use_scheduling,
-                use_scheduling_barriers=use_scheduling_barriers,
-                run_bench=run_bench,
-            )
+            cache_manager = get_cache_manager()
+            if cached_kernel_hash:
+                # If a cached_kernel_hash is passed in, we assume it was generated in a previous run
+                # and since we always return a list of size 1, we can just grab the first element.
+                kernel_hash = cached_kernel_hash[0]
```
raikonenfnu, line 535: just double checking, so it wouldn't be possible for previous `kernel_hash` to != `cached_kernel_hash[0]`?

raikonenfnu (approved):
Overall LGTM! just a QQ

iree/turbine/kernel/wave/wave.py
harsh-nod: That is possible. Essentially what this new arg does is allow the user to see what the kernel hash is for a given kernel. The user could actually run 5 different kernels and get their hashes and then decide which kernel they want to run by choosing the appropriate hash and feeding it back in. So its not always guaranteed to be the previous one. Since it is passed in from the user, it could also contain more than one hash. But we only use the first one because that is consistent with what we passed back to the user.

PR #532, opened 2025-02-26
[NFC] Refactor to allow more composable usage of LaunchableWave towar…

nicolasvasilache:
…ds more comprehensive tests

### DIFF
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -355,7 +355,7 @@ class LaunchableWave(Launchable):
             print(f"***After {p.__name__}***\n")
             print_trace(trace)
 
-    def compile_to_scf_dialect(
+    def compile_to_mlir(
         self,
         trace: CapturedTrace,
         context: Context,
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/wave/wave.py
```diff
@@ -321,6 +321,130 @@ def initialize_symbolic_constraints(self, trace: CapturedTrace) -> None:
...
+            p()
+        except Exception:
+            print(f"Error in pass: {p.__name__}\n")
+            print_trace(trace)
+            raise
+        if "all" in print_ir_after or p.__name__ in print_ir_after:
+            print(f"***After {p.__name__}***\n")
+            print_trace(trace)
+
+    def compile_to_scf_dialect(
```
Hardcode84, line 360: Not sure if `compile_to_scf_dialect` is a good name, probably just `compile_to_mlir`

Hardcode84 (approved):

Hardcode84 (approved):

iree/turbine/kernel/wave/wave.py
Hardcode84: You reverted it back with last rebase I think

nicolasvasilache: you're right, I force pushed to get rid of the merges that I introduced clicking on the UI, fixing ..

PR #537, opened 2025-02-27
[NFC] Move WaveConstraint related convention into the WaveConstraint …

nicolasvasilache:
…class


Hardcode84:
Some tests are flacky and can be ignored:
```
FAILED tests/dynamo/tensor_test.py::TensorTest::test_nn_linear - RuntimeError: allocating memory on queue: ABORTED
FAILED tests/dynamo/tensor_test.py::TensorTest::test_unary_op - RuntimeError: allocating memory on queue: ABORTED
```
But some seems to be broken for real
```
FAILED tests/kernel/wave/runtime/cache_test.py::testSameSizeDifferentBlock - AssertionError: Wave id already set
```

PR #541, opened 2025-02-27
[TKW] Unroll gathers/scatters

Hardcode84:
* Manually unroll gathers/scatters into 1-element masked ops.

### DIFF
--- a/iree/turbine/kernel/wave/codegen/read_write.py
+++ b/iree/turbine/kernel/wave/codegen/read_write.py
@@ -585,7 +585,7 @@ def _create_vec_read_write(
                     mask_elem = extract(mask, i)
                     mask_elem = vector_d.splat(vec1_mask, mask_elem)
 
-                    offset = vector_d.extract(offsets_vec, i)
+                    offset = extract(offsets_vec, i)
 
                     elem = vector_d.maskedload(
                         vec1, data, [offset], mask_elem, passthru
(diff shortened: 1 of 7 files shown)


harsh-nod (approved):
Lgtm! Thanks!

nicolasvasilache (commented):
@Hardcode84 I have 1 comment and 1 question for this PR.

Comment: there seems to be a behavior change with this PR when one reads indirect (e.g. `A[B[i], ...] `), when `B[i]` indexes out of bounds. We seem to go from a `vector.gather` to an unconditional `vector.load`, which seems to not be conservative. This seems it would benefit from an annotation B to say that the values are known to be constrained to a symbolic range which would make the simplification conservative.
In https://github.com/iree-org/iree-turbine/pull/568 I played with a minimal load_indirect, if you want an example.

Question: Given the previous point, would it make sense to put most of TKW optimizations behind flags? (could be enabled by default). The rationale is that, like for canonicalization, it is useful to see -O0 IR generated before letting canonicalizations and then various optimizations kick in.

nicolasvasilache:
> There is no behavior change, this code generates 1-element masked load/stores with same mask as gather/scatter would be using. Masked ops can be converted to non-masked by vector canonicalizations if mask is constant, but such mask wouldn't have any effect on gather either.

Ah you're right, the original gather has a constant true mask, just does not get canonicalized in the absence of unrolling, sorry for the noise.

What should we do about generic `A[B[i]]` ? 
It seems we will miscompile if `B[i]` accesses out of bounds.
I am unclear that it is reasonable to assume "must be in-bounds" by default at the TK programming level ?

PR #549, opened 2025-02-28
[TKW] Warn if IREE versions is too low

Hardcode84:
Emit a warning if IREE version is too low. Warning is only emitted when we trying to compile kernel, so if users don't use Wave they won't see it. We also don't emit it if user uses a cached kernel to not add any additional overhead to 'happy' path.

### DIFF
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -89,6 +89,14 @@ __all__ = ["wave", "wave_trace_only"]
 _warned = False
 
 
+def _are_versions_compatible(ver1: "Version", ver2: "Version") -> bool:
+    if ver1.is_prerelease or ver2.is_prerelease:
+        return ver1 == ver2
+    else:
+        # For stable releases, it is fine if the patch level mismatches.
+        return (ver1.major == ver2.major) and (ver1.minor == ver2.minor)
+
+
 def _warn_iree_is_too_old():
     """
     Issue a warning if IREE runtime and compiler versions mismatch or IREE
@@ -110,7 +118,7 @@ def _warn_iree_is_too_old():
 
     iree_compiler_ver = Version(version("iree-base-compiler"))
     iree_runtime_ver = Version(version("iree-base-runtime"))
-    if iree_compiler_ver != iree_runtime_ver:
+    if not _are_versions_compatible(iree_compiler_ver, iree_runtime_ver):
         warnings.warn(
             f"IREE compiler and runtime versions mismatch: {iree_compiler_ver} and {iree_runtime_ver}"
         )
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -90,6 +90,12 @@ _warned = False
 
 
 def _warn_iree_is_too_old():
+    """
+    Issue a warning if IREE runtime and compiler versions mismatch or IREE
+    version is too low.
+
+    Warning is issued only once.
+    """
     global _warned
     if _warned:
         return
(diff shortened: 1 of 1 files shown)


marbre (changes requested):

ScottTodd (commented):
The shape of this LGTM. Agreed with the comment from Marius then good to merge IMO.

iree/turbine/kernel/wave/wave.py
```diff
@@ -80,10 +80,50 @@
...
+        from importlib.metadata import version
+    except ImportError:
+        return
+
+    iree_compiler_ver = Version(version("iree-base-compiler"))
+    iree_runtime_ver = Version(version("iree-base-runtime"))
+    if iree_compiler_ver != iree_runtime_ver:
+        warnings.warn(
+            f"IREE compiler and runtime versions mismatch: {iree_compiler_ver} and {iree_runtime_ver}"
+        )
```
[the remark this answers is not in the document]
Hardcode84: done

iree/turbine/kernel/wave/wave.py
```diff
@@ -446,6 +494,12 @@ def _trace_and_get_kernel_signature(
         kernel_codegen.KernelSignature,
         str,
     ]:
+        # Issue a warning if IREE ver is too low.
+        # Warning will only be issued if we are compileing the kernel and won't
```
marbre, line 498: NIT: Typo:
```suggestion
        # Warning will only be issued if we are compiling the kernel and won't
```

marbre (approved):

iree/turbine/kernel/wave/wave.py
```diff
@@ -80,10 +80,58 @@
...
+
+    _warned = True
+
+    try:
+        from packaging.version import Version
+        from importlib.metadata import version
+    except ImportError:
+        return
+
+    iree_compiler_ver = Version(version("iree-base-compiler"))
```
ScottTodd, line 119: Ah, this doesn't handle locally built development sources using PYTHONPATH instead of an installed packages: https://github.com/iree-org/iree-turbine?tab=readme-ov-file#configuring-python

```
iree\turbine\kernel\_support\tracing.py:427: in __call__
    return launch_context.launch(self, args, kwargs)
iree\turbine\kernel\_support\tracing.py:490: in launch
    return launchable.test_execute(args, kwargs)
iree\turbine\kernel\wave\wave.py:635: in test_execute
    ) = self._trace_and_get_kernel_signature(args, kwargs)
iree\turbine\kernel\wave\wave.py:493: in _trace_and_get_kernel_signature
    _warn_iree_is_too_old()
iree\turbine\kernel\wave\wave.py:120: in _warn_iree_is_too_old
    iree_compiler_ver = Version(version("iree-base-compiler"))
C:\Program Files\Python311\Lib\importlib\metadata\__init__.py:1008: in version
    return distribution(distribution_name).version
C:\Program Files\Python311\Lib\importlib\metadata\__init__.py:981: in distribution
    return Distribution.from_name(distribution_name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'importlib.metadata.Distribution'>, name = 'iree-base-compiler'

    @classmethod
    def from_name(cls, name: str):
        """Return the Distribution for the given package name.

        :param name: The name of the distribution package to search for.
        :return: The Distribution instance (or subclass thereof) for the named
            package, if found.
        :raises PackageNotFoundError: When the named package's distribution
            metadata cannot be found.
        :raises ValueError: When an invalid value is supplied for name.
        """
        if not name:
            raise ValueError("A distribution name is required.")
        try:
            return next(cls.discover(name=name))
        except StopIteration:
>           raise PackageNotFoundError(name)
E           importlib.metadata.PackageNotFoundError: No package metadata was found for iree-base-compiler

C:\Program Files\Python311\Lib\importlib\metadata\__init__.py:565: PackageNotFoundError
```

Logging a warning would be okay, or doing nothing, but throwing an Exception is not.

ScottTodd: (those logs are a snippet from output of `pytest -n auto` after following the instructions I linked)

Hardcode84: probably better to just wrap the entire fun into try/except

PR #564, opened 2025-03-06
added PERF and CI TK to ossci cluster

Eliasj42:
added PERF and CI TK to ossci cluster


.github/workflows/ci-tk.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
           echo "$VENV_DIR/bin" >> "$GITHUB_PATH"
 
       - name: Install pip deps
-        if: "!contains(matrix.os, 'amdgpu') && !cancelled()"
+        if: "!contains(matrix.os, 'gpu') && !cancelled()"
```
ScottTodd, line 52: If we ever want to run these tests on GPUs that aren't AMD GPUs, I'd prefer to use a more specific check here.

Please restructure the `matrix` above to be explicit about which types of tests should run on which 'os' (that should really be 'runner' or 'runs-on', not 'os'). Maybe something like
```yml
      matrix:
        include:
          - name: cpu
            runs-on: ubuntu-24.04
          - name: amdgpu_mi300
            runs-on: linux-mi300-1gpu-ossci-iree-org
          - name: amdgpu_mi250
            runs-on: nodai-amdgpu-mi250-x86-64
```

Then check `matrix.name` below using `amdgpu` or `mi300`/`mi250`. Or you could get more explicit with a field for the enabled test types instead of encoding that information in the name.

.github/workflows/ci-tk.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: Run e2e tests on AMD GPU MI300
         if: "contains(matrix.os, 'mi300') && !cancelled()"
         run: |
-          WAVE_CACHE_ON=0 pytest -n 4 --capture=tee-sys -vv --run-e2e --gpu-distribute 8 ./tests/kernel/wave/
+          WAVE_CACHE_ON=0 pytest -n 4 --capture=tee-sys -vv --run-e2e --gpu-distribute 1 ./tests/kernel/wave/
```
ScottTodd, line 88: Comparing this to a prior run:

* Baseline: https://github.com/iree-org/iree-turbine/actions/runs/13708286757/job/38338837063
  * 1m20s installing pip deps, including
    ```
    Thu, 06 Mar 2025 21:18:21 GMT Collecting torch<2.6.0,>=2.3.0 (from -r pytorch-rocm-requirements.txt (line 2))
    Thu, 06 Mar 2025 21:18:21 GMT   Using cached https://download.pytorch.org/whl/rocm6.2/torch-2.5.1%2Brocm6.2-cp311-cp311-linux_x86_64.whl (3973.6 MB)
    ```
  * Tests took 24 minutes, distributing across 4 CPU workers and 8 GPUs?
* This PR: https://github.com/iree-org/iree-turbine/actions/runs/13710789914/job/38346685232?pr=564
  * 12m installing pip deps, including
    ```
    Fri, 07 Mar 2025 00:13:14 GMT Collecting torch<2.6.0,>=2.3.0 (from -r pytorch-rocm-requirements.txt (line 2))
    Fri, 07 Mar 2025 00:13:14 GMT   Downloading https://download.pytorch.org/whl/rocm6.2/torch-2.5.1%2Brocm6.2-cp311-cp311-linux_x86_64.whl (3973.6 MB)
    Fri, 07 Mar 2025 00:23:46 GMT      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 4.0/4.0 GB 6.8 MB/s eta 0:00:00
    ```
  * Tests took 20 minutes, distributing across 4 CPU workers and 1 GPU?

So the GPU distribution seems fine. Will those pip packages be cached on the cluster runners for future runs?

@raikonenfnu switched this workflow to use the rocm torch wheels instead of cpu wheels in https://github.com/iree-org/iree-turbine/pull/276. I assume that's still needed for some native PyTorch / eager mode usage?

.github/workflows/ci-tk.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         version: [3.11]
-        os: [ubuntu-22.04, nodai-amdgpu-mi300-x86-64, nodai-amdgpu-mi250-x86-64]
+        os: [ubuntu-22.04, linux-mi300-1gpu-ossci-iree-org, nodai-amdgpu-mi250-x86-64]
```
ScottTodd, line 31: The name "os" is still a poor fit here as mentioned in my prior comment, but this is okay for an incremental change. Could also refactor _prior_ to this PR and then have this PR just switch the label and have the conditionals remain unchanged.

Eliasj42: Ok.  matrix.os is used throughout a lot workflows though

ScottTodd: Yep. We can clean that up in a separate PR.

ScottTodd (approved):
Let's keep an eye on the pip download time and make sure it gets cached on the runners. Adding 10 minutes of overhead to jobs is pretty a pretty large cost.

Eliasj42:
> Let's keep an eye on the pip download time and make sure it gets cached on the runners. Adding 10 minutes of overhead to jobs is pretty a pretty large cost.

Ok.  Do you know why its taking longer, I didn't add any pip requirements, and I'm pretty sure Stanley's patch is included

ScottTodd:
> > Let's keep an eye on the pip download time and make sure it gets cached on the runners. Adding 10 minutes of overhead to jobs is pretty a pretty large cost.
> 
> Ok. Do you know why its taking longer, I didn't add any pip requirements, and I'm pretty sure Stanley's patch is included

See the logs I linked above. Pip has a cache (typically at `~/.cache/pip`) and the previous runners had the packages already downloaded there.

Eliasj42:
> > > Let's keep an eye on the pip download time and make sure it gets cached on the runners. Adding 10 minutes of overhead to jobs is pretty a pretty large cost.
> > 
> > 
> > Ok. Do you know why its taking longer, I didn't add any pip requirements, and I'm pretty sure Stanley's patch is included
> 
> See the logs I linked above. Pip has a cache (typically at `~/.cache/pip`) and the previous runners had the packages already downloaded there.

ok, thanks.  I'll look into adding a pip cache to the ossci runners

ScottTodd:
It might just work after a few jobs get scheduled on the runners and populate the existing cache... but we probably want a more explicit strategy for populating and maintaining the various caches on the runners. For these 5GB, 5-10 minute downloads it's less critical but still very useful to get cache hits. For large language model weights (10GB+), avoiding cache misses is essential.

.github/workflows/ci-tk.yaml
yamiyysu: Question for my own understanding: why change the gpu distribute from 8 to 1? Is it b/c the runner only has 1 GPU?

PR #565, opened 2025-03-06
[Wave] Use a heuristic to pick the tile sizes for extend

harsh-nod:
This PR implements a empirically determined heuristic to determine the tile size when one is not explicitly provided by the user.

### DIFF
--- a/iree/turbine/kernel/wave/templates/extend_attention.py
+++ b/iree/turbine/kernel/wave/templates/extend_attention.py
@@ -69,6 +69,17 @@ def get_extend_attention_kernel(
     STORE_ELEMS_PER_THREAD = tkl.sym.STORE_ELEMS_PER_THREAD
 
     SEQ_TILE_SIZE = shape.block_size
+    N_KV_SCALE = 2
+
+    if SEQ_TILE_SIZE is None:
+        # Apply tile heuristic.
+        if shape.max_seq_len <= 128:
+            SEQ_TILE_SIZE = 64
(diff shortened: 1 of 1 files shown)


Hardcode84 (approved):

iree/turbine/kernel/wave/templates/extend_attention.py
```diff
@@ -69,6 +69,17 @@ def get_extend_attention_kernel(
     STORE_ELEMS_PER_THREAD = tkl.sym.STORE_ELEMS_PER_THREAD
 
     SEQ_TILE_SIZE = shape.block_size
+    N_KV_SCALE = 2
+
+    if SEQ_TILE_SIZE is None:
+        # Apply tile heuristic.
+        if shape.max_seq_len <= 128:
```
Hardcode84, line 76: Does our untitests cover both paths?

harsh-nod: No currently SEQ_TILE_SIZE is always set and we dont do a full sweep. But I think we should add that to our perf tests @nithinsubbiah

nithinsubbiah: Do we want to add other attention variants to perf test as well? My understanding was we check perf for only vanilla attention for the interested sizes

harsh-nod: Let's start with that and once you have that we can add some extend attention tests as well.

PR #571, opened 2025-03-10
[TKW] Turn off kernel cache by default

GMNGeoffrey:
https://github.com/iree-org/iree-turbine/pull/351 said that it did this,
but it actually didn't. The cache is actually quite unhelpful if you're
trying to debug Wave itself, which I think at this point is the main use
case for the majority of users.

### DIFF
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -29,7 +29,7 @@ from .utils import invoke_vmfb, _read_file, _write_file
 default_cache_base_dir = Path.home() / ".wave"
 CACHE_BASE_DIR = Path(os.environ.get("WAVE_CACHE_DIR", default_cache_base_dir))
 WAVE_ALWAYS_COMPILE = int(os.environ.get("WAVE_ALWAYS_COMPILE", 0))
-WAVE_CACHE_ON = int(os.environ.get("WAVE_CACHE_ON", 1))
+WAVE_CACHE_ON = int(os.environ.get("WAVE_CACHE_ON", 0))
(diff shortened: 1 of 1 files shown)


GMNGeoffrey:
Do we want to enable the cache for perf tests? https://github.com/iree-org/iree-turbine/blob/aff664e034e1ef11c7deb90f50680d68046c3f92/.github/workflows/perf.yaml#L67

raikonenfnu:
I think from previous offline convo with @Hardcode84 and @harsh-nod we decided to keep it on, since it is easy for "advanced"/"expert" user to just put it in our bashrc.

GMNGeoffrey:
Ah, I see. I went ahead with the PR because it looked like #351 had meant to do this. I think if you want to leave it on by default, there should be more documentation about it and a suggestion to turn it off for kernel authoring/debugging, perhaps in https://github.com/iree-org/iree-turbine/tree/main/iree/turbine/kernel/wave/README.md. I know it caused me some issues until I figured out what was going on and https://github.com/iree-org/iree-turbine/pull/500 suggests that Son was also having problems with it.

GMNGeoffrey:
Did we come to a conclusion here? I think @Hardcode84 was still advocating that this should be off by default. My argument is that right now most users of Wave are actually working on the compiler and so the cache is more harmful than it is helpful and should be off by default for now and can be turned on by default once the project is more stable and a higher fraction of usage is actually in a scenario where the cache is helfpul. I don't feel super strongly about that though. Instead we can add a note to https://github.com/iree-org/iree-turbine/blob/main/iree/turbine/kernel/wave/README.md that developers probably want to turn the cache off by default. In general, I think Wave would benefit from significantly expanded documentation.

PR #573, opened 2025-03-10
[TKW] Give boolean parametrizations better names

GMNGeoffrey:
This introduces a general mechanism for creating a boolean parametrization with a a more helpful name than the default. I'm also motivated to identify scheduling tests in particular because they seem to be either broken or impossibly slow on the machine I'm using. This at least gives me something in the name I can filter on (another option would be a marker, but that would be more invasive).


GMNGeoffrey:
I'm kind of skeptical that I caused some attention tests to output nans...

harsh-nod (approved):
lgtm! thank you!

PR #574, opened 2025-03-11
[TKW] Consistently return asm in test execution

GMNGeoffrey:
Cached kernels were returning the asm and uncached ones were returning the module builder.

I added cache statistics to the `WaveCacheManager` instead of the cache test using the returned type to distinguish a cache hit. I also renamed the `module_op` property on `WaveCache` to `asm as that's actually what it is (treating it like it was a module op was what was causing failures in the first place).

Another option would be to consistently return a dataclass that has the extra information present in `WaveCache`, but I didn't want to use a cache type the path where the cache is turned off and `WaveCache` also fetches asm lazily, which complicates things a bit. I think we'd want a `WaveKernel` type and two duck-typed implementations if we went that route.

Fixes https://github.com/iree-org/iree-turbine/issues/419

### DIFF
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -635,7 +635,6 @@ class LaunchableWave(Launchable):
             entrypoint_name,
         ) = self._trace_and_get_kernel_signature(args, kwargs)
 
-        asm = mb.module_op.get_asm()
         if run or run_bench or create_vmfb_file:
             if not config:
                 raise ValueError("no config provided")
@@ -643,6 +642,7 @@ class LaunchableWave(Launchable):
             host_codegen.isolated_test_call(
                 mb, exe, kernel_sig, entrypoint_name, dynamic_symbols
             )
+            asm = mb.module_op.get_asm()
             if config.get("print_mlir", False):
                 print(asm)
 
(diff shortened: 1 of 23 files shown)


iree/turbine/kernel/wave/wave.py
```diff
@@ -677,7 +677,7 @@ def test_execute(self, args, kwargs):
                 cache_manager.store_kernel(
                     compiled_wave_vmfb,
                     kernel_usages,
-                    mb.module_op.get_asm(),
```
raikonenfnu, line 680: this one would still need to be `mb.module_op.get_asm()` I reckon, since at this stage we are storing into the cache

iree/turbine/kernel/wave/wave.py
```diff
@@ -694,7 +694,7 @@ def test_execute(self, args, kwargs):
                 kernel_hash=kernel_hash,
             )
 
-        return mb
+        return mb.module_op.get_asm()
```
raikonenfnu, line 697: Going to defer on @harsh-nod  on this one whether or not we ever need to return the mb, if not this makes sense

raikonenfnu (commented):
Thanks Geoffery for the really nice refactorings! some comments though

iree/turbine/kernel/wave/wave.py
harsh-nod: Lets go with returning asm for now, we can revisit later.

harsh-nod (approved):
thanks for fixing!

PR #579, opened 2025-03-12
Implement Conv and backward Conv IR generation

zjgarvey:
First steps in auto-generating some BOO kernels for various convolution configurations. 

This commit adds some convenience structs for specifying a convolution signature, either by parsing an MiOpen driver command string, or by instantiation of the `ConvSignature` class in python.

Each `ConvSignature` instance uniquely determines an associated `torch.nn.Module`, which is then exported to MLIR.

Next steps:

1. Implement a base `CacheManager` similar to <https://github.com/iree-org/iree-turbine/blob/e83bc4f835239fdab5b56f24fda1cc7ac5b93076/iree/turbine/kernel/wave/cache.py#L108> from which we can customize hashing.
2. Define python callable "kernel templates", which should infer a signature from pytorch inputs and interact with the cache manager to launch a kernel. Think "kernel template" ~ iree-turbine CustomOp + cache management. 
3. Tune individual convolution kernels and provide pre-compiled high-performance vmfbs. Depending on the difficulty, allow automatically compiling and tuning a kernel if not in the cache. 
4. Develop a global template registry for all of our kernel "templates" and an API to easily lookup a registered template (e.g., from tkw, boo, elsewhere).


zjgarvey:
I'm really not sure what is going on with the implementation tests. I have the exact same torch version, and for some reason `w.grad` on the CI runs differs slightly from my machine for `bfloat16`. I think I'm just going to remove the bfloat16 param from the test for now. I just wanted to have something to verify the implementations are correct.

rsuderman (approved):
:ghost:

PR #581, opened 2025-03-13
[Wave] Refactor LoopConstructor infra to prep for heuristic base strat

raikonenfnu:
- Set scheduling to be an enum option/param instead of bool S.T we can choose different strategies in the future
- Refactor loop_reconstruction to use num_stages and initiation_interval as opposed to scheduler to make it easy to pass heuristic based schedule in the future.


harsh-nod (approved):
lgtm! thanks!

PR #582, opened 2025-03-14
[Wave] Add non-transposed V Attention kernel variant

nithinsubbiah:
Adds non-transposed V Attention kernel variant and test

### DIFF
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -166,7 +166,7 @@ class WaveCacheManager(object):
 
         # Benchmark related hash
         if run_bench and config != None:
-            key += config.get("benchmark_batch_size", "")
+            key += [config.get("benchmark_batch_size", "")]
         return hashlib.sha256(str(key).encode("utf-8")).hexdigest()
 
     ###############################################################################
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/wave/cache.py
```diff
@@ -166,7 +166,7 @@ def get_hash(
 
         # Benchmark related hash
         if run_bench and config != None:
-            key += config.get("benchmark_batch_size", "")
+            key += [config.get("benchmark_batch_size", "")]
```
raikonenfnu, line 169: Do we need this change?

nithinsubbiah: Yeah, otherwise there's an issue when an `int` is being subscripted

iree/turbine/kernel/wave/templates/vanilla_attention.py
```diff
@@ -164,4 +231,8 @@ def repeat(
         del hyperparams[B]
         del hyperparams[K2]
 
+    base_attention = base_attention_transposed_v
+    if not is_v_transposed:
```
raikonenfnu, line 235: NIT: Any thoughts on renaming `base_attention_non_transposed_v` to just `base_attention`, and then this will just be
```
if is_v_transposed:
  base_attention = base_attention_transposed_v
```

raikonenfnu (commented):
Overall looks good, just some NITs

iree/turbine/kernel/wave/cache.py
raikonenfnu: makes sense, thanks!

raikonenfnu (commented):
Looks good, will approve and land once tests passes!

harsh-nod (approved):
nice!

PR #584, opened 2025-03-14
[TKW] Paged decode attention rewrite

Hardcode84:
Rewrite paged decode to get it closer to the triton kernel impl.

* Distribute K2 loop across splits, which requires custom `iters` and `start` on `TilingConstraint` as well as some custom masking code.
* Add `start` support to `TilingConstraint`
* `global_to_shared_gathers.py`: clone dependent read nodes instead of modifying exiting ones inplace, as they can be used by the other ops.

### DIFF
--- a/iree/turbine/kernel/wave/global_to_shared_gathers.py
+++ b/iree/turbine/kernel/wave/global_to_shared_gathers.py
@@ -172,18 +172,32 @@ def construct_shared_read_mapping(read: Read) -> IndexMapping:
     return new_mapping
 
 
-def update_read_mapping_dynamic_values(read: Read) -> list[fx.Node]:
+def update_read_mapping_dynamic_values(read: Read):
     """
     Since every thread is now only reading one element along the gather dimension,
     we need to modify the read of the dynamic value to read only one element
     and do it at the specified index.
     """
+    new_dyn_vals = []
     for dyn_val in read.mapping_dynamic_vals:
         custom = get_custom(dyn_val)
-        custom.update_arg("elements_per_thread", 1)
-        for dim in dyn_val.index:
-            if dim in read.index:
-                dyn_val.index[dim] = read.index[dim]
+        assert isinstance(custom, Read), f"Only read nodes are supported, got {custom}"
+        # We cannot just update dynamic val inplace, as it may be used by other nodes
+        with custom.graph.inserting_before(dyn_val):
+            new_read = Read(
+                custom.memory,
+                1,
+                custom.mapping,
+                custom.mapping_dynamic_vals,
+            ).add_to_graph(custom.graph)
+            new_dyn_vals.append(new_read)
+
+            new_read.index = deepcopy(custom.index)
+            for dim in custom.index:
+                if dim in read.index:
+                    new_read.index[dim] = read.index[dim]
+
+    read.update_arg("mapping_dynamic_vals", new_dyn_vals)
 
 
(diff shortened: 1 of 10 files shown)


iree/turbine/kernel/wave/global_to_shared_gathers.py
```diff
@@ -172,18 +172,32 @@ def construct_shared_read_mapping(read: Read) -> IndexMapping:
...
     """
+    new_dyn_vals = []
     for dyn_val in read.mapping_dynamic_vals:
         custom = get_custom(dyn_val)
-        custom.update_arg("elements_per_thread", 1)
-        for dim in dyn_val.index:
-            if dim in read.index:
-                dyn_val.index[dim] = read.index[dim]
+        assert isinstance(custom, Read), f"Only read nodes are supported, got {custom}"
+        # We cannot just update dynamic val inplace, as it may be used by other nodes
```
harsh-nod, line 185: Can you move this comment to the function comment? Also can you add some comments saying that we generate redundant reads but expect them to be CSE-ed downstream ?

iree/turbine/kernel/wave/global_to_shared_gathers.py
```diff
@@ -172,18 +172,32 @@ def construct_shared_read_mapping(read: Read) -> IndexMapping:
...
+        with custom.graph.inserting_before(dyn_val):
+            new_read = Read(
+                custom.memory,
+                1,
```
harsh-nod, line 189: why is this 1? Shouldn't it be elements per thread?

lit_tests/kernel/wave/attention/decode_attention.py
```diff
@@ -63,24 +63,25 @@ def test_paged_flash_decoding():
...
+    # CHECK:                   scf.for %{{.*}} = %[[C0]] to %[[COUNT2]] step %[[C1]]
     # CHECK:                        amdgpu.lds_barrier
-    # 1 masked load for sequence idx, 4 for k_cache, and 4 for v_cache.
-    # CHECK-COUNT-1:                vector.maskedload
-    # CHECK-NEXT:                   arith.index_cast
-    # CHECK-COUNT-4:                vector.maskedload
-    # CHECK-NEXT:                   vector.store
-    # CHECK-COUNT-4:                vector.maskedload
-    # CHECK-NEXT:                   vector.store
+    # 2 masked load block table, 2 for k_cache, and 2 for v_cache.
```
harsh-nod, line 80: Are these the additional loads introduced in global_to_shared_gathers.py?

harsh-nod (approved):
nicely done! thanks for fixing decode and making it performant :) Just some minor comments.

iree/turbine/kernel/wave/global_to_shared_gathers.py
Hardcode84: Old code was setting it as well `custom.update_arg("elements_per_thread", 1)`

harsh-nod:
One more request and it doesn't have to be in this PR - can you please modify the symbols in this kernel so that they are consistent with what is in the other attention kernels (so use N_Q, D_Q etc)?

lit_tests/kernel/wave/attention/decode_attention.py
Hardcode84: Not by `global_to_shared_gathers` specifically, but by just paged decode rewrite

PR #587, opened 2025-03-15
[Wave] Add wave runtime

harsh-nod:
This PR adds a custom wave runtime that can be used to launch kernels. The runtime uses nanobind to transfer the kernel arguments from Python to C++ and then launches the kernels using the hip API. In order to enable this, we add the C++ code for wave runtime and link against nanobind and torch to generate a shared library that can be loaded.

For this runtime, we need the hsaco binary instead of the vmfb. So we cache the hsaco and associated kernel launch info when the cache is enabled. Finally, we add a pass to compute the shared memory usage by summing up the bytes used by all the shared memory allocations.

This PR will require all old cache directories to be removed.

### DIFF
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -59,10 +59,21 @@ jobs:
           pip install --no-cache-dir -r requirements-iree-pinned.txt --upgrade
           pip install -r requirements.txt -e .
 
-      - name: Install GPU rocm and pip deps
-        if: "(contains(toJSON(matrix.os), 'amdgpu') || contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
+      - name: Install GPU rocm and pip deps (mi300)
+        if: "(!contains(toJSON(matrix.os), 'amdgpu') && contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
         run: |
           sudo apt install -y rocm
+          python -m pip install --no-compile --upgrade pip
+          # Note: We install in three steps in order to satisfy requirements
+          # from non default locations first. Installing the PyTorch CPU
+          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          pip install --no-compile -r pytorch-rocm-requirements.txt
+          pip install --no-cache-dir -r requirements-iree-pinned.txt --upgrade
+          pip install -r requirements.txt -e .
+
+      - name: Install pip deps (mi250)
+        if: "(contains(toJSON(matrix.os), 'amdgpu') && !contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
+        run: |
           python -m pip install --no-compile --upgrade pip
           # Note: We install in three steps in order to satisfy requirements
           # from non default locations first. Installing the PyTorch CPU
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -59,9 +59,10 @@ jobs:
           pip install --no-cache-dir -r requirements-iree-pinned.txt --upgrade
           pip install -r requirements.txt -e .
 
-      - name: Install GPU pip deps
+      - name: Install GPU rocm and pip deps
         if: "(contains(toJSON(matrix.os), 'amdgpu') || contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
         run: |
+          sudo apt install -y rocm
           python -m pip install --no-compile --upgrade pip
           # Note: We install in three steps in order to satisfy requirements
           # from non default locations first. Installing the PyTorch CPU
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -16,7 +16,7 @@ import torch
 import threading
 
 from collections import OrderedDict
-from dataclasses import dataclass
+from dataclasses import dataclass, asdict
 from pathlib import Path
 import functools
 from typing import Any, Callable, Optional
@@ -30,8 +30,6 @@ from .utils import (
     _read_file,
     _write_file,
     KernelLaunchInfo,
-    _write_pickle_file,
-    _read_pickle_file,
 )
 
 default_cache_base_dir = Path.home() / ".wave"
@@ -228,8 +226,9 @@ class WaveCacheManager(object):
         if cur_hsaco_path:
             cur_hsaco_path = cur_hsaco_path[0]
             shutil.copy(cur_hsaco_path, cur_cache_basefile.with_suffix(".hsaco"))
-        cur_pkl_path = cur_cache_basefile.with_suffix(".pkl")
-        _write_pickle_file(cur_pkl_path, "wb", kernel_launch_info)
+        cur_kernel_info_path = cur_cache_basefile.with_suffix(".kernel_info.json")
+        kernel_info_str = json.dumps(asdict(kernel_launch_info))
+        _write_file(cur_kernel_info_path, "w", kernel_info_str)
 
     @staticmethod
     @functools.lru_cache
@@ -249,8 +248,9 @@ class WaveCacheManager(object):
         vmfb = _read_file(cur_vmfb_path, "rb")
         kernel_sig_str = json.loads(_read_file(cur_kernelsig_path, "r"))
         kernel_sig = [KernelBufferUsage[usage] for usage in kernel_sig_str]
-        cur_pkl_path = cur_cache_basefile.with_suffix(".pkl")
-        kernel_launch_info = _read_pickle_file(cur_pkl_path, "rb")
+        cur_kernel_info_path = cur_cache_basefile.with_suffix(".kernel_info.json")
+        kernel_info_str = json.loads(_read_file(cur_kernel_info_path, "r"))
+        kernel_launch_info = KernelLaunchInfo(**kernel_info_str)
         return WaveCache(kernel_hash, kernel_sig, vmfb, kernel_launch_info)
 
     ###############################################################################
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -6,6 +6,7 @@
 
 
 import copy
+import glob
 import hashlib
 import inspect
 import json
@@ -17,17 +18,25 @@ import threading
 from collections import OrderedDict
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable
 import functools
+from typing import Any, Callable, Optional
 
 from .constraints import Constraint, TilingConstraint, WaveConstraint
 from ..compiler.kernel_codegen import KernelBufferUsage
 from ..lang.wave_types import IndexMapping
 from .._support.indexing import IndexExpr
-from .utils import invoke_vmfb, _read_file, _write_file
+from .utils import (
+    invoke_vmfb,
+    _read_file,
+    _write_file,
+    KernelLaunchInfo,
+    _write_pickle_file,
+    _read_pickle_file,
+)
 
 default_cache_base_dir = Path.home() / ".wave"
 CACHE_BASE_DIR = Path(os.environ.get("WAVE_CACHE_DIR", default_cache_base_dir))
+WAVE_RUNTIME_DIR = CACHE_BASE_DIR / "wave_runtime"
 WAVE_ALWAYS_COMPILE = int(os.environ.get("WAVE_ALWAYS_COMPILE", 0))
 WAVE_CACHE_ON = int(os.environ.get("WAVE_CACHE_ON", 1))
 WAVE_CACHE_LIMIT = int(os.environ.get("WAVE_CACHE_LIMIT", 16))
@@ -48,10 +57,15 @@ class WaveCache:
     cache_id: str
     kernel_sig: tuple[KernelBufferUsage]
     vmfb: bytes
+    kernel_launch_info: Optional[KernelLaunchInfo] = None
 
     @property
-    def asm(self):
-        filepath = (CACHE_BASE_DIR / self.cache_id / self.cache_id).with_suffix(".mlir")
+    def asm_filepath(self):
+        return (CACHE_BASE_DIR / self.cache_id / self.cache_id).with_suffix(".mlir")
+
+    @staticmethod
+    @functools.lru_cache
+    def asm(filepath: str):
         with open(filepath, "r") as f:
             module_str = f.read()
         return module_str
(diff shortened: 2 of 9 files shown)


iree/turbine/kernel/wave/utils.py
```diff
@@ -670,6 +696,54 @@ def compile_to_vmfb(
...
+
+    # Ensure that the tensors are contiguous.
+    kern_args = []
+    for arg_tensor in kernel_inputs + kernel_outputs:
+        if not arg_tensor.is_contiguous():
+            arg_tensor = arg_tensor.contiguous()
+        kern_args.append(arg_tensor.data_ptr())
+
+    kernel_args = wave_runtime.Int64Vector(kern_args)
+    dyn_dims = wave_runtime.Int32Vector(list(kernel_dynamic_dims))
```
harsh-nod, line 742: should be Int64Vector

iree/turbine/kernel/wave/cache.py
```diff
@@ -206,6 +223,13 @@ def store_kernel_to_file(
         _write_file(cur_module_path, "w", module_str)
         kernel_sig_str = json.dumps([usage.name for usage in kernel_sig])
         _write_file(cur_kernelsig_path, "w", kernel_sig_str)
+        cur_hsaco_path = glob.glob(str(WAVE_RUNTIME_DIR / "*.hsaco"))
+        # Copy the hsaco file to the cache directory only if it exists.
+        if cur_hsaco_path:
+            cur_hsaco_path = cur_hsaco_path[0]
+            shutil.copy(cur_hsaco_path, cur_cache_basefile.with_suffix(".hsaco"))
+        cur_pkl_path = cur_cache_basefile.with_suffix(".pkl")
+        _write_pickle_file(cur_pkl_path, "wb", kernel_launch_info)
```
raikonenfnu, line 232: Is this to pickle the dataclass of?
```python
    grid: tuple[int] = None
    blocks: tuple[int] = None
    shared_memory_bytes: int = 0
    func_name: str = ""
```
If yes, any thoughts on using jsons instead? I think json serialization and deserialization is faster thank python pickle + could be easier to debug and move around?

iree/turbine/kernel/wave/cache.py
```diff
@@ -206,6 +223,13 @@ def store_kernel_to_file(
         _write_file(cur_module_path, "w", module_str)
         kernel_sig_str = json.dumps([usage.name for usage in kernel_sig])
         _write_file(cur_kernelsig_path, "w", kernel_sig_str)
+        cur_hsaco_path = glob.glob(str(WAVE_RUNTIME_DIR / "*.hsaco"))
+        # Copy the hsaco file to the cache directory only if it exists.
+        if cur_hsaco_path:
+            cur_hsaco_path = cur_hsaco_path[0]
+            shutil.copy(cur_hsaco_path, cur_cache_basefile.with_suffix(".hsaco"))
+        cur_pkl_path = cur_cache_basefile.with_suffix(".pkl")
```
raikonenfnu, line 231: I have a some questions on this:
1. Do we expect the directory/file to exist because the one who generated it will be the `dump_binaries` config?
2. Why always choose the first file found? `cur_hsaco_path[0]`, shouldn't we match to hash name?

.github/workflows/ci-tk.yaml
```diff
@@ -59,9 +59,10 @@ jobs:
           pip install --no-cache-dir -r requirements-iree-pinned.txt --upgrade
           pip install -r requirements.txt -e .
 
-      - name: Install GPU pip deps
+      - name: Install GPU rocm and pip deps
         if: "(contains(toJSON(matrix.os), 'amdgpu') || contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
         run: |
+          sudo apt install -y rocm
```
raikonenfnu, line 65: Wouldn't the dockers already come with rocm? if not how are we running the e2e test rn?

iree/turbine/kernel/wave/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
+cmake_minimum_required(VERSION 3.15...3.27)
+project(wave_runtime)
+
+if (CMAKE_VERSION VERSION_LESS 3.18)
+  set(DEV_MODULE Development)
+else()
+  set(DEV_MODULE Development.Module)
+endif()
+
+find_package(Python 3.10 COMPONENTS Interpreter ${DEV_MODULE} REQUIRED)
```
raikonenfnu, line 10: Silly Q, if we use 3.10 here, and assume I have 3.10 and 3.12 installed on my local machine but my venv is running at 3.12 , would this still work? Additionally, what if my machine do not have a 3.10 but only a 3.12?

Not needed in this PR itself but perhaps need to add a TODO here to make it work with different python versions..

raikonenfnu (commented):
This is awesome stuff! overall looks great! just some minor questions

iree/turbine/kernel/wave/wave.py
```diff
@@ -488,13 +539,24 @@ def _trace_and_get_kernel_signature(
...
         _warn_iree_is_too_old()
 
+        # Store kernel launch info for the wave runtime.
+        kernel_launch_info = KernelLaunchInfo()
+
+        # Build wave runtime, if specified.
+        run_config = kwargs.get("run_config", {})
+        if run_config.get("wave_runtime", None):
+            # Remove any existing hsaco files in this directory.
+            remove_files_with_extension(WAVE_RUNTIME_DIR, ".hsaco")
```
raikonenfnu, line 557: wouldn't this mean between python runs, we are re-compiling the kernels?

iree/turbine/kernel/wave/utils.py
```diff
@@ -670,6 +696,54 @@ def compile_to_vmfb(
...
+    """
+    import wave_runtime
+    from .cache import WAVE_RUNTIME_DIR, CACHE_BASE_DIR
+
+    # Get the path to the binary.
+    if kernel_hash:
+        binary = str(CACHE_BASE_DIR / kernel_hash / kernel_hash) + ".hsaco"
+    else:
+        bin_dir = WAVE_RUNTIME_DIR
+        binary = glob.glob(str(bin_dir / "*.hsaco"))[0]
```
raikonenfnu, line 717: similar Q to previous one, can you explain the mechanism behind selecting the first found file in WAVE_RUNTIME_DIR/*.hsaco?

Also NIT: what about `binary = glob.glob(str(WAVE_RUNTIME_DIR / "*.hsaco"))[0]`?

iree/turbine/kernel/wave/cache.py
raikonenfnu: Actually wrt speed of serialization and deseralization after python3 it doesn't seem to be the case anymore since it uses cPickle by default. But I think json is still worth considering for it's portability and debugability . :)

.github/workflows/ci-tk.yaml
harsh-nod: Yes good question. They should come with rocm but they dont. I think some parts of rocm are installed but not the entire suite of apps. Once we switch to docker images with all of rocm installed, we shouldn't need this ( I was surprised that we did ).

iree/turbine/kernel/wave/runtime/CMakeLists.txt
harsh-nod: So the 3.10 here represents a minimum version. But good question since I ran into something similar on the ci. The find package here is being guided by the way we invoke cmake. So if you see we pass in the path of the python executable that is running the wave code and this way we make sure that it doesnt pick up some other system python installed.

iree/turbine/kernel/wave/wave.py
harsh-nod: We do this because we want to make sure that at any point of time there's only a single hsaco file located inside the wave runtime directory inside the cache directory. This only happens in situations where we are not caching any kernels and so it is okay to remove the last compiled kernel.

iree/turbine/kernel/wave/utils.py
harsh-nod: By removing the last compile file we make sure that there will always be only one kernel inside the wave runtime directory inside the cache directory. That is why it's okay for us to just use the first file that we find. Sure I can modify that.

iree/turbine/kernel/wave/cache.py
harsh-nod: Yes, when the wave runtime is enabled, then we do expect the file to exist because it will be generated by iree because we pass in the dump_binaries flag. We can always choose the first file found because we are maintaining the invariant that at any point of time they will always be only one hsaco file in the wave runtime directory.

iree/turbine/kernel/wave/cache.py
harsh-nod: I am not particularly tied to pickle, it was the easiest thing to use at the time. I think speed of pickle may even be faster than json, but I can change it to make it more consistent (I believe we are not using pickle anywhere else in the codebase) and for debuggability.

harsh-nod: Done.

.github/workflows/ci-tk.yaml
raikonenfnu: For education purposes, do you know which things we are missing, maybe we can fix it at the root cause in the future

iree/turbine/kernel/wave/wave.py
raikonenfnu: Is this because we would use `(CACHE_BASE_DIR / kernel_hash / kernel_hash) + ".hsaco"` if we have cached and `WAVE_RUNTIME_DIR/*.hsaco` is only used as a temporary location/cases where we do not have it cached?

.github/workflows/ci-tk.yaml
harsh-nod: hipsparse was the first one missing.

iree/turbine/kernel/wave/wave.py
harsh-nod: yes thats right.

raikonenfnu: Doesn't have to be this PR, but we should add this in a comment somewhere since it is not very obvious (at least for me :D)

harsh-nod: Sounds good, will add in next PR.

harsh-nod: Runners dont seem to be picking this up, so will add the comment in this PR and resubmit.

iree/turbine/kernel/wave/wave.py
```diff
@@ -136,6 +148,44 @@ def _warn_iree_is_too_old():
...
+    args = [
+        "cmake",
+        f"-DPython_EXECUTABLE={sys.executable}",
+        "-S",
+        str(runtime_path),
+        "-B",
+        str(runtime_build_path),
+    ]
+    try:
+        subprocess.run(args, check=True)
```
ScottTodd, line 171: Python should not be running CMake commands on the system at runtime. This needs a system architecture discussion.

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+std::unordered_map<std::string, std::tuple<hipModule_t, hipFunction_t>> cache;
+using Int64Vector = std::vector<uint64_t>;
+using Int32Vector = std::vector<uint32_t>;
+
+int launch(const KernelLaunchInfo &info, const Int64Vector &tensors, const Int64Vector &dynamicDims)
+{
+    hipStream_t stream = at::hip::getCurrentHIPStream();
+    hipModule_t module;
+    hipFunction_t function;
```
ScottTodd, line 54: This "wave runtime" is using a generic name but is only supporting a single GPU API. IREE and iree-turbine both support CPU, Vulkan, CUDA, ROCm, Metal, and out of tree APIs. Extensions of the PyTorch ecosystem like iree-turbine could be limited to what PyTorch already supports (CPU, CUDA, ROCm, and a few other APIs IREE doesn't target), but we should at least be careful to set up an architecture that doesn't tie us directly to HIP/ROCm.

iree/turbine/kernel/wave/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+
+# Link against torch, using the libraries and include dirs in site-packages.
+execute_process(
+  COMMAND "${Python_EXECUTABLE}" -c "import torch; print(torch.utils.cmake_prefix_path)"
+  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_cmake_PATH)
+list(APPEND CMAKE_PREFIX_PATH ${torch_cmake_PATH})
+list(APPEND CMAKE_PREFIX_PATH "/opt/rocm")
+
+find_package(Torch REQUIRED)
+find_package(HIP REQUIRED)
```
ScottTodd, line 31: These required deps are going to be very painful across systems and platforms. What does the runtime actually need from these packages? Could we use header files and use dynamic linking to access APIs that we need?

ScottTodd (commented):
I have some major concerns about the architecture here. Can we roll back to an RFC / design discussion for this new component?

iree/turbine/kernel/wave/wave.py
raikonenfnu: Yeah, it's a plan of record to move this to setup.py. IIRC this was semi inspired by triton, who's runtime is compiled by calling clang from python haha

iree/turbine/kernel/wave/runtime/CMakeLists.txt
raikonenfnu: but nanobind without cmake may be difficult

iree/turbine/kernel/wave/wave.py
harsh-nod: Yes, this was a stepping stone to prove this concept. Now that we have proved its worth, the plan is to move it to setup.py.

iree/turbine/kernel/wave/runtime/runtime.cpp
harsh-nod: Good point. I think this is an area where documentation can help where currently we only support MI, but could do more in the future.

iree/turbine/kernel/wave/runtime/CMakeLists.txt
harsh-nod: We need to link against torch and hip. For linking against torch, I believe find_package is the most robust way but open to other suggestions? find_package for Torch requires rocm to be installed on the system (this is all of rocm including hipblaslt, hipsparse etc.)(https://github.com/pytorch/pytorch/blob/1b373f6cd4bfe3ffdbab8aeee06decb1dc656f7d/cmake/public/LoadHIP.cmake#L146) but our rocm builders should have this already installed?

iree/turbine/kernel/wave/wave.py
ScottTodd: iree-turbine is currently a pure Python package, portable across all operating systems and python versions. Ignoring all details about what the new native code is doing (and those are certainly nontrivial), that immediately introduces additional complexity for building, testing, and packaging in this repository.

There are multiple different ways to take the various packages involved here, hence this needs a system architecture discussion. That should ideally have happened before merging the code to `main`, and I'm personally not comfortable with iterating on the `main` branch without some strong signaling in documentation and in code that the new components are optional and experimental.

PR #588, opened 2025-03-15
[TKW] Self index refactoring

Hardcode84:
Use sympy expression instead of constructing arith ops directly.


harsh-nod (approved):
This looks good, but please make sure it does not regress extend until the perf-ci comes online.

Issue #420, opened 2025-01-27
[TKW] Bug: cache tests fail if cache already exists


GMNGeoffrey:
The cache tests are supposed to reset the cache before they run by calling `reset_cache_manager`, but that method just returns early if the cache hasn't been initialized without deleting the cache directory:

https://github.com/iree-org/iree-turbine/blob/bca982a7776ec13be73a2f6627fb129f47b6d2a4/iree/turbine/kernel/wave/cache.py#L269-L275

so if you run them once and create a cache entry and then run them again, the check that the first call wasn't cached fails:

https://github.com/iree-org/iree-turbine/blob/bca982a7776ec13be73a2f6627fb129f47b6d2a4/tests/kernel/wave/runtime/cache_test.py#L207-L209

GMNGeoffrey:
Either `reset_cache_manager` should delete the cache directory regardless, the tests should first initialize the cache manager before resetting it (seems a bit weird), or there should be a separate method for clearing the cache independent of resetting the cache manager. Given that the cache tests are the only users of `reset_cache_manager` AFAICT, I think option 1 makes the most sense.

GMNGeoffrey:
It's also maybe a bit aggressive to have the cache tests reset someone's entire cache. Maybe they should use a private cache in a temporary directory instead? That solves both problems.

PR #590, opened 2025-03-17
Inject cache directory into the cache manager

GMNGeoffrey:
The current setup hard-codes the cache directory for the global cache manager based on an environment variable. This is a problem particularly in tests because it causes the tests to be non-hermetic and to mess with the users cache directory.

I think it would be better if the tests could instantiate their own cache manager explicity rather than relying on the global, but that would require a more significant restructuring.

Note that this removes the lazy loading of cached kernel asm. I realized that https://github.com/iree-org/iree-turbine/pull/574 effectively made this useless as the only place where a `WaveCache` was being used, the asm was immediately retrieved. If we want to restore that behavior then we'd need to go back to returning a more complicated structure instead of the asm, but it should be a consistent structure as discussed in https://github.com/iree-org/iree-turbine/pull/574.

It incidentally removes a bunch of usages of the custom `_read_file` and `_write_file` convenience functions that were being called on `pathlib.Path` objects. These functions are pretty much redundant with the `pathlib` standard library functions. I didn't go so far as purging their usage in the rest of the codebase, as I didn't want to expand this PR beyond the kernel cache, but I think they should probably be removed.

Fixes https://github.com/iree-org/iree-turbine/issues/420


martin-luecke (approved):
LGTM after fixing the conflict and rebasing

PR #591, opened 2025-03-17
[WIP] Set up BOO kernel-generation and caching

zjgarvey:
The general idea is that we should be able to:

- Write a pytorch function / Module that should really be a single dispatch.
- Import to MLIR and try to compile it as a single kernel. 
- Re-factor existing kernel cache manager from wave to allow caching these kernels too.

The `WaveCacheManager` is converted to `KernelCacheManager`, which allows customizing the `get_hash` method by defining a new `KernelNamespace`, and will organize the file cache into `BASE_CACHE_DIR / namespace / func_name / kernel_hash`. Ideally, I'd like to eventually be able to download some pre-tuned kernels to the cache and fetch them based on the `func_name` and input types. 

A really simple `BOOLaunchable` is defined for compiling/running/fetching kernels from cache. A decorator `@boo_kernel` is added so that a user can decorate a pytorch function to define a new `BOOLaunchable` with ease. 

### Some TODO's: 

1. Remove `WaveCacheManager` and refactor `LaunchableWave` to use the `KernelCacheManager` with a registered "wave" namespace. 
2. Find a less expensive way to infer the output shape if not specified by the user (or change the invocation method used in `BOOLaunchable` to not require an existing `torch.Tensor` for each output?).
3. Add some default compile/run configs for `BOOLaunchable` that can be inferred by which device inputs are located on, and implement an `eager_execute` that is a bit easier to use.
4. Figure out how to do in-place invocations for `BOOLaunchable`.
5. Actual Tests instead of examples
6. Merge with `conv_exports` work in <https://github.com/iree-org/iree-turbine/pull/579> and add some conv launchables to "the bag" for users to import and apply.


zjgarvey:
The last commit is scratch work trying to do some kind of eager launchable, but it is really bad. I don't think the launchables created within the eager calls persist, so their own internal executable binary cache is getting wiped each time. There are too many different caches, and I'm going to really strip this down tomorrow and figure out exactly what needs to be cached and where. 

There is also an example of using CustomOp to try and "fuse" CustomOps with torch ops. CustomOp eager execution is apparently not async compatible, so I'm just adding this temporarily.

PR #600, opened 2025-03-18
Added roundeven op

xintin:
This PR adds roundeven op (listed [here](https://github.com/iree-org/iree-turbine/issues/278)).


raikonenfnu (commented):
Overall looks good, let's wait until tests pass that we can land

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -84,6 +84,7 @@
     shared_memory_barrier,
     shuffle,
     tanh,
+    roundeven,
```
Hardcode84, line 87: nit: please keep this list sorted

Hardcode84 (approved):

PR #606, opened 2025-03-20
[Wave] Compute grid before kernel launch

harsh-nod:
This PR fixes an erroneous assumption made by the previous PR that when using the wave runtime the grid is always constant. Instead, we use sympy to generate a lambda (and lambda str when caching the kernel) that can be used to determine the grid prior to launching the kernel.

### DIFF
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 # Lang, compiler, ops, constraints
+from sympy.utilities.lambdify import lambdastr
 import iree.turbine.kernel.lang as tkl
 from ..compiler import builder, dispatch_codegen, kernel_codegen, host_codegen
 from ..lang import Grid, IndexMapping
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/wave/wave.py
```diff
@@ -648,11 +649,13 @@ def _trace_and_get_kernel_signature(
...
-        try_convert_to_int = lambda x: int(x) if isinstance(x, sympy.Integer) else x
-        kernel_launch_info.grid = [
-            try_convert_to_int(safe_subs(x, dynamic_symbols_map))
-            for x in self.grid_type.dims
-        ]
+        # Convert the grid into a lambda that we can use to compute the grid dimension.
+        kernel_launch_info.grid = sympy.lambdify(
+            [list(dynamic_symbols_map.keys())], self.grid_type.dims
+        )
+        kernel_launch_info.grid_str = lambdastr(
```
raikonenfnu, line 656: is grid_str mostly for caching purposes?

harsh-nod: Yes and also helps from a debuggability perspective because you can see the lambda function.

raikonenfnu: makes sense! thanks :)

raikonenfnu: Seems like most of the time we are rematerializing it with `kernel_info_str["grid"] = eval(kernel_info_str["grid_str"])`. Is it possible to refactor to get rid of `kernel_info_str["grid"] ` and `kernel_launch_info.grid` by opting to do a local `grid = eval(kernel_info_str["grid"])` inside `invoke_with_wave_runtime`

nithinsubbiah (approved):
Looks good to me

iree/turbine/kernel/wave/wave.py
harsh-nod: There are two separate use cases. In the case where there's no caching, it does not make sense to construct a string and then convert the string to a lambda when we are about to launch the kernel, because we can directly convert the sympy expression into a lambda function. When we have to cache, then it makes sense to use the string representation since we can use json to store it and inspect it. In that situation, we have to use eval to convert a string into a lambda but luckily it's only a one time cost. In general I don't think it makes sense to eval every single time we want to run the kernel as that is an expensive operation.

PR #608, opened 2025-03-21
[BOO] Adds a function for getting a Launchable from a ConvSignature

zjgarvey:
Adds a function to convert a `ConvSignature` into a `Launchable`, with added mlir and vmfb file caching/retrieval. 

Modifies `Launchable.jit_compile` to be able to optionally save and load jit-compiled artifacts within a directory.

### DIFF
--- a/iree/turbine/kernel/boo/examples/conv_launchable.py
+++ b/iree/turbine/kernel/boo/examples/conv_launchable.py
@@ -9,60 +9,51 @@ from iree.turbine.support.logging import runtime_logger as logger
 
 # run this script with the env variable: TURBINE_DEBUG="log_level=DEBUG"
 
+devices = [torch.device("cpu")]
+if torch.cuda.is_available():
+    devices.append(torch.device("cuda:0"))
+    if torch.cuda.device_count() > 1:
+        devices.append(torch.device("cuda:1"))
 
-def main():
+
+def main(clear_cache: bool):
+    """Gets a launchable conv kernel and applies it twice on each device in devices."""
+
     # get a conv signature
-    sig = ConvSignature(input_shape=[1, 2, 16, 16], kernel_shape=[1, 2, 2, 2])
+    sig = ConvSignature(
+        input_shape=[1, 2, 16, 16], kernel_shape=[1, 2, 2, 2], dtype=torch.float32
+    )
 
     # get a launchable kernel
     conv = get_launchable(sig)
+    seed = 1
+    for torch_device in devices:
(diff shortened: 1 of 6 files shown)


iree/turbine/kernel/boo/examples/conv_launchable.py
```diff
@@ -0,0 +1,34 @@
...
+
+# get a conv signature
+sig = ConvSignature(input_shape=[1, 2, 16, 16], kernel_shape=[1, 2, 2, 2])
+
+# get a launchable kernel
+conv = get_launchable(sig)
+
+# get some sample args on a specific device
+torch_device = torch.device("cuda:0") if torch.cuda.is_available() else None
+args = sig.get_sample_conv_args(seed=10, device=torch_device)
```
MaheshRavishankar, line 18: Are the arguments already on the device or do they need to be copied?

zjgarvey: I'm creating new random inputs and moving them to the device (I'm not sure if you can make the random tensor natively on GPU, but I can look into this). If the tensors happened to live on the same GPU to start with, that would also be fine and `Launchable` will not move them around.

zjgarvey: Yeah, it looks like there is a way to make the generator work on GPU. I'll try it out.

MaheshRavishankar (commented):
From what I have read the code, this seems OK to me, but this is new to me. Can either @rsuderman  or @stellaraccident  review this.

cc @AWoloszyn  as well.

Do we need some smoke tests here?

zjgarvey:
@MaheshRavishankar I added a test to at least check the vmfb file caching, but it's a bit hacky. I'll try to grab a review from Rob.

iree/turbine/kernel/boo/examples/conv_launchable.py
```diff
@@ -0,0 +1,59 @@
...
+    clear_cache_dir,
+)
+from time import time
+from iree.turbine.support.logging import runtime_logger as logger
+
+# run this script with the env variable: TURBINE_DEBUG="log_level=DEBUG"
+
+devices = [torch.device("cpu")]
+if torch.cuda.is_available():
+    devices.append(torch.device("cuda:0"))
```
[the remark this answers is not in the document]
zjgarvey: I don't want to run it on every available device, just more than 1 gpu to check out the caching from the debug logs.

PR #610, opened 2025-03-21
[runtime] Tier out internal Launchable caching and get hal device from uuid

zjgarvey:
This breaks up the `Launchable`'s internal cache to store two mappings:

`device_type_key -> (entry_point, main_module)`
`device_instance_key -> (vm_context, main_function)`

And converts `_Loader` to be a callable returning tuples of the form `(entry_point, main_module)` for each device, essentially making `Launchable.__init__` roughly equivalent to the old `Launchable.from_vm_module` (which is removed). 

This PR also tries to infer the correct hal device from the `torch.device` by looking up the associated gpu uuid in the hal driver infos (although there may be runtime api I'm missing to do this directly).

Added tests to `launch_test.py` for validating that a single `Launchable` indeed can launch on multiple GPUs.

### DIFF
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -246,6 +246,7 @@ class Device:
 
     def _recompute_target_keys(self):
         self.type_cache_key = f"{self.driver_id}:{';'.join(self.compile_target_flags)}"
+        self.instance_cache_key = f"{self.driver_id}:{self._s.enumerated_device_id}"
 
     @property
     def hal_device(self) -> HalDevice:
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from functools import lru_cache
-from typing import Callable, Optional, Union
+from typing import Any, Callable, Dict, Optional, Union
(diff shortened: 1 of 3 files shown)


iree/turbine/runtime/device.py
```diff
@@ -211,6 +211,7 @@ def __new__(
 
     def _initialize(self):
         d = self._s.device
+        print(d)
```
stellaraccident, line 214: Remove extra debugging.

tests/runtime/launch_test.py
```diff
@@ -84,5 +87,46 @@ def testLaunchParams(self):
         torch.testing.assert_close(expected, result)
 
 
+class SameLaunchableDifferentDevicesTest(unittest.TestCase):
```
stellaraccident, line 90: I think this will work even for a single device so can leave it like this. Otherwise, it needs some kind of skipIf tag. Best to include a comment saying that this should also pass with a single HIP device but it is intended to validate the multiple device case.

stellaraccident (approved):
Thanks!

tests/runtime/launch_test.py
zjgarvey: Ah, good point. On cpu this is doing the same thing as the other tests anyway, so we should skip it if not `torch.cuda.is_available`. I'll also add some comments.

PR #612, opened 2025-03-23
[Wave] Introduce wave compile and partition utils

harsh-nod:
This PR introduces a wave_compile API to replace the old TestLaunchContext approach which was confusing for users and conflated the compilation and invocation steps. The new approach consolidates all the compile options in a single class and utilizes this in both the compile and invocation of the kernel.

Also, we partition the utils file into several smaller specific utilities so other files in the project can import what they need as opposed to a large utils file with several dependencies.


iree/turbine/kernel/wave/compile.py
```diff
@@ -0,0 +1,132 @@
...
+            if usage == kernel_codegen.KernelBufferUsage.INPUT:
+                kernel_inputs.append(arg)
+
+            if usage == kernel_codegen.KernelBufferUsage.OUTPUT:
+                kernel_outputs.append(arg)
+
+        # Get cached kernel when available.
+        cache_enabled = is_cache_enabled()
+        kernel_hash = None
+        if cache_enabled:
```
Hardcode84, line 52: None of the caching logic depends on invoke args. You can gen vmfb once in `__init__` (from cache or otherwise) and just use it here without any additional checks.

iree/turbine/kernel/wave/compile.py
```diff
@@ -0,0 +1,132 @@
...
+                kernel_hash = cache_manager.get_hash(self.subs, self.options)
+                self.cached_kernel_hash[:] = [kernel_hash]
+
+            cached_kernel = cache_manager.load_kernel(kernel_hash)
+            if cached_kernel:
+                invoke_vmfb(cached_kernel, self.options, kernel_inputs, kernel_outputs)
+                return cached_kernel.asm(cached_kernel.asm_filepath)
+
+        invoke_vmfb(self.executable, self.options, kernel_inputs, kernel_outputs)
+        return self.options.mb.module_op.get_asm()
```
Hardcode84, line 69: `return self.asm`?

iree/turbine/kernel/wave/compile.py
harsh-nod: good point, thanks that should speed up things!

iree/turbine/kernel/wave/compile.py
```diff
@@ -0,0 +1,128 @@
...
+    compiled_wave_vmfb = compile_to_vmfb(asm, options)
+    if options.create_vmfb_file:
+        _write_file(options.create_vmfb_file, "wb", compiled_wave_vmfb)
+
+    kernel_usages = [
+        binding.kernel_buffer_type.usage
+        for binding in kernel_sig.kernel_buffer_bindings
+    ]
+
+    if is_cache_enabled():
```
raikonenfnu, line 111: I may be missing something here, can you explain when are we "skipping" the compilation pipeline if we found it in the cache? Seems like right now it is only storing into cache

iree/turbine/kernel/wave/compile.py
```diff
@@ -0,0 +1,128 @@
...
+        kernel_outputs = []
+        for arg, b in zip(args, self.options.kernel_sig.kernel_buffer_bindings):
+            usage = b.kernel_buffer_type.usage
+            if usage == kernel_codegen.KernelBufferUsage.INPUT:
+                kernel_inputs.append(arg)
+
+            if usage == kernel_codegen.KernelBufferUsage.OUTPUT:
+                kernel_outputs.append(arg)
+
+        invoke_vmfb(self.executable, self.options, kernel_inputs, kernel_outputs)
```
raikonenfnu, line 58: btw seems like the old kernel required us to feed in dynamic values/symbols, do we not need it any more here?
```py
    kernel_dynamic_dims = []
    if dynamic_symbols:
        kernel_dynamic_dims = dynamic_symbols_map.values()
```

iree/turbine/kernel/wave/compile.py
```diff
@@ -0,0 +1,128 @@
...
+        return self.asm
```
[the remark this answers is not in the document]
raikonenfnu: I wonder if this has any runtime perf deg too, since we are potentially moving some large strings around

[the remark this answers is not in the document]
harsh-nod: Makes sense, we should not always return the asm but instead query it and make it a property, in the next PR (since that will also require significant changes)

iree/turbine/kernel/wave/compile.py
harsh-nod: We never skip the compilation pipeline. Now that compilation and invocations are separated, we compile only once and invoke multiple times. Does that make sense?

iree/turbine/kernel/wave/compile.py
harsh-nod: It is now contained inside the options so we don't need to explicitly pass it.

raikonenfnu: nice!

iree/turbine/kernel/wave/compile.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def wave_compile(options: WaveCompileOptions, kernel: "LaunchableWave") -> WaveKernel:
+    """
+    Compiles the wave kernel to an executable.
+    """
```
harsh-nod, line 65: Move load_from_cache here

iree/turbine/kernel/wave/compile.py
raikonenfnu: 1. is resolved with LRU caching at sglang level
2. will be resolved by moving `load_from_cache` into start of wave_compile

raikonenfnu (approved):

iree/turbine/kernel/wave/compile.py
harsh-nod: done

iree/turbine/kernel/wave/compile.py
```diff
@@ -0,0 +1,126 @@
...
+
+class WaveKernel:
+    """
+    Represents a wave kernel that can be invoked by the user.
+    """
+
+    def __init__(self, options: WaveCompileOptions, executable: Any, asm: str):
+        self.options = options
+        self.cached_kernel_hash = self.options.kernel_hash
+        self.subs = options.subs
```
Hardcode84, line 24: Is  `cached_kernel_hash` and `subs` used anywhere?

raikonenfnu: >cached_kernel_hash 
+1,  kernel_hash itself is useful for users to be able to specify kernel hash they want, with that and LRU cache and they can re-use the kernel efficiently without recompiling and recomputing hash at all.  I think within the scope of this class may not be very useful though

harsh-nod: yeah not being used, will remove. Could be useful to keep track of the hash in the kernel, but can't think of a use case yet. We can bring it back when required.

Hardcode84 (approved):

PR #614, opened 2025-03-24
[Wave] Add 2-stage prefetch scheduling strategy for GEMM

raikonenfnu:
This PR is a MVP/first bring up for a heuristics based scheduling strategy into Wave. In this PR, we introduce 2-stage prefetch scheduling (strategy currently only tested for GEMM).  Just for a brief recap, what we mean by 2 stage prefetch strategy is as such:

```
    Convert vanilla schedule of:
        for i = 0 to N:
            a = READ_GLOBAL i
            WRITE_SHARED a
            barrier
            b = READ_SHARED
            COMPUTE b

    into prefetch schedule:
        a_0 = READ_GLOBAL 0
        WRITE_SHARED a_0
        for i = 0 to N - 1:
            a_{i+1} = READ_GLOBAL i + 1
            // a_{i+1} is NOT blocked by this barrier because barriers only block shared memory transfers
            barrier
            b_i = READ_SHARED
            COMPUTE b_i
            barrier
            WRITE_SHARED a_{i+1}
        barrier
        b_N = READ_SHARED
```

In order to do that we needed to bring these changes:

1. Implement a scheduler class that generate a schedule and initiation interval that generates a 2 stage prefetch strategy. The crux of this implementation is:
  a. we'd need to sort the nodes inside `reduction_graph` but it' accumulative scheduling weight (accumulative on it's predecessors), this should be rather similar to topological graph but more robust since it is based on scheduling edges weight. 
  b. After the nodes are sorted, we classify them to one of the 4 stages in prefetch strategy which is `GLOBAL_LOAD`, `LOCAL_STORE`, `LOCAL_LOAD`,  and `COMPUTE`, where each stage will correspond to a clock cycle. 
  c. Once this is done, we would simply set initiation interval to 2 which should let the loop reconstructor to generate a 2 stage prefetch strategy.
 
2. Modify scheduling/resources.py's implementation of `get_custom_operation_type` to index on shared reads, since global reads cannot be found in `reduction_graph`, which is the form used by prefetch scheduler (and other schedulers for that matter). 

3. General infra update and tests

Future work:
- Currently performance does not improve, but big hunch we'd just need to remove a redundant lds barrier to unlock perf.
- Another perf factor is to add set_prio instruction for better time slicing / resource overlap/use in a single SIMD.
- Apply to attention and extend attention

### DIFF
--- a/iree/turbine/kernel/wave/scheduling/graph_utils.py
+++ b/iree/turbine/kernel/wave/scheduling/graph_utils.py
@@ -6,7 +6,7 @@
 
 import torch.fx as fx
 from random import Random
-from collections import defaultdict
+from collections import defaultdict, deque
 from ..._support.indexing import index_symbol, IndexExpr
 from .resources import *
 from dataclasses import dataclass
@@ -15,7 +15,7 @@ import math
 from functools import partial
 from ..utils.symbol_utils import safe_subs
 import multiprocessing as mp
-from typing import Optional
+from typing import Optional, Callable
 
 T = index_symbol("$INITIATION_INTERVAL")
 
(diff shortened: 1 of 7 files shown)


raikonenfnu:
Generated schedule:
schedule_gen

iree/turbine/kernel/wave/scheduling/prefetch_scheduling.py
```diff
@@ -0,0 +1,215 @@
...
+        graph: fx.Graph,
+        edges: list[Edge],
+        resources: list[int],
+    ) -> None:
+        self.graph = graph
+        self.edges = edges
+        self.resources = resources
+        self.seed = 2024
+
+    def weighted_topological_sort(self, graph, edges):
```
raikonenfnu, line 101: Move this to graph utils

raikonenfnu: done!

iree/turbine/kernel/wave/scheduling/graph_utils.py
```diff
@@ -362,3 +362,80 @@ def create_scheduling_edges(
...
+    scheduling weight is suppose to quantify how early or late a node should live in a graph.
+    Similar to topoological ordering but more robust.
+
+    This is achieved by setting up a workqueue which starts with root of graph.
+    we then explore successors of nodes in workqueue and compute it's scheduling
+    weight as defined in above formula. If ancestor/producer to current node has
+    no value yet, we move current node to end of queue to try again later.
+    """
+    schedule_weight = {}
+    root_nodes = get_root_nodes_from_edges(edges)
```
harsh-nod, line 405: Should we treat all root nodes identically or is there a priority among them?

iree/turbine/kernel/wave/scheduling/graph_utils.py
```diff
@@ -362,3 +362,80 @@ def create_scheduling_edges(
...
+    workqueue = deque(root_nodes)
+    non_solved_counter = 0
+    while len(workqueue) > 0:
+        node = workqueue.popleft()
+        is_producer_edge = lambda edge: edge._to == node
+        producers_edges = filter_edges(is_producer_edge, edges)
+        filter_producer_edge = [
```
harsh-nod, line 412: is this to ignore the accumulator nodes?

iree/turbine/kernel/wave/scheduling/graph_utils.py
```diff
@@ -362,3 +362,80 @@ def create_scheduling_edges(
...
+
+def sort_graph_by_edge_weight(nodes: list[fx.Node], edges: list[Edge]):
+    """
+    Sort nodes based on scheduling weight.
+
+    Where scheduling weight is defined as:
+    scheduling_weight[node] = sum(edge.weight + scheduling_weight[edge.source] for edge in edges)
+
+    scheduling weight is suppose to quantify how early or late a node should live in a graph.
+    Similar to topoological ordering but more robust.
```
harsh-nod, line 397: why is this more robust?

iree/turbine/kernel/wave/scheduling/graph_utils.py
```diff
@@ -362,3 +362,80 @@ def create_scheduling_edges(
...
+
+    This is achieved by setting up a workqueue which starts with root of graph.
+    we then explore successors of nodes in workqueue and compute it's scheduling
+    weight as defined in above formula. If ancestor/producer to current node has
+    no value yet, we move current node to end of queue to try again later.
+    """
+    schedule_weight = {}
+    root_nodes = get_root_nodes_from_edges(edges)
+    workqueue = deque(root_nodes)
+    non_solved_counter = 0
```
harsh-nod, line 407: Can we do this with a topological sort?

iree/turbine/kernel/wave/scheduling/prefetch_scheduling.py
```diff
@@ -0,0 +1,132 @@
...
+import math
+
+
+class PrefetchStage(Enum):
+    GLOBAL_LOAD = 0
+    LOCAL_STORE = 1
+    LOCAL_LOAD = 2
+    COMPUTE = 3
+
+    def next(self):
```
harsh-nod, line 21: Could you a comment here?

raikonenfnu: done!

nicolasvasilache (commented):
Nice, We have also seen this work well for matmul and some convs when true multi-buffering is actually harder to extract performance with!

PR #622, opened 2025-03-25
Add a basic jinja custom op for `conv_2d_nhwc_fhwc`

zjgarvey:
This template op is added so that the `boo` convolution kernel IR for some channels-last layouts can be simplified. 

Once this gets merged I'd like to do the following in another PR:

1. Add some input verification to the custom op.
2. Integrate this custom op into boo convolution kernel generation for forward and weight_backward variants.

Later steps: implement custom ops for certain pre-processing steps needed to convert input_backwards convolutions into forward convolutions, and integrate those into boo conv kernel generation too.

### DIFF
--- a/iree/turbine/ops/conv_fwd.py
+++ b/iree/turbine/ops/conv_fwd.py
@@ -1,0 +1,143 @@
+# Copyright 2025 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from typing import Tuple, no_type_check
+import torch
+
+from ..support.ir_imports import (
+    RankedTensorType,
+    IrType,
+)
+
+from ..runtime.op_reg import (
+    CustomOp,
+    KernelBuilder,
+    KernelSelection,
+    def_library,
+    impl_helper,
+)
+
+LIBRARY = def_library("turbine_ops")
+_templates = impl_helper.JinjaTemplateLoader(__name__)
+
+
+def unpack_tensor_type(tensor_type: IrType) -> Tuple[str, str, IrType]:
+    """Unpacks a RankedTensorType into components usually needed for templating.
+
+    Returns:
+        * The stringified asm form.
+        * An "identifier friendly" form of the shape and element type.
+        * The raw element type.
+    """
+    rtt = RankedTensorType(tensor_type)
+    ident = f"{'x'.join([str(dim) if dim >= 0 else 'D' for dim in rtt.shape])}x{rtt.element_type}"
+    return str(rtt), ident, rtt.element_type
+
+
+@CustomOp.register(library=LIBRARY)
+class conv_2d_nhwc_fhwc(CustomOp):
+    """
+    Basic 2-D channels-last forward convolution op.
+
+    Input format  : N x H x W x C.
+    Weight format : F x Hk x Wk x C.
+    Return format : N x Ho x Wo x F, dtype=f32.
+
+    Assumptions:
+        static dims
+        no bias
+        no pad
+        non-quantized
+        equal input and weight floating point dtypes
+        float32 accumulator
+
+    The argument signature is: input tensor, weight tensor, list of stride ints, list of dilation ints.
+    """
+
+    signature = "conv_2d_nhwc_fhwc(Tensor x, Tensor w, int[] s, int[] d) -> (Tensor)"
+
+    @no_type_check
+    def select(self, ksel: KernelSelection):
+        # tensor args
+        x_desc = ksel.arg_tensor(0)
+        x_desc.specialize_all_dims()
+        w_desc = ksel.arg_tensor(1)
+        w_desc.specialize_all_dims()
+        # attr args
+        s_desc = ksel.attr_list_int(2)
+        d_desc = ksel.attr_list_int(3)
+
+        # assume 2d and NHWC for now
+        n, *spatial, c_x = x_desc.t.shape
+        f, *spatial_k, c_k = w_desc.t.shape
+
+        # TODO: check that the input shapes and dtypes are valid
+        # TODO: check that the number of strides and dilations are valid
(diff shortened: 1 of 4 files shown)


iree/turbine/ops/conv_fwd.py
```diff
@@ -0,0 +1,143 @@
...
+        for name, l in attr_dict.items():
+            for i, v in enumerate(l):
+                kwargs[f"{name}{i}"] = v
+        # output dim values
+        for i in range(spatial_dims):
+            v = res_desc.t.shape[i + 1]
+            kwargs[f"OUT_DIM{i}"] = v
+
+        kwargs["conv_op"] = f"linalg.conv_2d_nhwc_fhwc"
+        kwargs["accum_dtype"] = "f32"
```
nirvedhmeshram, line 130: I wonder if we should just use the dtypes of the input as output dtype, is this what torch.convolution would do? I know we want f32 output dtype for bf16/f16 input dtype in certain use cases, but that needs to happen in our lowering, (which will also add the trunc). I believe a pass to do that already exists in torch-mlir. Not sure if its on linalg ops or something else though. But we can add such a pass if needed.

tests/ops/conv_2d_nhwc_fhwc_test.py
```diff
@@ -0,0 +1,57 @@
...
+    [1, 1],
+    [2, 2],
+]
+
+dilations = [
+    [1, 1],
+    [2, 2],
+]
+
+dtype = torch.float32
```
nirvedhmeshram, line 31: nit: unused variable.

iree/turbine/ops/templates/conv_2d_nhwc_fhwc.mlir
```diff
@@ -0,0 +1,43 @@
...
+!X_asm_type = {{X_asm_type}}
+!W_asm_type = {{W_asm_type}}
+// !bias_asm_type = {{B_asm_type}}
+!result_asm_type = {{result_asm_type}}
+!dynamic_result_asm_type = tensor<?x?x?x?x{{accum_dtype}}>
+
+module {
+
+util.func private @conv_2d_nhwc_fhwc_{{spec_sig}}
+  (%input_pad: !X_asm_type, %weights: !W_asm_type)
```
nirvedhmeshram, line 20: should %input_pad be just called %input?

iree/turbine/ops/templates/conv_2d_nhwc_fhwc.mlir
```diff
@@ -0,0 +1,43 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#map0 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+#map1 = affine_map<(d0, d1, d2, d3) -> (d1)>
+
+!accum_type = {{accum_dtype}}
+!X_asm_type = {{X_asm_type}}
+!W_asm_type = {{W_asm_type}}
+// !bias_asm_type = {{B_asm_type}}
```
nirvedhmeshram, line 13: not sure if you meant to leave this as I am thinking the bias conv will be a different custom op?

tests/ops/conv_2d_nhwc_fhwc_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+device = torch.device("cuda:0") if torch.cuda.is_available() else None
+
+
+@pytest.mark.parametrize("s", strides)
+@pytest.mark.parametrize("d", dilations)
+def testCustomConvImplementationEager(s, d):
+    gen = torch.Generator(device=device)
+    gen.manual_seed(10)
+    x = torch.randn([N, H, W, C], generator=gen, dtype=torch.float32, device=device)
```
nirvedhmeshram, line 41: if you decide to take the output dtype suggestion, then would be good to add tests for different types, bf16, I32 perhaps?

nirvedhmeshram:
Looking at a recent PR, the promotion of type happens at torch-to-linalg lowering, so I guess we cant use the same infrastructure in custom ops
https://github.com/llvm/torch-mlir/pull/3924/files
Feel free to ignore the dtype comment for now, maybe we will have to make the custom op logic smarter as a next step to infer the defaultaccType like what the linked PR is doing.

raikonenfnu:
Any thoughts on moving this inside iree/turbine/kernel/boo? I know there are pre-existing directory here but if it all goes to boo, maybe will makes more sense to put it in one centralized easy to find place?

iree/turbine/ops/conv_fwd.py
zjgarvey: Yeah, my intention was to add a trunc after calling this if we wanted to get back to a f16 type. I think there should be a default accumulator type mapping somewhere in torch, but I couldn't find it immediately.

iree/turbine/ops/templates/conv_2d_nhwc_fhwc.mlir
zjgarvey: Agreed. I thought about adding the bias-add step in this custom op, but I think we can make this custom op more atomic and allow authoring the bias-add afterwards.

iree/turbine/ops/conv_fwd.py
zjgarvey: This mapping <https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/AccumulateType.h> appears to also be device-specific, so I think it will make sense to keep this custom op simple for now and leave `f32` accumulator type as an assumption of the op. 

We can modify this in subsequent patches if the need arises.

nirvedhmeshram: There is `getDefaultAccType` in torch-mlir we could copy that logic here? But fine with making that the next step though.

nirvedhmeshram (approved):
LGTM

iree/turbine/ops/conv_fwd.py
zjgarvey: Yeah, I'd like to try and land this now and iterate as needed.

PR #623, opened 2025-03-26
[TKW] Add `ThreadConstraint`

Hardcode84:
* Add `ThreadConstraint`, allowing to distribute dim across threads inside workgroup, but not depend on wg index.
* `ThreadConstraint` can be combined with `WaveConstraint` in the same way as `WorkgroupConstraint`.
* Add `DistributionConstraint` base class and refactor masking generation.
* Refactor `test_toy_online_softmax` and fix it for unaligned shapes while we at here.


Hardcode84:
closing in favor of https://github.com/iree-org/iree-turbine/pull/644

Issue #620, opened 2025-03-25
One testEvoformerAttentionForward test case failing on mi250 CI


ScottTodd:
Test case: `testEvoformerAttentionForward[dtype0-MMAType.F32_32x32x8_F16-no_sched-tile_sizes0-1x256x256x4x32x256x32]` seems to be failing 100% of the time on `Unit Tests and Type Checking (3.11, nodai-amdgpu-mi250-x86-64)` for the last two weeks.

Logs from one run:
* https://github.com/iree-org/iree-turbine/actions/runs/13865434393/job/38906713550#step:10:107
* https://github.com/iree-org/iree-turbine/actions/runs/13865434393/job/38906713550#step:10:2150

```
>           assert (
                torch.max(torch.abs(torch_ref - output)).item() < eps
            ), f"out eps: {torch.max(torch.abs(torch_ref - output))}"
E           AssertionError: out eps: 1.7998046875
E           assert 1.7998046875 < 0.01
E            +  where 1.7998046875 = <built-in method item of Tensor object at 0x7f755005ed50>()
E            +    where <built-in method item of Tensor object at 0x7f755005ed50> = tensor(1.7998, device='cuda:0', dtype=torch.float16).item
E            +      where tensor(1.7998, device='cuda:0', dtype=torch.float16) = <built-in method max of type object at 0x7f7ac3afc240>(tensor([[[[[0.0000e+00, 2.4414e-04, 2.4414e-04,  ...
```

Failing since at least https://github.com/iree-org/iree-turbine/commit/9f0a536d166885e4d65b79a0299a640db43dbb23.

Closest actual code change to the failure is https://github.com/iree-org/iree-turbine/commit/e63443faf09d896ddd2fe92e61ff5b4f092c996d, with this IREE diff: https://github.com/iree-org/iree/compare/iree-3.3.0rc20250312...iree-3.3.0rc20250313. I've queued up a re-run on that commit here: https://github.com/iree-org/iree-turbine/actions/runs/13859171934/job/39385242200, though tests did pass on the associated PR prior to merging.

ScottTodd:
Re-run at that IREE bump failed: https://github.com/iree-org/iree-turbine/actions/runs/13859171934/job/39389965084#step:10:2150. Triggered a run at the parent commit to check if the regression was due to the code change or something else (runner, unpinned packages, etc.): https://github.com/iree-org/iree-turbine/actions/runs/13844311314/job/39392133288

PR #624, opened 2025-03-26
[Wave] temporarily turn off test for non default paths.

raikonenfnu:
failing on CI for non default path:
1. evoformer on MI250
2. C++ wave runtime
Signed-off-by: Stanley Winata <stanley.winata@amd.com>


harsh-nod (approved):

ScottTodd:
Looks like this resolves https://github.com/iree-org/iree-turbine/issues/620, at least for now?

PR #625, opened 2025-03-26
[Wave] Build wave runtime using torch pip package

harsh-nod:
Previously, the runtime needed a system wide install of rocm. We remove this constraint and instead use the torch libraries inside the torch pip package. Also, rather than build the wave-runtime during execution, we provide an option to pip install wave-runtime.

### DIFF
--- a/iree/turbine/kernel/wave/runtime/CMakeLists.txt
+++ b/iree/turbine/kernel/wave/runtime/CMakeLists.txt
@@ -22,13 +22,26 @@ find_package(nanobind CONFIG REQUIRED)
 
 # Link against torch, using the libraries and include dirs in site-packages.
 execute_process(
-  COMMAND "${Python_EXECUTABLE}" -c "import torch; print(torch.utils.cmake_prefix_path)"
-  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_cmake_PATH)
-list(APPEND CMAKE_PREFIX_PATH ${torch_cmake_PATH})
-list(APPEND CMAKE_PREFIX_PATH "/opt/rocm")
+  COMMAND "${Python_EXECUTABLE}" -c "import torch; print(torch.__path__[0])"
+  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_PATH)
 
-find_package(Torch REQUIRED)
-find_package(HIP REQUIRED)
+# Add all directories in /opt/rocm/lib/cmake to CMAKE_PREFIX_PATH
+file(GLOB subdirectories "/opt/rocm/lib/cmake/*")
+foreach(subdir ${subdirectories})
+  if(IS_DIRECTORY ${subdir})
+    list(APPEND CMAKE_PREFIX_PATH "${subdir}")
+  endif()
+endforeach()
+
+include_directories(${torch_PATH}/include ${torch_PATH}/include/torch/csrc/api/include/)
+link_directories(${torch_PATH}/lib)
+
+execute_process(
+  COMMAND "${Python_EXECUTABLE}" -c "import torch; print(int(torch._C._GLIBCXX_USE_CXX11_ABI))"
+  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_cxx_abi)
+set(TORCH_CXX_FLAGS -D_GLIBCXX_USE_CXX11_ABI=${torch_cxx_abi})
+
+find_package(hip REQUIRED)
 add_definitions(-D__HIP_PLATFORM_AMD__)
 
 # Build the core parts of nanobind once
@@ -38,8 +51,9 @@ nanobind_build_library(nanobind SHARED)
 add_library(wave_runtime MODULE runtime.cpp)
 
 # .. and link it against the nanobind parts
+set(TORCH_LIBRARIES ${torch_PATH}/lib/libtorch.so ${torch_PATH}/lib/libc10.so)
 target_link_libraries(wave_runtime PRIVATE nanobind hip::host ${TORCH_LIBRARIES})
-include_directories(${TORCH_INCLUDE_DIRS})
+set_target_properties(wave_runtime PROPERTIES LINK_WHAT_YOU_USE TRUE)
 
 # .. enable size optimizations
 nanobind_opt_size(wave_runtime)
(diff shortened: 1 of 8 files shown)


raikonenfnu (approved):

iree/turbine/kernel/wave/runtime/CMakeLists.txt
```diff
@@ -22,13 +22,26 @@ find_package(nanobind CONFIG REQUIRED)
...
-  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_cmake_PATH)
-list(APPEND CMAKE_PREFIX_PATH ${torch_cmake_PATH})
-list(APPEND CMAKE_PREFIX_PATH "/opt/rocm")
+  COMMAND "${Python_EXECUTABLE}" -c "import torch; print(torch.__path__[0])"
+  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_PATH)
 
-find_package(Torch REQUIRED)
-find_package(HIP REQUIRED)
+# Add all directories in /opt/rocm/lib/cmake to CMAKE_PREFIX_PATH
+file(GLOB subdirectories "/opt/rocm/lib/cmake/*")
```
ScottTodd, line 29: ROCm can be installed to other locations (especially coming from projects like https://github.com/ROCm/TheRock, or on Windows instead of Linux). There should be _zero_ hardcoded references to `/opt/rocm/`. The first preference should be the `ROCM_PATH` environment variable.

iree/turbine/kernel/wave/runtime/CMakeLists.txt
```diff
@@ -22,13 +22,26 @@ find_package(nanobind CONFIG REQUIRED)
...
+
+include_directories(${torch_PATH}/include ${torch_PATH}/include/torch/csrc/api/include/)
+link_directories(${torch_PATH}/lib)
+
+execute_process(
+  COMMAND "${Python_EXECUTABLE}" -c "import torch; print(int(torch._C._GLIBCXX_USE_CXX11_ABI))"
+  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_cxx_abi)
+set(TORCH_CXX_FLAGS -D_GLIBCXX_USE_CXX11_ABI=${torch_cxx_abi})
+
+find_package(hip REQUIRED)
```
ScottTodd, line 44: I'd feel a lot more comfortable with this sort of code if it was optional behind a setting like `WAVE_RUNTIME_BACKENDS` (that could only support `HIP` for now, but could be generalized in the future) that had error handling with good error messages for when package discovery fails. As written now, this assumes a lot about the system that this code is running on and fails without a clear error if those assumptions are invalid.

iree/turbine/kernel/wave/runtime/setup.py
```diff
@@ -0,0 +1,57 @@
+from setuptools import setup, Extension
```
ScottTodd, line 1: Source files should have copyright headers (we can discuss shortening the headers, but we should generally be consistent within the repository).

ScottTodd (commented):
Thanks, this takes steps in good directions! I feel like this is tenable if we keep it separated in its own package, but if we want to start building this into more CI/CD pipelines and then want to also publish on pypi then that will need more infrastructure support (e.g. building wheels under manylinux, explicit support policy for operating systems, compute APIs, AMD gfx targets, etc.).

I do wonder if the wave runtime code could somehow lean on the HIP runtime code that the iree-runtime package already links in.

iree/turbine/kernel/wave/runtime/CMakeLists.txt
harsh-nod: thanks, that makes sense.  So how would I access ROCM_PATH? Several clusters on which I am running this do not have that environment variable set.

iree/turbine/kernel/wave/runtime/setup.py
harsh-nod: Sure I will put up a PR with the headers in this file.

iree/turbine/kernel/wave/runtime/CMakeLists.txt
harsh-nod: Right now, we only support HIP and the assumption for this find_package is that we have rocm installed on the system. What kind of error message are you looking for?

iree/turbine/kernel/wave/runtime/CMakeLists.txt
ScottTodd: The pattern I've seen used is to default to the environment variable then fall back to `/opt/rocm` on Linux if the environment variable is not set.

iree/turbine/kernel/wave/runtime/CMakeLists.txt
ScottTodd: Let's be clear about which "we" we're talking about with support. IREE, under the Linux Foundation, supports multiple backends. If this component will only ever support HIP, that should be called out clearly in the docs, naming, samples, etc. It may be time to split wave further from the rest of iree-turbine if a deeper hole is getting dug there. The common point is PyTorch integration, and keeping iree-turbine as the center of mass for that is useful though.

Brainstorming:

* Organize the top level namespaces under https://github.com/iree-org/iree-turbine/tree/main/iree/turbine and relate them back to the high level features of iree-turbine: https://github.com/iree-org/iree-turbine?tab=readme-ov-file#iree-turbine. Add more READMEs to subfolders and structure to https://iree-turbine.readthedocs.io/ reflecting that.
* Add a support table to documentation showing which components are supported where.
* Fork Wave into a new repository that depends on iree-turbine? I'd like to avoid this if we can.

---

As for error messages, here are logs on a Windows system without ROCm/HIP installed trying to follow the setup instructions: https://gist.github.com/ScottTodd/142bc17c53001d92f81683109db09304

iree/turbine/kernel/wave/runtime/CMakeLists.txt
ScottTodd: Here are the official docs: https://rocm.docs.amd.com/en/latest/conceptual/cmake-packages.html . These recommend `find_package(hip REQUIRED)`. I don't think you can avoid a dependency on ROCm/HIP by relying on what torch provides and trusting that this path will exist, though the new PyTorch + ROCm wheels that @stellaraccident is putting together may provide some new options. Until then, it would be more robust to follow the recommendations on that page.

PR #627, opened 2025-03-26
[BOO] enable using `conv_2d_nhwc_fhwc` in conv exports

zjgarvey:
This plugs in the `conv_2d_nhwc_fhwc` into a custom implementation for `ConvSignature.get_nn_module` to avoid generating unnecessary transposes. 

Some small additional fixes:

1. remove a debug print in the custom op generate method
2. fix an issue with disabling the boo_cache
3. adds the func op `preprocessing_pipeline` attr after lowering to iree-input (it appeared to be getting stripped).

Some examples of IR before vs. after these changes for the following signature (+/- `bias=True`) :

```python
ConvSignature(input_shape=[2,14,14,4], kernel_shape=[3,3,3,4], shared_layout="NHWC", dtype=torch.bfloat16)
```

are contained in this gist: <https://gist.github.com/zjgarvey/7d66d8ec6dc4470b22dd8329ed9e7e0a>

### DIFF
--- a/iree/turbine/kernel/boo/conv_exports/conv.py
+++ b/iree/turbine/kernel/boo/conv_exports/conv.py
@@ -263,6 +263,8 @@ class ConvSignature:
             str(self.dtype).removeprefix("torch."),
             str(self.mode).lower(),
         ]
+        if self.bias and self.mode == Mode.FORWARD:
+            name_items.append("_b_")
         l2s = lambda l: "x".join([str(i) for i in l])
         name_items.extend(
             [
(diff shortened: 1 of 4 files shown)


MaheshRavishankar:
What happens with this IR https://gist.github.com/zjgarvey/7d66d8ec6dc4470b22dd8329ed9e7e0a#file-non_custom_bias-mlir . Does it compile?

zjgarvey:
> What happens with this IR https://gist.github.com/zjgarvey/7d66d8ec6dc4470b22dd8329ed9e7e0a#file-non_custom_bias-mlir . Does it compile?

Yeah, it compiles.

rsuderman (approved):

Issue #392, opened 2025-01-16
[TKW] Feature Request: waves_per_block dimensionality should match workgroups


GMNGeoffrey:
The `waves_per_block` argument to `tkw.HardwareConstraint` currently has to be of length 3. Wave supports logical workgroup distribution along more than 3 dimensions and then figures out the mapping later. It should do the same with `waves_per_block`. Related to this, there are several sharp edges with the current setup:

1. If you have a workgroup dimension with block size other than 1 at index 3 or greater, the Wave compiler tries to look up `waves_per_block` at that index and fails with an IndexError
2. If you try to make `waves_per_block` longer than 3, the program crashes (looks like a segfault on the C side or something).
3. You have to specify `waves_per_block` for dimensions that you're not distributing along.

FYI @harsh-nod

PR #632, opened 2025-03-27
[wave] Make WaveConstraint implicit and internal to tkw

fabianmcg:
### TL;DR;

Due to the nature of the wave programming model and its current implementation, given a `WorkGroupConstraint` and `waves_per_block`, these fully determine a uniquely valid `WaveConstraint`. Therefore, this patch transforms `WaveConstraint` into `ImplicitWaveConstraint` and infers this constraint from the `WorkGroupConstraint` and `waves_per_block`.

### Motivation:

Currently the usage of `WaveConstraint` by an user can introduce many subtle bugs. For example, if `waves_per_block=(2, 2, 1)` and we add the following constraints:
```python
WaveConstraint(M, BLOCK_M)
WaveConstraint(N, BLOCK_N)
```
the resulting code will likely cause a memory fault error. Another example, would be  `waves_per_block=(1, 1, 1)` with no `WaveConstraint`, which results in wrong results.

As mentioned above, the root of the issue is that given the model and implementation there's only one valid `WaveConstraint` for a WorkGroupConstraint` and `waves_per_block` pair.

This change, should improve the user experience, as it reduces the complexity of understanding how to specify the constraint.

### DIFF
--- a/iree/turbine/kernel/wave/constraints.py
+++ b/iree/turbine/kernel/wave/constraints.py
@@ -80,6 +80,10 @@ class Constraint(ABC):
     def apply(self) -> IndexSequence:
         """Apply the constraint and get the resulting index sequence."""
         ...
+    
+    def is_implicit(self) -> bool:
+        """Returns whether this is an implicit constraint internal to tkw."""
+        return False
 
 
 @dataclass
@@ -471,9 +475,9 @@ class TilingConstraint(Constraint):
 
 
 @dataclass
-class WaveConstraint(Constraint):
+class ImplicitWaveConstraint(Constraint):
     """
-    A constraint of the form `tkw.WaveConstraint(K, WAVE_K)` specifies
+    A constraint of the form `tkw.ImplicitWaveConstraint(K, WAVE_K)` specifies
     that we want distribute the K dimension among multiple waves which
     each wave operating on a tile size of WAVE_K. The assumption is
     that the K dimension has already been distributed among workgroups.
@@ -496,6 +500,9 @@ class WaveConstraint(Constraint):
     wave_id_0 = floor(thread_id_0 / threads_per_wave)
     wave_id_1 = thread_id_1
     wave_id_2 = thread_id_2
+
+    Important, this constraint is internal to TKW and cannot be specified by
+    a user.
     """
 
     dim: IndexExpr
@@ -506,6 +513,9 @@ class WaveConstraint(Constraint):
         if self.wave_id is None:
             raise ValueError("Index is being computed without setting wave id")
         return IndexSequence(self.tile_size * self.wave_id, 1)
+    
+    def is_implicit(self) -> bool:
+        return True
 
     def set_wave_id_from_hardware_and_workgroup_constraint(
         self,
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -27,7 +27,7 @@ from .constraints import (
     Constraint,
     HardwareConstraint,
     TilingConstraint,
-    WaveConstraint,
+    ImplicitWaveConstraint,
     WorkgroupConstraint,
     get_grid_shape,
 )
@@ -204,6 +204,18 @@ class LaunchableWave(Launchable):
     ):
         super().__init__(eager_function)
 
+        assert (
+            constraints is None
+            or next(
+                (
+                    c
+                    for c in constraints
+                    if isinstance(c, Constraint) and c.is_implicit()
+                ),
+                None,
+            )
+            is None
+        ), "found an implicit constraint in the constraint set."
         self.constraints = constraints if constraints else []
         self.induction_vars: dict[CustomOp, IndexExpr] = {}
         self._name = name
@@ -229,11 +241,11 @@ class LaunchableWave(Launchable):
         ]
 
     @property
-    def wave_constraints(self) -> list[WaveConstraint]:
+    def wave_constraints(self) -> list[ImplicitWaveConstraint]:
         return [
             constraint
             for constraint in self.constraints
-            if isinstance(constraint, WaveConstraint)
+            if isinstance(constraint, ImplicitWaveConstraint)
         ]
 
     @property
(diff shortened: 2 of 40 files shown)


GMNGeoffrey:
> Another example, would be waves_per_block=(1, 1, 1) with no WaveConstraint, which results in wrong results.

Can you elaborate on this? I've been using this by default as the simplest configuration

This seems sort of related to https://github.com/iree-org/iree-turbine/issues/392. Since `waves_per_block` is currently the one that doesn't behave, I wonder if it would be better to make that one implicit instead? It's also order dependent in a sort of annoying way and is stuck in `HardwareConstraint`, which I don't totally get. OTOH, I think specifying the number of waves in a direction is simpler and more ergonomic than specifying the size of the sub-block wave grouping, especially since the latter is basically always done by taking `BLOCK_X` and dividing it by waves in that direction. Perhaps then the best solution to all of these is a new constraint like:

`WaveConstraint(M, 2)` -> 2 waves in the M dimension. Of course, it would probably need a different name to avoid confusion... Maybe `WaveCountConstraint(M, 2)`?

iree/turbine/kernel/wave/constraints.py
```diff
@@ -80,6 +80,10 @@ class Constraint(ABC):
     def apply(self) -> IndexSequence:
         """Apply the constraint and get the resulting index sequence."""
         ...
+    
+    def is_implicit(self) -> bool:
```
raikonenfnu, line 84: NIT: Can we set this with @property decorator

lit_tests/kernel/wave/index_sequence_analysis.py
```diff
@@ -198,13 +196,13 @@ def test_gemm():
         # CHECK-NEXT: placeholder(_name=b
         # CHECK-NEXT: placeholder(_name=c
         # CHECK-NEXT: register
-        # CHECK-SAME: index={M: $WG0*BLOCK_M + 4*floor((Mod($T0, 64))/16) : 4 : 16, N: $WG1*BLOCK_N + BLOCK_N/2 + Mod($T0, 16) : 1 : 1})
```
raikonenfnu, line 201: Can we double check why we need to change the LIT test at all? If the implicitWaveConstraint is indeed 1:1 functionality with the WaveConstraint, we shouldn't be changing these

iree/turbine/kernel/wave/wave.py
```diff
@@ -294,21 +306,41 @@ def is_reduction(node: fx.Node):
...
-            for workgroup_constraint in self.workgroup_constraints:
-                if wave_constraint.dim == workgroup_constraint.dim:
-                    wave_constraint.set_wave_id_from_hardware_and_workgroup_constraint(
-                        hardware_constraint, workgroup_constraint
-                    )
+        constraints = []
+        for workgroup_constraint in self.workgroup_constraints:
+            # Ignore `workgroup_dim > 2` and dimensions with only one thread (ie. the value is always 0).
+            if (
+                workgroup_constraint.workgroup_dim > 2
```
raikonenfnu, line 322: Styling NIT: I'd recommend for this massive if condition to be broken apart for better legibility

raikonenfnu: Once, we break apart the if(s), I'd also add a comment wrt `workgroup_dim !=0 and waves_per_block == 1` with something simlar to https://github.com/iree-org/iree-turbine/blob/de8f92ebcf86058338efe5833d4d863a3a636e44/iree/turbine/kernel/wave/constraints.py#L493-L495

iree/turbine/kernel/wave/wave.py
```diff
@@ -204,6 +204,18 @@ def __init__(
     ):
         super().__init__(eager_function)
 
+        assert (
+            constraints is None
+            or next(
+                (
+                    c
+                    for c in constraints
+                    if isinstance(c, Constraint) and c.is_implicit()
```
raikonenfnu, line 213: This is quite verbose, how about:
```
if isinstance(constraints, Sequence):
    assert all([not c.is_implicit for c in constraints if isinstance(c, Constraint)])
```
Are there cases where `constraints == None` or not specified? If not, we should omit that

raikonenfnu:
Upsignalling Geoffrey's message, would be good to think about whether it's more ergonomic and intuitive to specify wave count or the workload_per_wave.

iree/turbine/kernel/wave/wave.py
```diff
@@ -294,21 +306,42 @@ def is_reduction(node: fx.Node):
...
+                    workgroup_constraint.workgroup_dim != 0
+                    and hardware_constraint.waves_per_block[
+                        workgroup_constraint.workgroup_dim
+                    ]
+                    == 1
+                )
+                or not workgroup_constraint.primary
+            ):
+                continue
+            constraint = ImplicitWaveConstraint(
```
nicolasvasilache, line 333: I am a big fan of "explicitly" calling out that this is an implicit constraint and removing the footgun of 2 specifications that should agree but may sometimes not. I didn't dare going this path myself as I expected things to generally break. 

If we can get confirmation that this is indeed the intended semantics this would be fantastic and a really great catch!

Maybe let's phrase it differently, how would we expect the system to behave when the 2 don't agree and what are theoretical use cases? 

I personally have a lot of cases where mapping and number of threads/waves per block disagree but this involves non-trivial predications (up to an including wave-specialization); it seems like a big stretch to support these in TKW but maybe this is the intended future @harsh-amd ?

fabianmcg (commented):
> It is still in question whether or not it's good to completely phase out explicit WaveConstraints or not. I reckon this OK for 99.999% of the cases but would be good to reason a little about this first. i.e is there a case here for power-users who wants to specify their own WaveConstraints.

Right now, the `wave_id` is always set to an specific value during the standard  compilation pipeline, see https://github.com/iree-org/iree-turbine/blob/main/iree/turbine/kernel/wave/wave.py#L305-L311
and
https://github.com/iree-org/iree-turbine/blob/main/iree/turbine/kernel/wave/wave.py#L496

This means, that unless a user constructs the pipeline, no real customization is possible. Which provides another reason to remove this footgun, as it's currently only used in tests. Further, I would argue if customization is required (so far I don't see real need), it'd be better to introduce a new constraint explicitly just for that.

### Why tests changed

There are 2 main reasons why tests changed:

1. Some tests were using `WaveConstraint(..., ..., MyWaveId)`, and with this change I remove this option. Which in turns alters the results of the tests. See above comment for some rationale.
2. I'm skipping adding constraints when `waves_per_block[d] == 1 and d != 0`, as `THREAD_d` is always 0 in that case, so is adding no ops.

### Re: specifying `waves_per_block ` vs `WaveConstraint`

I believe it's better to specify the thread block (specify `waves_per_block`) rather than inferring the thread block from `WaveConstraint`, as it gives the user more control over the hardware mapping. Also, it avoids many potential footguns, for example, a user could easily get confused with `WaveConstraint(M, 2)` vs `WaveConstraint(M, BLOCM_M / 2)`. The first one is likely to cause a miscompilation error, while the second one is more likely to be correct.

>> Another example, would be waves_per_block=(1, 1, 1) with no WaveConstraint, which results in wrong results.

> Can you elaborate on this? I've been using this by default as the simplest configuration

That was an error on the message, it should've been `waves_per_block = (2, 2, 1)`. 

> WaveConstraint(M, 2) -> 2 waves in the M dimension. Of course, it would probably need a different name to avoid confusion... Maybe WaveCountConstraint(M, 2)?

Also, I would rather move to less constraints than more from a front-user perspective.

harsh-nod:
Thanks for this PR and bringing this issue to attention. I do agree that there is redundancy between WaveConstraint and waves_per_block that can lead to incorrect behavior and that we should do away with one of these. Out of the two, I think it makes more sense to get rid of the waves_per_block (this was the idea originally but fell through the cracks) for multiple reasons. Geoffrey made some good points about it being more ergonomic but aside from that it, one point that came up was that the WaveConstraint specification is a superset of the waves_per_block specification.

Say, we want to distribute distribute dimension M, among workgroups with a tile size of BLOCK_M and then further among waves with a tile size of BLOCK_M_WAVE, then the waves per block would be ceildiv(BLOCK_M, BLOCK_M_WAVE). Since ceildiv is non-invertible, multiple inputs map to the same output. Hence, there would be no way to deduce the tile size given a waves_per_block specification. But with a WaveConstraint tile size specification, we can always deduce the waves_per_block.

GMNGeoffrey:
I think https://github.com/iree-org/iree-turbine/issues/643 is also sort of related. There is surprising implicit behavior in `threads_per_block` and `WaveConstraint`. But I think it's also an example of where you want `WaveConstraint` and not `waves_per_block` because in this case I really want the wave to be distributed 8x8, which is still 1 wave per block but in a different way.

fabianmcg:
In the end I;m working on removing `waves_per_block`, so I'm closing this one.

PR #645, opened 2025-03-28
[Wave] Add missing header

harsh-nod:
This PR adds the copyright header to the wave runtime CMakeLists.txt.


ScottTodd (approved):
Lots of files are missing header comments, so I'd rather do a project-wide fix (and add them before merge), instead of sending individual PRs file by file. This is fine though, thanks.

harsh-nod:
I was just addressing your comments from the last PR. Good point on files missing headers.  Will add an issue to track that.

Issue #643, opened 2025-03-28
[Wave] All threads are distributed along first workgroup (x) dimension -> surprising behavior for 2d kernels


GMNGeoffrey:
I've got a really simple 2D copy kernel that I think should work

<details><summary>kernel</summary>

```python
def get_copy2d_kernel(dim_m: int, dim_n: int, block_m: int, block_n: int):
    threads_per_wave = 64
    elements_per_block = block_m * block_n
    assert elements_per_block >= threads_per_wave
    elements_per_thread = elements_per_block // threads_per_wave

    M = tkl.sym.M
    N = tkl.sym.N

    BLOCK_N = tkl.sym.BLOCK_N
    BLOCK_M = tkl.sym.BLOCK_M

    constraints: list[tkw.Constraint] = [
        tkw.WorkgroupConstraint(M, BLOCK_M, 0),
        tkw.WorkgroupConstraint(N, BLOCK_N, 1),
        tkw.HardwareConstraint(
            threads_per_wave=64,
            waves_per_block=(1, 1, 1),
            vector_shapes={M: block_m, N: block_n},
        ),
    ]

    @tkw.wave(constraints)
    def copy2d(
        a: tkl.Memory[M, N, GLOBAL_ADDRESS_SPACE, tkl.i32],
        b: tkl.Memory[M, N, GLOBAL_ADDRESS_SPACE, tkl.i32],
    ):
        a_reg = tkw.read(a, elements_per_thread=elements_per_thread)
        tkw.write(a_reg, b, elements_per_thread=elements_per_thread)

    hyperparams = {
        BLOCK_M: block_m,
        BLOCK_N: block_n,
        M: dim_m,
        N: dim_n,
    }

    return copy2d, hyperparams
```

</details>

but Wave appears to be putting all 64 threads along the 0th/x dimension. I'm calling this with dim_m = dim_n = block_m = block_n = 8 and get the following IR:

<details><summary>MLIR</summary>

```mlir
#translation = #iree_codegen.translation_info<pipeline = None workgroup_size = [64, 1, 1] subgroup_size = 64, {llvm_func_attrs = {"amdgpu-waves-per-eu" = "2", "denormal-fp-math-f32" = "preserve-sign"}}>
module attributes {transform.with_named_sequence} {
  stream.executable private @copy2d {
    stream.executable.export public @copy2d workgroups() -> (index, index, index) {
      %c1 = arith.constant 1 : index
[...]

PR #648, opened 2025-03-28
[Wave] Make it possible for waves to be distributed in multiple dimensions

GMNGeoffrey:
Right now, threads within a wave are always distributed so that there are 64 threads along the x dimension. This creates confusing behavior for kernels that are naturally multi-dimensional. I've removed this hardcoding and instead made that the *default* behavior with the ability for the user to specify other distributions. I've included tests of multi-dimensional copies that weren't possible otherwise.

Fixes https://github.com/iree-org/iree-turbine/issues/643

### DIFF
--- a/tests/kernel/wave/wave_e2e_test.py
+++ b/tests/kernel/wave/wave_e2e_test.py
@@ -25,12 +25,15 @@ from iree.turbine.kernel.wave.utils.torch_utils import (
     device_randint,
     device_randperm,
     device_zeros,
+    get_default_device,
 )
 from .common.utils import (
     require_e2e,
     require_cdna3,
     perf_test,
     param_bool,
+    maybe_dump_mlir,
+    format_shape,
 )
 import torch
 from torch.testing import assert_close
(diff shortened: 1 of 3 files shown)


GMNGeoffrey:
FYI @fabianmcg this might intersect with what you're working on. I'd be happy for this to be expressible as wave constraints instead.

raikonenfnu (commented):
This is a slightly controversial (not necessarily a bad thing) PR because thread_per_block would can conflict with waves_per_block and is changing the programming model/view a little bit. 

I do see value in being able to do multi dimension as seen in your copy_2d example, but not sure if this is the best interface (not saying it's not but would be good to also explore and think about what would be most intuitive for users as well).

raikonenfnu:
a potentially less controversial way is probably to specify `threads_per_wave` as 3D as well, since that is a little bit more aligned with rest of the interface and the wave centric nature of Wave programming language

GMNGeoffrey:
Yeah I figured this would require some discussion, so I mostly just wanted to provide the motivating test case and then made it work with the easiest change possible :-) I thought about making `threads_per_wave` multi-dimensional as well. I was a bit hesitant though because that one's currently simple and is one of the few things in the `HardwareConstraint` struct that's actually a constraint of the hardware. It kind of seems like some unrelated stuff has been accumulated in there. I also wondered if we might want to make this more in the format of `WaveConstraint` and `WorkgroupConstraint`. This would perhaps help with removing the 3D restriction that's not consistently applied (#392).

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -193,6 +196,225 @@ def test(
...
+    BLOCK_M = tkl.sym.BLOCK_M
+
+    constraints: list[tkw.Constraint] = [
+        tkw.WorkgroupConstraint(M, BLOCK_M, 0),
+        tkw.WorkgroupConstraint(N, BLOCK_N, 1),
+        tkw.HardwareConstraint(
+            threads_per_wave=threads_per_wave,
+            waves_per_block=(1, 1, 1),
+            threads_per_block=(threads_per_block_m, threads_per_block_n, 1),
+            vector_shapes={M: block_m, N: block_n},
```
nicolasvasilache, line 299: Side note: I keep on shooting myself in the foot with the `vector_shape` terminology, as I expect it to be the vector_shape.

If at all possible, I would prefer if we could rename to something much more explicit like: "per_wave_computation_shape".

GMNGeoffrey: Yeah this also confuses me every time...

GMNGeoffrey:
> Fly-by comment: I think this is better but this still subject to footguns. The only convention I have been able to reliably work with is to go to the linear thread id.
> 
> 3-D threads per block -> linearized to 1-D -> delinearized to 3-D warps per block. Also works well with ranks != 3-D which is often necessary to properly specify e.g 5-D iteration spaces and let all the div/mod machinery work naturally.
> 
> I do not know how to expose this in TKW properly atm.

Thanks :-) @fabianmcg is going to take this over and incorporate into his work in https://github.com/iree-org/iree-turbine/pull/661. I'm not totally sure I follow your suggestion, but I agree that the 3D hardcoding is unfortunate in the current implementation, especially since Wave gives the impression of supporting arbitrary dimensionality.

Fabian, see Nico's comments for input into what you're working on.

PR #649, opened 2025-03-29
[Wave][Cache] Add caching support for nested functions

raikonenfnu:
Currently, we are caching nested function as freevars which is detected in getclosurevars. However we are only caching it's function pointer which means every time we have a nested function inside wave kernel, it will recompile every single time since it has a different function pointer.

Even worse, the kernel defintion that we have cached for these nested functions are often just something like:
```
    @tkw.wave(constraints)
    def base_attention(
        q: tkl.Memory[B, N_Q, D_Q, GLOBAL_ADDRESS_SPACE, LOGIT_DTYPE],
        k: tkl.Memory[B, N_KV, D_Q, ADDRESS_SPACE, LOGIT_DTYPE],
        v: tkl.Memory[B, N_KV, D_KV, ADDRESS_SPACE, LOGIT_DTYPE],
        c: tkl.Memory[B, N_Q, D_KV, GLOBAL_ADDRESS_SPACE, tkl.f32],
    ):
        base_attention_core(q, k, v, c)
```

Which is very incomplete since we do not know what is inside base_attention_core and could change between different functions.

To solve this, we added a simple BFS search that detects all subfunctions, and then we just accumulate all these function's generated key to compute the final key. This way we can solve/cache all sorts of nested function.

### DIFF
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -96,7 +96,7 @@ def get_nested_functions(root_fn: Callable):
     """Simple BFS search to get all sub functions inside a wave kernel."""
     workqueue = deque([root_fn])
     fn_list = set([root_fn])
-    while len(workqueue) > 0:
+    while workqueue:
         cur_fn = workqueue.pop()
         # Add var to workqueue and fn_list freevar that
         # we have not seen before and who's type is a function.
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -15,7 +15,7 @@ import shutil
 import threading
 import math
 
-from collections import OrderedDict
+from collections import OrderedDict, deque
 from dataclasses import dataclass, asdict
 from pathlib import Path
 import functools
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/wave/cache.py
```diff
@@ -88,10 +88,28 @@ def extract_free_vars(kernel_fn: Callable):
...
-        if not isinstance(v, IndexMapping)
+        if not isinstance(v, (IndexMapping, Callable))
     ]
 
 
+def get_nested_functions(root_fn: Callable):
+    """Simple BFS search to get all sub functions inside a wave kernel."""
+    workqueue = deque([root_fn])
+    fn_list = set([root_fn])
+    while len(workqueue) > 0:
```
Hardcode84, line 99: nit: `while workqueue:`

Hardcode84 (approved):
nice

iree/turbine/kernel/wave/cache.py
raikonenfnu: ooo nice! didn't know we could do that

raikonenfnu: done!

Hardcode84: yeah, all containers treated as `False` when empty and `True` otherwise, when casted to bool.

PR #650, opened 2025-03-29
[TKW] Make elements-per-thread optional

Hardcode84:


iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -573,6 +579,12 @@ def get_index(custom: CustomOp):
             return custom.acc.index
         return custom.index
 
+    for node in trace.walk(lambda x: isinstance(get_custom(x), (Read, Write))):
+        custom = get_custom(node)
+        if custom.elements_per_thread is None:
```
raikonenfnu, line 584: This is so cool!

raikonenfnu:
also kind of interesting that `Read` already have it as optional

raikonenfnu:
@Hardcode84 can we do the same for the `templates/vanilla_attentionn.py`, I think if we have that, we can be more confident that it's fully safe and can land :)

raikonenfnu (approved):
super awesome stuff man!

PR #658, opened 2025-03-29
[TKW] Infer `elements_per_thread` for elementwise ops

Hardcode84:
* Infer `elements_per_thread` for elementwise ops from tile size and num threads
* Some lit tests are not updated as they have suspicious `elements_per_thread` hardcoded

### DIFF
--- a/iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
+++ b/iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
@@ -317,16 +317,19 @@ def populate_read_write_source_indices(
             else node.elements_per_thread
         )
 
-        stride = compute_stride(
-            node.indexing_dims, hardware_constraint.vector_shapes, dim
-        )
         wg_constraint = [x for x in workgroup_constraints if x.dim == dim]
 
         assert len(wg_constraint) <= 1, f"Multiple workgroup constraints for dim {dim}"
         if not wg_constraint:
             continue
 
+        stride = compute_stride(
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -316,19 +317,23 @@ def populate_read_write_source_indices(
...
         wg_constraint = [x for x in workgroup_constraints if x.dim == dim]
+
+        assert len(wg_constraint) <= 1, f"Multiple workgroup constraints for dim {dim}"
         if not wg_constraint:
             continue
+
+        wg_dim = wg_constraint[0].workgroup_dim
+        if elements_per_thread is None:
+            tile_size = wg_constraint[0].tile_size
+            threads_count = hardware_constraint.threads_per_block[wg_dim]
```
harsh-nod, line 332: Right now, we allow workgroup_dims > 2, wouldnt that create a problem here when we try to access threads_per_block > 2? Maybe add an assert with a TODO that we should allow threads_per_block > 2 ?

harsh-nod (approved):
lgtm! thanks for cleaning this up :) Just 1 request for an assert and should be good to land.

iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
Hardcode84: Good idea, done

PR #661, opened 2025-03-31
[wave] Remove `waves_per_block` in favor of `thread_block`

fabianmcg:
### TL;DR;

Due to the nature of the wave programming model and its current implementation, given a `WorkGroupConstraint` and `waves_per_block`, these fully determine a uniquely valid `WaveConstraint`. Therefore, this patch removes `waves_per_block` and adds `thread_block` that can be inferred from `WorkGroupConstraint` and `WaveConstraint`.

### Motivation:

Currently the usage of `waves_per_block` by an user can introduce many subtle bugs. For example, if `waves_per_block=(2, 2, 1)` and we add the following constraints:
```python
WaveConstraint(M, BLOCK_M)
WaveConstraint(N, BLOCK_N)
```
the resulting code will likely cause a memory fault error. 

As mentioned above, the root of the issue is that given the model and implementation there's only one valid `waves_per_block` for a `WorkGroupConstraint` and `WaveConstraint` pair.

This change, should improve the user experience, as it reduces the complexity of understanding how to specify the constraint.

### Explanation:

[NFC] This patch moves `index_utils` to `_support.indexing`, as they make more sense there and allows `lang` to not depend in `wave` for those utils.

To better represent GPU hardware semantics, this patch replaces `waves_per_block` in favor of `thread_block`, which has the traditional GPU definition. This field is now optional as it can be inferred from `WorkGroupConstraint` and `WaveConstraint`. To maintain customizability, the field can also be specified. Currently, `ThreadBlock` will produce an error if `thread_block.x` is not divisible by `threads_per_wave`, as this patch preserves the current wave semantics.
In a future patch this could be expanded.


Hardcode84:
Maybe a good idea to keep `waves_per_block` for this PR even if unused so we don't need to update all tests at once and can keep changes smaller. Then we can do test cleanup and remove it in in follow PRs.

fabianmcg:
> Maybe a good idea to keep `waves_per_block` for this PR even if unused so we don't need to update all tests at once and can keep changes smaller.

I'd argue not cleaning the tests at once introduces the possibility of creating an inconsistent state in the repo.
Is it possible to do stack PR's? I'm guessing that the answer is no because of lack of rights. Also, in the end it's pretty similar to a single PR.

But, I can change the current PR.

raikonenfnu:
> > Maybe a good idea to keep `waves_per_block` for this PR even if unused so we don't need to update all tests at once and can keep changes smaller.
> 
> I'd argue not cleaning the tests at once introduces the possibility of creating an inconsistent state in the repo. Is it possible to do stack PR's? I'm guessing that the answer is no because of lack of rights. Also, in the end it's pretty similar to a single PR.
> 
> But, I can change the current PR.

Thanks Fabian, another reason why we should do this in a separate PR is introducing `thread_block` in lieu of `wave_per_block` deserves an actual RFC since it is modifying the way we conceptualize things.

fabianmcg:
> deserves an actual RFC since it is modifying the way we conceptualize things.
Thank you for the info @raikonenfnu ! 
Where should I post the RFC?

raikonenfnu:
> > deserves an actual RFC since it is modifying the way we conceptualize things.
> 
> Thank you for the info @raikonenfnu ! Where should I post the RFC?

A github issue on iree-turbine, but if you already have the code, could be a standalone PR

PR #666, opened 2025-04-01
[TKW] Use affine apply

Hardcode84:
* Add ability to use `affine.apply` for index codegen
* Limited to non-vector index type for now, need https://github.com/llvm/llvm-project/pull/129442 for vector support

### DIFF
--- a/iree/turbine/kernel/wave/codegen/emitter.py
+++ b/iree/turbine/kernel/wave/codegen/emitter.py
@@ -204,12 +204,8 @@ _Rational = namedtuple("_Rational", ["numerator", "denominator"])
 _ApplyExpr = namedtuple("_ApplyExpr", ["expr", "args"])
 
 
-def gen_sympy_index(
-    dynamics: dict[IndexSymbol, Value],
-    expr: sympy.Expr,
-    *,
-    use_affine_expr: bool = _use_affine_expr,
-) -> Value:
+def gen_sympy_index(dynamics: dict[IndexSymbol, Value], expr: sympy.Expr) -> Value:
+    use_affine_expr = _use_affine_expr
     stack: list[OpResult] = []
 
     def _get_ir_value(arg):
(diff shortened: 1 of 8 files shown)


iree/turbine/kernel/wave/codegen/emitter.py
```diff
@@ -282,70 +326,123 @@ def addi(lhs, rhs):
...
+
+    def muli_expr(lhs, rhs):
+        if not use_affine_expr:
+            return muli(*_broadcast(lhs, rhs))
+
+        return op_expr(lhs, rhs, lambda a, b: a * b)
+
+    def rem_expr(lhs, rhs):
+        if not use_affine_expr:
+            return
```
Hardcode84, line 362: missing expr

iree/turbine/kernel/wave/codegen/emitter.py
```diff
@@ -193,55 +197,95 @@ def add_emitter_subs(
     return dynamics
 
 
-_emulate_ceildiv = True
+_emulate_ceildiv = bool(int(environ.get("WAVE_EMULATE_CEILDIV", 1)))
+_use_affine_expr = bool(int(environ.get("WAVE_USE_AFFINE_EXPR", 0)))
 
 _Rational = namedtuple("_Rational", ["numerator", "denominator"])
+_ApplyExpr = namedtuple("_ApplyExpr", ["expr", "args"])
```
harsh-nod, line 204: Could you instead make this a class and in the class verify that all the args are of type index? If they are of type index, you can return an ApplyExpr, otherwise you can return the value. This way the parts that can leverage affine apply will and others won't and you won't need to throw an exception.

Hardcode84: fixed in different way

Hardcode84:
Switched `affine.apply` on by default

harsh-nod (approved):
lgtm! thanks for this, should help debugging!

nicolasvasilache:
nice, much easier to read indeed, thanks!

PR #668, opened 2025-04-01
[TKW] Disable ceildiv emulation

Hardcode84:
* It was used to workaround upstream issues, but looks like they are fixed
* Disable it but keep the code around for the future.


harsh-nod (approved):
thanks!

PR #669, opened 2025-04-01
[Wave] Implement fused bias-add on Wave kernel Linear Layer

raikonenfnu:
- Implement Wave kernel for linear with bias add by fusing it into the init_reg for the mma accumulator.
- Refactoring function getter to enable both bias and non-bias
- Remove element per thread from kernel since PR landed to not require explicit setting of that anymore.


nithinsubbiah (approved):
Looks good!

PR #670, opened 2025-04-01
[BOO] add an autograd function for conv launchables

zjgarvey:
This is some initial work on being able to "plug-in" a boo conv launchable into a `torch.nn.Module` so that running the model backward will call into corresponding backward conv kernels. 

1. Adds a user-friendly `autograd.Function` called `boo_conv` which essentially acts like a regular forward torch convolution, but generates a launchable and invokes it.
2. Adds a runtime cache of conv launchables so that `boo_conv` can re-use the same launchable (if the configuration matches a previous use). 
3. Adds a `torch.nn.Module` variant of `Conv2d`, called `BooConv2d`, and a helper function which will replace each `Conv2d` named_module in a model into an equivalent `BooConv2d`. 
4. Adds a very basic training example for resnet-18 with the boo convs swapped in. This is not currently functional as the input_backward convolution IR does not currently compile (as a single dispatch) for some of the convolution configs present in the model, but this may be useful as a litmus test for basic functionality. 
5. Fixes a bug with non-list kwargs in `ConvSignature` getting removed.

Note: It may be better to eventually allow defining backward implementations for `CustomOp`s instead.

### DIFF
--- a/iree/turbine/kernel/boo/modeling/__init__.py
+++ b/iree/turbine/kernel/boo/modeling/__init__.py
@@ -1,0 +1,1 @@
+from .replace import *
--- a/iree/turbine/kernel/boo/ops/__init__.py
+++ b/iree/turbine/kernel/boo/ops/__init__.py
@@ -1,0 +1,1 @@
+from .conv import *
--- a/iree/turbine/kernel/boo/conv_exports/conv.py
+++ b/iree/turbine/kernel/boo/conv_exports/conv.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from typing import (
+    Any,
     List,
     NamedTuple,
     Optional,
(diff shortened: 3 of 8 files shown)


iree/turbine/kernel/boo/ops/__init__.py
```diff
@@ -0,0 +1 @@
+from .conv import *
```
rsuderman, line 1: Remember to include the copyright header

```
# Copyright 2025 The IREE Authors
#
# Licensed under the Apache License v2.0 with LLVM Exceptions.
# See https://llvm.org/LICENSE.txt for license information.
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

iree/turbine/kernel/boo/modeling/__init__.py
```diff
@@ -0,0 +1 @@
+from .replace import *
```
rsuderman, line 1: Remember to include copyright header:

```
# Copyright 2025 The IREE Authors
#
# Licensed under the Apache License v2.0 with LLVM Exceptions.
# See https://llvm.org/LICENSE.txt for license information.
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

rsuderman (approved):

PR #671, opened 2025-04-02
[TKW] Skip slow tests

Hardcode84:
* Print slowest tests on CI and skip the worst offenders (some of them were taking 300+sec).
* Reduced mi300 test time 18min -> 11min
* Let's keep individual tests on CI under 1min, everything else goes into `expensive_tests`

### DIFF
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -95,15 +95,15 @@ jobs:
           rm -rf ./.wave
           WAVE_CACHE_ON=1 pytest --capture=tee-sys -vv --run-e2e --durations=100 ./tests/kernel/wave/runtime
 
-      - name: Run e2e tests on AMD GPU MI300
-        if: "contains(matrix.os, 'mi300') && !cancelled()"
+      - name: Run e2e tests on AMD GPU
+        if: "(contains(toJSON(matrix.os), 'amdgpu') || contains(toJSON(matrix.os), 'mi300')) && (github.event_name == 'pull_request') && !cancelled()"
         run: |
-          WAVE_CACHE_ON=0 pytest -n 4 --capture=tee-sys -vv --run-e2e --gpu-distribute 1 --durations=100 ./tests/kernel/wave/
+          WAVE_CACHE_ON=0 pytest -n 4 --capture=tee-sys -vv --run-e2e --durations=100 ./tests/kernel/wave/
 
-      - name: Run e2e tests on AMD GPU MI250
-        if: "contains(matrix.os, 'mi250') && !cancelled()"
+      - name: Run expensive e2e tests on AMD GPU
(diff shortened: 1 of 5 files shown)


harsh-nod:
Just trying to understand, is this only for benchmarking the tests? Or are you proposing skipping the expensive tests?

Hardcode84:
@ScottTodd we are using the same `ci-tk.yaml` both for the PRs and when pushed to main but I want to run expensive tests (`--run-expensive-tests`) only when merged to main but not on PRs. Any ideas how to handle it in yaml without duplicating?

Hardcode84:
Merged mi300 and mi250 steps (the only real difference was `-n` value) and choosing between expensive and non-expensive pipeline using `github.event_name == 'pull_request'`.

@ScottTodd can you take a look on GHA part?

ScottTodd (approved):
Seems reasonable enough. I generally prefer having separate workflows for different sets of tests, so you can customize the triggers and choose which to run with workflow_dispatch though. If there is too much boilerplate for forking like that to be reasonable... reduce the boilerplate by simplifying install steps, move steps into utility actions, etc.

Hardcode84:
I will merge is this state for now to speed up the tests.
The biggest boilerplate as for now is this part:
```
      - name: Install pip deps
        if: "(!contains(toJSON(matrix.os), 'amdgpu') && !contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
        run: |
          python -m pip install --no-compile --upgrade pip
          # Note: We install in three steps in order to satisfy requirements
          # from non default locations first. Installing the PyTorch CPU
          pip install --no-compile -r pytorch-cpu-requirements.txt
          pip install -r requirements.txt -e .

      - name: Install GPU rocm and pip deps (mi300)
        if: "(!contains(toJSON(matrix.os), 'amdgpu') && contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
        run: |
          sudo apt install -y rocm
          python -m pip install --no-compile --upgrade pip
          # Note: We install in three steps in order to satisfy requirements
          # from non default locations first. Installing the PyTorch CPU
          pip install -r requirements.txt -e .
          pip install -r requirements-wave-runtime.txt -e .

      - name: Install pip deps (mi250)
        if: "(contains(toJSON(matrix.os), 'amdgpu') && !contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
        run: |
          python -m pip install --no-compile --upgrade pip
          # Note: We install in three steps in order to satisfy requirements
          # from non default locations first. Installing the PyTorch CPU
          pip install -r requirements.txt -e .
```
Some way to simplify it and move to common place would be nice.
Also, as a side note, with test times reduced to ~10min, these additional 5min we are taking to install packages are quite noticeable now, some way to speed it up would also be nice.

PR #674, opened 2025-04-02
[Wave] Use reciprocal to compute softcap logits

harsh-nod:
This PR uses reciprocals instead of divs to
compute the logits faster.


raikonenfnu (approved):

raikonenfnu:
lgtm but not sure if we need to update lit

raikonenfnu:
lgtm but not sure if we need to update lit

harsh-nod:
> lgtm but not sure if we need to update lit

thanks, just updated it

raikonenfnu (approved):

PR #675, opened 2025-04-02
Capture location information from Python

ftynse:
Representing and propagating file/line location information from Python kernel source significantly improves debugging experience.

Note that column information is only available since Python 3.11 so additional logic is added to handle older versions.


Hardcode84:
Nice!

Is there anything else to do here, as you didn't added any reviewers?
(also, please rebase)

ftynse:
I am not allowed to add reviewers...

Hardcode84 (approved):

PR #677, opened 2025-04-02
Added support for scalar codegen

xintin:
Added support for scalar codegen.
Added lit tests for i32, f32 scalars
Tested on `wave` runtime
Manually tested using `iree-run-module` and `iree-compile`

### DIFF
--- a/iree/turbine/kernel/compiler/host_codegen.py
+++ b/iree/turbine/kernel/compiler/host_codegen.py
@@ -32,7 +32,7 @@ def memref_to_tensor(memrefs: list[IrType]):
     tensors = []
     for m in memrefs:
         # append scalars as-it-is to tensors list
-        if isinstance(m, (F32Type, IntegerType)):
+        if isinstance(m, F32Type) or (isinstance(m, IntegerType) and m.is_signless):
             tensors.append(m)
             continue
         assert isinstance(m, MemRefType)
--- a/iree/turbine/kernel/compiler/host_codegen.py
+++ b/iree/turbine/kernel/compiler/host_codegen.py
@@ -84,7 +84,7 @@ def isolated_test_call(
         func_op = func_d.FuncOp("isolated_benchmark", ftype)
         actual_loc = FileLineColInfo.capture_current_location().to_mlir()
         arg_locs = [
-            (Location.name(b.name) if b.name is not None else Location.unknown())
+            (Location.name(b.name, actual_loc) if b.name is not None else actual_loc)
             for b in sig.kernel_buffer_bindings
             + sig.scalar_bindings
             + sig.dynamic_dim_bindings
--- a/iree/turbine/kernel/compiler/host_codegen.py
+++ b/iree/turbine/kernel/compiler/host_codegen.py
@@ -18,7 +18,9 @@ from .ir import (
     RankedTensorType,
     flow_d,
     func_d,
+    F32Type,
     IntegerAttr,
+    IntegerType,
 )
 
 from .._support.indexing import IndexSymbol
@@ -29,6 +31,10 @@ from .kernel_codegen import BindingDesc
 def memref_to_tensor(memrefs: list[IrType]):
     tensors = []
     for m in memrefs:
+        # append scalars as-it-is to tensors list
+        if isinstance(m, (F32Type, IntegerType)):
+            tensors.append(m)
+            continue
         assert isinstance(m, MemRefType)
         t = RankedTensorType.get(m.shape, m.element_type)
         tensors.append(t)
@@ -56,7 +62,9 @@ def isolated_test_call(
     dynamic_symbols: list[IndexSymbol] = [],
 ):
     with InsertionPoint(mb.body_block), Location.unknown():
-        input_types = [b.as_mlir_type() for b in sig.kernel_buffer_bindings]
+        input_types = [b.as_mlir_type() for b in sig.kernel_buffer_bindings] + [
+            b.as_mlir_type() for b in sig.scalar_bindings
+        ]
         input_tensors = memref_to_tensor(input_types)
         argument_dims = get_dynamic_dims(sig.kernel_buffer_bindings, dynamic_symbols)
         # Adding unique dynamic dims as inputs.
@@ -76,8 +84,10 @@ def isolated_test_call(
         func_op = func_d.FuncOp("isolated_benchmark", ftype)
         actual_loc = FileLineColInfo.capture_current_location().to_mlir()
         arg_locs = [
-            (Location.name(b.name, actual_loc) if b.name is not None else actual_loc)
-            for b in sig.kernel_buffer_bindings + sig.dynamic_dim_bindings
+            (Location.name(b.name) if b.name is not None else Location.unknown())
+            for b in sig.kernel_buffer_bindings
+            + sig.scalar_bindings
+            + sig.dynamic_dim_bindings
         ]
         entry_block = func_op.add_entry_block(arg_locs)
         offset = len(sig.kernel_buffer_bindings)
--- a/iree/turbine/kernel/compiler/kernel_codegen.py
+++ b/iree/turbine/kernel/compiler/kernel_codegen.py
@@ -322,8 +322,6 @@ class KernelSignature:
         for node in placeholder_nodes:
             index = None
             for i, binding in enumerate(self.bindings):
-                # if binding.binding_type == BindingType.SCALAR_VALUE:
-                #     break
                 if binding.reference[1] == node:
                     index = i
                     break
--- a/iree/turbine/kernel/compiler/kernel_codegen.py
+++ b/iree/turbine/kernel/compiler/kernel_codegen.py
@@ -22,6 +22,8 @@ from dataclasses import dataclass
 import torch.fx as fx
 import sympy
 
+from iree.turbine.kernel._support.dtype import DataType
+
 from .._support.indexing import (
     IndexingContext,
     IndexSymbol,
@@ -47,9 +49,11 @@ from .builder import (
 from .ir import (
     Block,
     FunctionType,
+    F32Type,
     IndexType,
     InsertionPoint,
     IrType,
+    IntegerType,
     Location,
     Operation,
     Value,
@@ -78,6 +82,7 @@ class BindingType(Enum):
     KERNEL_BUFFER = 0
     INDEX_VALUE = 1
     SYMBOL_VALUE = 2
+    SCALAR_VALUE = 3
 
 
 def _is_symbolic(value: list[sympy.Expr | int]) -> bool:
@@ -109,6 +114,9 @@ class BindingDesc:
 
     # If a SYMBOL_VALUE, then this is the corresponding IndexSymbol.
     symbol_type: Optional[Type[IndexSymbol]] = None
+
+    # If there is a scalar, then this is the corresponding type.
+    scalar_type: Optional[Type[DataType]] = None
 
     def as_mlir_type(self) -> IrType:
         idx_context = IndexingContext.current()
@@ -152,6 +160,12 @@ class BindingDesc:
             return IndexType.get()
         elif binding_type == BindingType.SYMBOL_VALUE:
             return IndexType.get()
+        elif (
+            binding_type == BindingType.SCALAR_VALUE and self.scalar_type.is_float_asm()
+        ):
+            return F32Type.get()
+        elif binding_type == BindingType.SCALAR_VALUE and self.scalar_type.is_int_asm():
+            return IntegerType.get_signless(32)
         else:
             raise AssertionError("Unhandled switch BindingType")
 
@@ -205,6 +219,11 @@ class KernelSignature:
         """Gets all dynamic dimension bindings."""
         return [b for b in self.bindings if b.binding_type == BindingType.SYMBOL_VALUE]
 
+    @property
+    def scalar_bindings(self) -> list[BindingDesc]:
+        """Gets all scalar bindings."""
+        return [b for b in self.bindings if b.binding_type == BindingType.SCALAR_VALUE]
+
     def add_from_dynamic_symbols(self, dynamic_symbols: list[IndexSymbol]):
         for symbol in dynamic_symbols:
             self.bindings.append(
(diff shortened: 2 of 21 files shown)


iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -65,22 +68,51 @@ int launch(const KernelLaunchInfo &info, const Int64Vector &tensors, const Int64
             cache[info.kernelHash] = std::make_tuple(module, function);
     }
 
+    ScalarKind scalarKind;
+    size_t scalarSize = 4;
```
harsh-nod, line 72: instead of hardcoding the size, can you please do a sizeof?

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -65,22 +68,51 @@ int launch(const KernelLaunchInfo &info, const Int64Vector &tensors, const Int64
...
+
     // Since we always pass our dynamic dims as index type, iree converts them to i64
     // and then splits them to two i32s, i64 = hi | lo where
     // lo = trunc(i64) and hi = trunc(i64 >> 32).
-    size_t kernArgSize = tensors.size() * sizeof(uint64_t) + 2 * dynamicDims.size() * sizeof(uint32_t);
+    size_t kernArgSize = tensors.size() * sizeof(uint64_t) + 2 * dynamicDims.size() * sizeof(uint32_t) + scalarArgs.size() * scalarSize;
+    std::vector<uint8_t> scalarBytes(scalarArgs.size() * scalarSize);
+
+    for (size_t i = 0; i < scalarArgs.size(); ++i) {
+        if (scalarKind == ScalarKind::F32) {
```
harsh-nod, line 94: could be simplified using sizeof

xintin: done

iree/turbine/kernel/compiler/kernel_codegen.py
```diff
@@ -294,6 +322,8 @@ def any_write_dependencies(node):
         for node in placeholder_nodes:
             index = None
             for i, binding in enumerate(self.bindings):
+                # if binding.binding_type == BindingType.SCALAR_VALUE:
```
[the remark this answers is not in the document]
xintin: removed.

xintin:
After [latest commit ](33641a265dcd5524f0622e73533f5152394fec01), this PR will require some rework.

iree/turbine/kernel/wave/cache.py
```diff
@@ -70,6 +72,7 @@ def extract_arg_types(kernel_fn: Callable):
     return [
         (arg.annotation, arg.annotation.physical_layout)
         for arg in inspect.signature(kernel_fn).parameters.values()
+        if not isinstance(arg.annotation, DataType)
```
raikonenfnu, line 75: can we generalize this some more, can you check what is the expected arg.annotation type and check for `if isinstance(arg.annotation, <TargetType>)`

iree/turbine/kernel/wave/compile.py
```diff
@@ -32,16 +34,26 @@ def invoke(self, *args, **kwargs):
...
+        tensor_args = tuple(tensor_args)
+        scalar_args = tuple(scalar_args)
+
+        kernel_inputs, kernel_outputs = [], []
         # Partition arguments into kernel inputs and outputs.
-        kernel_inputs = []
-        kernel_outputs = []
-        for arg, usage in zip(args, self.options.kernel_usages):
+        for arg, usage in zip(tensor_args, self.options.kernel_usages):
             if usage == kernel_codegen.KernelBufferUsage.INPUT:
```
raikonenfnu, line 49: Any thoughts on combinining these two loops together? and just have a special case here if it's a scalar we add to scalar_args instead of kernel_inputs. Then at the end we just set kerneL_inputs.extend(scalar_args)

raikonenfnu: loop is kind of expensive in python, so ideally we don't bring extra there

iree/turbine/kernel/wave/utils/run_utils.py
```diff
@@ -109,13 +109,18 @@ def push_tensor_to_arg_list(arg_tensor: torch.Tensor):
...
             raise ValueError(f"Unsupported input type: {type(input)}")
     for output in outputs:
         if isinstance(output, torch.Tensor):
             push_tensor_to_arg_list(output)
         else:
             raise ValueError(f"Unsupported output type: {type(output)}")
+    # we want scalars to be at the end during codegen/dispatch to iree
+    # to maintain the consistency.
+    for input in inputs:
+        if isinstance(input, float):
```
raikonenfnu, line 122: We should make this a notImplementedError here and a comment on how push_float is currently not working on IREE side

lit_tests/kernel/wave/codegen.py
```diff
@@ -1934,3 +1935,113 @@ def get_item(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
...
+    options = WaveCompileOptions(
+        subs={
+            M: 1,
+            N: 27,
+            ADDRESS_SPACE: tkl.AddressSpace.GLOBAL_MEMORY.value,
+        },
+        canonicalize=True,
+        compile_to_mlir=True,
+    )
+    options = set_default_run_config(options)
```
raikonenfnu, line 1977: do we need default_run_config? seems like other lit test don't need it

lit_tests/kernel/wave/codegen.py
```diff
@@ -1934,3 +1935,113 @@ def get_item(a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16]):
...
```
raikonenfnu, line 2032: ditto

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -65,22 +68,45 @@ int launch(const KernelLaunchInfo &info, const Int64Vector &tensors, const Int64
...
+    if (scalarKind == ScalarKind::F32) {
+        for (auto&& arg : scalarArgs) {
+            float val = nb::cast<float>(arg);
+            std::memcpy(dst++, &val, sizeof(float));
+        }
+    } else {
+        for (auto&& arg : scalarArgs) {
+            *dst++ = static_cast<uint32_t>(nb::cast<int32_t>(arg));
+        }
+    }
```
raikonenfnu, line 95: do we need this specialization (between ScalarKind::F32 VS ScalarKind::I32)? Can we try treating everything as a uint32_t?

raikonenfnu: IIRC, in push constants in IREE, i32 and f32 is treated as uint32_t*

raikonenfnu: e.g
```cpp
for (iree_host_size_t i = 0; i < kernel_params->constant_count; i++) {
    *((uint32_t*)params_ptr[kernel_params->binding_count + i]) =
        ((const uint32_t*)constants.data)[i];
  }
```
https://github.com/iree-org/iree/blob/main/runtime/src/iree/hal/drivers/hip/graph_command_buffer.c#L800-L807

This way we can handle mixed of i32 and f32 inputs

iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -749,11 +757,16 @@ def get_index(custom: CustomOp):
         lhs = get_custom(custom.lhs)
         rhs = get_custom(custom.rhs)
 
-        lhs_index = get_index(lhs)
-        rhs_index = get_index(rhs)
+        lhs_index = get_index(lhs) if lhs.indexing_dims else None
```
raikonenfnu, line 760: How about modifying this inside the `get_index` function itself to reduce redundant code

iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -749,11 +757,16 @@ def get_index(custom: CustomOp):
...
+        rhs_index = get_index(rhs) if rhs.indexing_dims else None
 
-        lhs_dim, lhs_size = get_largest_index_and_size(lhs_index)
-        rhs_dim, rhs_size = get_largest_index_and_size(rhs_index)
+        # Updatedto broadcast implicitly when we perform mul binary op with scalar.
+        lhs_dim, lhs_size = (
+            get_largest_index_and_size(lhs_index) if lhs.indexing_dims else ((), 1)
```
raikonenfnu, line 765: ditto wrt reducing redundant code

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -29,6 +31,10 @@
 def memref_to_tensor(memrefs: list[IrType]):
     tensors = []
     for m in memrefs:
+        # append scalars as-it-is to tensors list
+        if isinstance(m, (F32Type, IntegerType)):
```
raikonenfnu, line 35: any thoughts on replacing `IntegerType` here  with `IntegerType.get_signless(32)`, in case someone feeds non 32 bit integer, we'd want to catch that

raikonenfnu (commented):
Awesome stuff @xintin! Just some refactoring comments and questions

iree/turbine/kernel/wave/cache.py
xintin: Done. Checking for `KernelBufferMeta`

iree/turbine/kernel/wave/compile.py
xintin: The logic was split into two loops because we are using `zip` and `scalar args` do not have `kernel_usages`. 
So, it will lead to incorrect result.

xintin: Addressed

lit_tests/kernel/wave/codegen.py
xintin: It is a mix. In some lit tests, it is used and in some it is not.

lit_tests/kernel/wave/codegen.py
xintin: We actually do not need them. Removed.

iree/turbine/kernel/wave/compile.py
```diff
@@ -32,15 +34,25 @@ def invoke(self, *args, **kwargs):
...
-        kernel_outputs = []
-        for arg, usage in zip(args, self.options.kernel_usages):
-            if usage == kernel_codegen.KernelBufferUsage.INPUT:
-                kernel_inputs.append(arg)
-
-            if usage == kernel_codegen.KernelBufferUsage.OUTPUT:
-                kernel_outputs.append(arg)
+        usage_idx = 0
+        for arg in args:
+            if isinstance(arg, torch.Tensor):
```
raikonenfnu, line 44: NIT 1:  tensor_args is no longer required 
NIT 2: I think IREE/MLIR style convention is try to do early exit/minimize nested, i.e, we can do something like:
```py
usage_idx = 0
        for arg in args:
            if not isinstance(arg, torch.Tensor):
               scalar_args.append(arg)
               continue
            usage = self.options.kernel_usages[usage_idx]
            usage_idx += 1
            if usage == kernel_codegen.KernelBufferUsage.INPUT:
              kernel_inputs.append(arg)
            if usage == kernel_codegen.KernelBufferUsage.OUTPUT:
              kernel_outputs.append(arg)
```

raikonenfnu: would the scalar inputs not have `self.options.kernel_usages`?

raikonenfnu: I see because it's `BindingType.SCALAR_VALUE` it doesn't work out of the box

raikonenfnu: perhaps could be good to make a function called `kernel_signature_bindings` that takes generate binding for all inputs/outputs that is part of the kernel signature. That way modification to this part will be very minimal

iree/turbine/kernel/wave/runtime/runtime.cpp
xintin: Yes, you are right. I tested it and it works.

iree/turbine/kernel/wave/compile.py
xintin: I agree with the early stopping but we may not want to add another function call as it is not used in many places. Also, it will be same as adding one more loop.

iree/turbine/kernel/wave/runtime/runtime.cpp
xintin: It is throwing `bad_cast` for float dtype.

harsh-nod (approved):
lgtm! thanks for adding this functionality.

raikonenfnu (approved):

raikonenfnu:
let's wait for MI300 test to pass before landing

iree/turbine/kernel/wave/compile.py
raikonenfnu: Moved to a TODO, where we need to add `kernel_signature_binding` but also propagate kernelbufferusage/scalartypeusage to the bindingdesc

lit_tests/kernel/wave/codegen.py
```diff
@@ -13,6 +13,7 @@
 from iree.turbine.kernel.wave.utils.compile_utils import (
     set_default_compile_config,
 )
+from iree.turbine.kernel.wave.utils.run_utils import set_default_run_config
```
raikonenfnu, line 16: we can remove this unused import

iree/turbine/kernel/wave/utils/run_utils.py
```diff
@@ -109,13 +109,21 @@ def push_tensor_to_arg_list(arg_tensor: torch.Tensor):
     for input in inputs:
         if isinstance(input, torch.Tensor):
             push_tensor_to_arg_list(input)
-        else:
+        elif not isinstance(input, float):
```
raikonenfnu, line 112: should this allow int as well?

iree/turbine/kernel/wave/cache.py
```diff
@@ -21,6 +21,9 @@
 import functools
 from typing import Any, Callable, Optional
 
+from iree.turbine.kernel._support.dtype import DataType
```
raikonenfnu, line 24: unused import?

iree/turbine/kernel/wave/utils/run_utils.py
xintin: resolved.

raikonenfnu:
Seems like still failing on ```This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101``` 

CC: @saienduri

saienduri:
Ah @xintin this is the place you need to upgrade: https://github.com/iree-org/iree-turbine/pull/677/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR80

PR #678, opened 2025-04-02
Add MOE kernel - Step 1

harsh-nod:
Currently exposes a corner case where `resolve_threads_shapes` chokes on the result of an `tkw.reduction` with a `tkw.mma` when followed by a binary.
The indexing contains quantities such as
```
'$WG0*BLOCK_B + BLOCK_B*floor($T0/64) + Piecewise((Mod($T0, 16), ~$MMA_ACC), (4*floor((Mod($T0, 64))/16), $MMA_ACC)) : Piecewise((1, ~$MMA_ACC), (4, $MMA_ACC)) : Piecewise((1, ~$MMA_ACC), (16, $MMA_ACC))'
```
and the `MMA_ACC` is not resolved, resulting in a comparison error btw static constants and Sympy expressions.

### DIFF
--- a/lit_tests/kernel/wave/moe.py
+++ b/lit_tests/kernel/wave/moe.py
@@ -1,0 +1,397 @@
+# RUN: python %s | FileCheck %s
+
+import copy
+import logging
+from typing import Sequence
+
+import torch
+import torch.nn.functional as F
+
+import iree.turbine.kernel as tk
+import iree.turbine.kernel.lang as tkl
+import iree.turbine.kernel.wave as tkw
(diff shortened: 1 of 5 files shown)


lit_tests/kernel/wave/moe.py
```diff
@@ -0,0 +1,397 @@
...
+)
+from iree.turbine.kernel.wave.utils.run_utils import (
+    set_default_run_config,
+)
+
+DOC = """
+# MoE TCTK
+This is a TCTK version of the MoE kernel. It is looking at the problem from first principles.
+
+Let's have fun with this one
```
harsh-nod, line 44: love this :)

PR #679, opened 2025-04-02
Feat: better scaling for fp8 quant

Giuseppe5:
Currently, there is a line that shifts the tensors before softmax quantization, however that is effectively a no-op.

With this change, the value quantized to FP8 are correctly rescaled from the range [0, 1] to [0, 240] before quantization, thus using more the fp8 value range.

We need to undo the rescaling at the end before the output is returned.

cc @raikonenfnu @KyleHerndon

### DIFF
--- a/iree/turbine/kernel/wave/templates/quantized_attention.py
+++ b/iree/turbine/kernel/wave/templates/quantized_attention.py
@@ -113,7 +113,7 @@ def get_brevitas_pertensor_fp8_attention_kernel(
     # full fp8 range. We can do this with a offset as post `exp2` this equates
     # to multiplying by a static value. We are able to do this as `max` and
     # `sum` are scaled by the same value so the end result is the same.
-    FP8_OFFSET_VAL = math.log2(F8_MAX/ATTENTION_SOFTMAX_MAX)
+    FP8_OFFSET_VAL = math.log2(F8_MAX / ATTENTION_SOFTMAX_MAX)
 
     # Dequant Tensor Scaling
     DEQUANT_QK = q_scale * k_scale
@@ -128,7 +128,7 @@ def get_brevitas_pertensor_fp8_attention_kernel(
     def base_attention_core(q, k, v, c):
         qk_scaling = tkl.Register[B, N_Q, N_KV, tkl.f32](DK_SQRT * LOG2E * DEQUANT_QK)
         v_dequant = tkl.Register[B, D_KV, N_Q, tkl.f32](v_scale)
-        fp8_offset = tkl.Register[B, N_Q, N_KV, tkl.f32](1/FP8_OFFSET_VAL)
+        fp8_offset = tkl.Register[B, N_Q, N_KV, tkl.f32](1 / FP8_OFFSET_VAL)
         fp8_max = tkl.Register[B, N_Q, N_KV, tkl.f32](F8_MAX)
         c_reg = tkl.Register[B, D_KV, N_Q, tkl.f32](0.0)
         init_sum = tkl.Register[B, N_Q, tkl.f32](0.0)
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/wave/templates/quantized_attention.py
```diff
@@ -168,13 +169,12 @@ def repeat(
...
             x_j = x_j + bias
             x_j *= qk_scaling
-            x_j += fp8_offset
             m_j = tkw.max(x_j, partial_max, dim=N_KV)
             e_delta_max = tkw.exp2(partial_max - m_j)
             e_delta = tkw.exp2(x_j - m_j)
             e_init = partial_sum * e_delta_max
             d_j = tkw.sum(e_delta, e_init, dim=N_KV)
-            imm_f8 = low_precision_clamp(e_delta, fp8_max)
+            imm_f8 = low_precision_clamp(e_delta/fp8_offset, fp8_max)
```
raikonenfnu, line 177: this is going to sound silly, but can we do:
```py
            imm_f8 = low_precision_clamp(e_delta * tkw.reciprocal(fp8_offset), fp8_max)
```

having direct div is really bad for performance

raikonenfnu:
Overall PR looks good, but a couple housekeeping things:

1. we'd need to do `pre-commit run --all-files` to make sure it passes the pre-commit check
2. we'd probably want to update the LIT tests if applicable in https://github.com/iree-org/iree-turbine/blob/main/lit_tests/kernel/wave/attention/quantized_attention.py if applicable

iree/turbine/kernel/wave/templates/quantized_attention.py
raikonenfnu: the log2 addition makes sense, but it's not obvious to me how it allows us to make the  post scaling redundant?

Let fp8_offset = ATTENTION_SOFTMAX_MAX / F8_MAX
```py
fp8_offset = tkl.Register[B, N_Q, N_KV, tkl.f32](fp8_offset )
e_delta = tkw.exp2(x_j - m_j)
imm_f8 = low_precision_clamp(e_delta/fp8_offset, fp8_max)
```

-> 
```py
1/FP8_OFFSET_VAL
fp8_offset = tkl.Register[B, N_Q, N_KV, tkl.f32](1/FP8_OFFSET_VAL)
e_delta = tkw.exp2(x_j - m_j + fp8_offset)
imm_f8 = low_precision_clamp(e_delta/fp8_offset, fp8_max)
```

raikonenfnu: Actually, seems like current version is doing `tkw.exp2(x_j - m_j + 1 / math.log2(F8_MAX/ATTENTION_SOFTMAX_MAX))`, so not too sure how we are doing that either actually 😅

iree/turbine/kernel/wave/templates/quantized_attention.py
```diff
@@ -187,7 +186,7 @@ def repeat(
         # repeat represents the results of the loop
         res_max, res_sum, res_mm = repeat
         reciprocal_sum = tkw.reciprocal(res_sum)
-        res = res_mm * reciprocal_sum * v_dequant * fp8_dequant
```
raikonenfnu: we can remove the fp8_dequant because we are now applying /fp8_offset into e_delta -> d_j (denominator) as well as e_delta -> mma(numerator) => we no longer need to do `fp8_dequant`.

iree/turbine/kernel/wave/templates/quantized_attention.py
raikonenfnu: the fp8_dequant question is resolved offline:
```
we can remove the fp8_dequant because we are now applying /fp8_offset into e_delta -> d_j (denominator) as well as e_delta -> mma(numerator) => we no longer need to do fp8_dequant.
```

raikonenfnu: > Actually, seems like current version is doing tkw.exp2(x_j - m_j + 1 / math.log2(F8_MAX/ATTENTION_SOFTMAX_MAX)), so not too sure how we are doing that either actually 😅

Resolved offline, it needs to be ` tkw.exp2(x_j - m_j + math.log2(F8_MAX/ATTENTION_SOFTMAX_MAX))`

raikonenfnu:
overall LGTM, I rebased and ran test, will land once tests passes :)

raikonenfnu:
This is super awesome stuff Giuseppe!

raikonenfnu (approved):

Issue #410, opened 2025-01-23
Index mismatch between op and users creates multiple problems


GMNGeoffrey:
I've run into a few issues that occur when the index for an op and the ops that use it differ. I think these come up due to my workarounds for https://github.com/iree-org/iree-turbine/issues/381 and https://github.com/iree-org/iree-turbine/issues/384. I haven't come up with a minimal reproducer yet, but it shows up in my kernel for the backward pass of flash attention 2 (specifically when computing dk, in this case).

https://github.com/GMNGeoffrey/iree-turbine/blob/4f18a93def1f487363cb403509144b9b89c2aa4f/tests/kernel/wave/attention/backward_attention_test.py#L872-L897

You'll see I've got two permutes in there that just permute an mma output to the shape it already was. In the first case this resolves an error complaining that it can only resolve thread shape discrepancies when one of the shapes is 1:

```
E   NotImplementedError: Currently only support resolving discrepancies when one of the shapes is 1.
E   binary_op=sub
E   lhs=CastOp(graph=<torch.fx.graph.Graph object at 0x74f10134b2e0>, fx_node=cast, tkw_op_name='cast', _tracing_function=<bound method define_op.<locals>.decorator.<locals>.new_function of ...>, arg=mma, dtype=DataType(f16))
E   lhs_index={B: BLOCK_B*(Mod($WG2, 2)) : 1 : 1, M: $ARGM*BLOCK_M + 4*floor((Mod($T0, 64))/16) : 4 : 16, K1: $WG1*BLOCK_K1 + 4*floor((Mod($T0, 64))/16) : 4 : 1, K2: $WG0*BLOCK_K2 + Mod($T0, 16) : 1 : 1}
E   lhs_dim=K1
E   lhs_size=4
E   lhs.type.symbolic_shape=(B, M, K2)
E   rhs=Read(graph=<torch.fx.graph.Graph object at 0x74f10134b2e0>, fx_node=read_2, tkw_op_name='read', _tracing_function=<bound method define_op.<locals>.decorator.<locals>.new_function of ...>, memory=lse, elements_per_thread=MFMA_OUTPUT_ELS_PER_THREAD, mapping=None, mapping_dynamic_vals=(), _write_dependency=None)
E   rhs_index={B: BLOCK_B*(Mod($WG2, 2)) : 1 : 1, M: $ARGM*BLOCK_M + 4*floor((Mod($T0, 64))/16) : 4 : 16}
E   rhs_dim=M
[...]

PR #680, opened 2025-04-03
[Wave] Add a GEMM-bias add example

harsh-nod:
This PR adds a GEMM bias example and fixes some issues relating to index propagation in this case.


GMNGeoffrey:
Perhaps related to #410?

harsh-nod:
> Perhaps related to #410?

Not sure, but you can try it out to see if it helps.

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1496,20 +1467,19 @@ def outputs(self, graph: fx.Graph) -> list[fx.Node]:
 
     @property
     def index(self) -> list[dict[IndexSymbol, IndexSequence]]:
+        def get_index(node: fx.Node) -> dict[IndexSymbol, IndexSequence]:
+            custom = get_custom(node)
+            if isinstance(custom, MMA):
+                return custom.acc_index
```
nicolasvasilache, line 1473: Resurfacing my question from slack here: why do we need 2 different fields `acc_index` and `index` and risk them getting out of sync /  ?

harsh-nod: index here contains the index of the LHS, RHS and ACC nodes altogether as a sympy piecewise and allows one to query the index of the LHS, RHS or ACC from a single source of truth by calling a "specialize_index" function on the index. The reason for constructing this "unified" index was that rather than propagate 3 separate indices to nodes in the graph, it was simpler to propagate a single index and have each node then specialize the index as required. I see the potential sources of error here so open to suggestions.

iree/turbine/kernel/wave/analysis/partition_strided_operators.py
```diff
@@ -50,9 +50,11 @@ def get_vector_shape(
 def _get_symbolic_shape_and_vector_shapes(
     custom: CustomOp,
 ):
-    register_shape = custom.register_type.symbolic_shape
     vector_shapes = custom.vector_shapes
     memory_shape = custom.memory_type.symbolic_shape
+    if isinstance(custom, Read):
```
nicolasvasilache, line 55: I've started digging into these partition strided operators pass, could you explain what the objective of this function is? (nit: maybe add a doc)
I am unclear why we want the register_shape in certain cases and in other the memory_shape.
Would this be a good place to assert that shapes should agree?
If they don't what is the expected behavior?

PR #684, opened 2025-04-03
[boo_driver] Add `--csv` and `--splat-input-value` options

rkayaith:
This adds some extra options to `boo_driver.py`:
- `--csv <file>`: this outputs timing info in csv format to the specified file
- `--splat-input-value <val>`: this uses splat inputs instead of the default random values

Additionally the script now continues if the traced invocation fails, so that all specified configurations are reported.

I've also documented a workaround I'm using to get tracing working with locally built python bindings. Using this requires a recent commit of IREE: iree-org/iree#20448


nirvedhmeshram (approved):
LGTM

PR #686, opened 2025-04-04
[Wave] Add fast math option

harsh-nod:
This PR adds an option to enable fast math for unary and binary floating point operations.


iree/turbine/kernel/wave/compile_options.py
```diff
@@ -61,6 +61,7 @@ class WaveCompileOptions:
     wave_runtime: str = None
     use_buffer_load_ops: bool = False
     use_buffer_store_ops: bool = False
+    use_fast_math: bool = True
```
Hardcode84, line 64: We shouldn't have them `True` by default as they can wreck numerics.

raikonenfnu (approved):

PR #690, opened 2025-04-07
[boo_driver] Add option to validate results

rkayaith:
Adds a  `--check-result` flag which uses the reference torch model to compare results against the IREE implementation using `torch.testing.assert_close`.

### DIFF
--- a/iree/turbine/kernel/boo/conv_exports/boo_driver.py
+++ b/iree/turbine/kernel/boo/conv_exports/boo_driver.py
@@ -154,7 +154,10 @@ def trace_gpu(func: Callable[[], str]) -> tuple[dict[str, list[int]], str]:
                 try:
                     queue.put(func())
                 except Exception as exc:
-                    queue.put(str(exc))
+                    # Return a short error string for printing. The full error
+                    # and traceback appear on stderr since we 'raise' here.
+                    err_str = str(exc).splitlines()[0]
+                    queue.put(err_str)
                     raise
 
             process = Process(target=proc_fn)
--- a/iree/turbine/kernel/boo/conv_exports/boo_driver.py
+++ b/iree/turbine/kernel/boo/conv_exports/boo_driver.py
@@ -100,6 +100,12 @@ def run(cli_args: Sequence[str]):
         type=int,
         help="use a splat value for inputs (defaults to random values)",
     )
+    parser.add_argument(
+        "--check-result",
+        default=False,
+        action="store_true",
+        help="check result against torch implementation",
+    )
     args = parser.parse_args(cli_args)
     sig = mio.get_signature(args)
     conv = get_launchable(sig)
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/boo/conv_exports/boo_driver.py
```diff
@@ -114,6 +120,11 @@ def run(cli_args: Sequence[str]):
 
     torch.set_printoptions(edgeitems=0)
     print(f">>> {result}")
+    if args.check_result:
+        reference_module = sig.get_nn_module(use_custom=False)
+        expected = reference_module(*conv_args)
+        torch.testing.assert_close(result, expected)
+        print(">>> results are correct")
```
zjgarvey, line 127: I have a few concerns about this:

1. I believe the fwd non-custom nn modules are always safe to use; however, the reference modules are not defined for some backward configurations. Namely, grouped backward convs, which don't have a nice pytorch decomposition.
2. The non-custom nn modules may have different numerics than actual pytorch backward convs. The real reference should ideally be used in those cases.
3. From the initial numerics testing I ran [here](https://gist.github.com/zjgarvey/a5eb1991a040cc31c4793602d9232f9d), we will likely be failing this quite frequently (and not necessarily because our numerics are poor, but because pytorch itself may be inconsistent with their numerics). It would be better to use a comparison function like `torch.all_close` and report the mismatch without raising an `AssertionError`. At the very least, we should also try to compare against the CPU reference result as well, as pytorch CPU vs. GPU numerics might differ substantially.

Give me a minute to wrap up some work and I'll share the current iteration of my local numerics runner with you so we can discuss.

iree/turbine/kernel/boo/conv_exports/boo_driver.py
```diff
@@ -143,7 +154,10 @@ def proc_fn():
                 try:
                     queue.put(func())
                 except Exception as exc:
-                    queue.put(str(exc))
+                    # Return a short error string for printing. The full error
+                    # and traceback appear on stderr since we 'raise' here.
+                    err_str = str(exc).splitlines()[0]
+                    queue.put(err_str)
```
zjgarvey, line 160: Ah I see you catch the exception, but it would be good to get a printout of the absolute and relative error values.

zjgarvey (commented):
I have a few initial reservations about adding this option right now. We will need to rework it quite substantially to get reasonable numerics validation, and I'm not sure the boo driver is the best place to implement that kind of thing (or if we should add another script). 

E.g., if I want to run only numerics validation and not performance, I wouldn't want to run `iter` number of times (or have to remove the `iter` flag from each command in my `commands-file`).

rkayaith:
> E.g., if I want to run only numerics validation and not performance, I wouldn't want to run `iter` number of times (or have to remove the `iter` flag from each command in my `commands-file`).

Additional command-line flags get appended to the flags from the file, so you can do `python boo_driver.py --commands-file ... --iter 1` to override the iteration counts from the file.

But other than that I don't have a strong opinion on whether we should have a separate script, this was just the most straightforward (at least with the current implementation).

rkayaith:
closing this as `numerics_runner` from https://github.com/iree-org/iree-turbine/pull/696 subsumes it

PR #695, opened 2025-04-08
[Wave] Add bias test to quantized linear layer

nithinsubbiah:
- Adds bias test to quantized linear layer
- Quantizes weight tensor during layer instantiation and caches it


Giuseppe5:
The issue with the pre-quantization of the weights is that it is done at initialization time, but then once you update them to match the one from the reference implementation, they are not correctly quantized anymore.

In theory, you would like to have a `requantize_weight` function, and also intercept the `load_state_dict` function so that when a new set of weights is set into the layer, then the requantization process happens as well.
Once you have that, I would change the test to do `wave_linear.load_state_dict(ref_linear.state_dict())`, instead of manually updating the weights.
Alternatively, once you have a `requantize_weight` function, you can call that manually whenever needed.

I am not sure what's causing the numerical issue with the bias, that seems to not be related to quantization

Giuseppe5:
The bias doesn't seem to be correctly initialized in the wave kernel.
If I change the second test to:
```python
wave_output = wave_linear.forward(test_inputs) + ref_linear.linear.bias.data
```
then the absmax error test passes, with a small discrepancy in the rmse error

PR #697, opened 2025-04-08
[TKW] GenericDot mma type

Hardcode84:
Implement MMA through vector dot prod instead of mfma intrinsic.

Each thread reads `k_size` vector along K dimension as position [`lane % out_vec_size, 0]` for the `A` matrix and [`lane, 0]` for the `B`, extracts data using shuffles from `out_vec_size` clusted of adjacent threads and produces `out_vec_size` vector output.


iree/turbine/kernel/wave/constraints.py
```diff
@@ -144,7 +190,11 @@ def mma_matrix_shapes(self, mma_type: Optional[MMAType]) -> tuple[int]:
         # TODO: Eventually the shapes and indices should be provided by a tool
         if mma_type == None:
             mma_type = self.mma_type
+
         match mma_type:
+            # M x N x K
+            case GenericDot():
```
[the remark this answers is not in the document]
Hardcode84: no, I've tried exactly that, and no, it doesn't work: `SyntaxError: name capture 'GenericDot' makes remaining patterns unreachable`

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -273,7 +267,109 @@ def handle_set_symbol(emitter: WaveEmitter, node: fx.Node):
 ###############################################################################
 
 
-def emit_mfma(m: int, n: int, k: int, acc: Value, values: list[Value]):
+def decompose_value(src: Value, bitsize: int) -> list[Value]:
```
harsh-nod, line 270: Going in line with what I said earlier, can we move this packing, unpacking and shuffles further up the compilation pipeline, so maybe in DecomposeReduceOps? The idea being maybe we introduce a pass that decomposes the mma into a reduce op and then we leverage the existing decomposition there and add the packing of shuffles. That would also simplify the code in the emitter.

Hardcode84: This is just a lowering quirk because `gpu.shuffle` AMDGPU lowering only supports 32bits (when the `gpu.shuffle` op itself supports arbitrary int/float/vector types). This really should be done on `gpu.shuffle` -> `llvm` lowering level. I don't think we will get anything useful by bringing it up in the stack and we should just fix lowering upstream.

raikonenfnu (commented):
Any thoughts on reusing the decompose reduction or create a mirror called decompose matvec, since in theory we can do everything in wave/fx and would make the handlers a little simpler?

iree/turbine/kernel/wave/constraints.py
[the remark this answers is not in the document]
raikonenfnu: haha I see, this would have been nice though. Resolved.

iree/turbine/kernel/wave/constraints.py
```diff
@@ -59,6 +59,58 @@ class MMAType(Enum):
...
+    """
+
+    out_vec_size: int = 1
+    k_vec_size: int = 4
+    k_mult: int = 1
+
+    def get_shape(self, threads_per_wave: int) -> tuple[int, int, int]:
+        return (
+            self.out_vec_size,
+            threads_per_wave // self.k_mult,
```
Hardcode84, line 79: FYI: Increasing K dim decreases N dim, so it may not always be beneficial to have big `k_mult`.

iree/turbine/kernel/wave/constraints.py
```diff
@@ -59,6 +59,58 @@ class MMAType(Enum):
...
+            self.k_vec_size * self.k_mult,
+        )  # M x N x K
+
+    def get_index_offset(
+        self, lane: IndexExpr, threads_per_wave: int
+    ) -> tuple[IndexExpr, IndexExpr, IndexExpr]:
+        return [
+            Piecewise((lane % self.out_vec_size, ~MMA_ACC), (0, MMA_ACC)),  # M
```
Hardcode84, line 87: Distributing across M or N dimension is completely symmetrical, but I only Implemented N for now.

harsh-nod:
> Any thoughts on reusing the decompose reduction or create a mirror called decompose matvec, since in theory we can do everything in wave/fx and would make the handlers a little simpler?

+1 to this, we should reuse the decompose reduction.

harsh-nod:
Next steps:
1. Pass to lower mma to reduction when GenericDot layout is used for mma
2. Add handling of parallel dimensions in DecomposeReduceOps

iree/turbine/kernel/wave/constraints.py
harsh-nod: makes sense.

Hardcode84:
> TODO: paralellize for packed shuffle with parallel dimensions as seen in https://gist.github.com/raikonenfnu/1991f9e91c93a7dbc8132723fa61ba1b

The problem
```
    b_thread_data_0 = b[tid*4 + 0, tid:tid+1] : vector<2xf32>
    b_thread_data_1 = b[tid*4 + 1, tid:tid+1] : vector<2xf32>
    b_thread_data_2 = b[tid*4 + 2, tid:tid+1] : vector<2xf32>
    b_thread_data_3 = b[tid*4 + 3, tid:tid+1] : vector<2xf32>
```
Is needed to be represented as single strided `read ... vector<4x2xf32>` and then decomposed into individual reads in the current mma lowering model, which I don't think we support currently.

harsh-nod (approved):
lgtm!

PR #706, opened 2025-04-11
[Wave] Improve WaveCacher to handle changing freevar in nested fn.

raikonenfnu:
- Actually use the subfunctions to do the hashing aside (this overall fixes changing freevar in nested fn)
   however brought up/exposed some other issues
- Hoist out function/subfunction related keys to outside the loop
- Use list to track subfunctions because set generates non-deterministic hashes
- error handling when fn.\_\_closure__ == None.


harsh-nod (approved):
lgtm!

PR #710, opened 2025-04-14
[Wave] Refactor layers to nn directory to mirror torch.

raikonenfnu:
Refactor wave layers directory to nn layer which mimics/mirrors pytorch style better.
```
torch.nn.linear -> wave.nn.WaveLinear
torch.nn.scaled_dot_product_attention ->wave.nn.functional.wave_sdpa_fp8
```

Future work would be to refactor out the WaveLinear kernel as a functional similar to torch.nn.Linear VS torch.nn.functional.linear.


nithinsubbiah (approved):
This is really nice putting our API on parity with Torch. Just one question from your comment on future work:
`Future work would be to refactor out the WaveLinear kernel as a functional similar to torch.nn.Linear VS torch.nn.functional.linear.`
By refactor WaveLinear, I hope you mean introducing `wave_nn.functional.linear` in addition to having `wave_nn.WaveQuantLinear`

PR #719, opened 2025-04-16
First pass of async execution in the runtime.

AWoloszyn:


AWoloszyn:
> be helpful to run `pre-commit run --all` to make sure formatting is valid.

Will do!

zjgarvey:
Hey, I'm looking into the test failures. I think it's happening because I fixed the session cache for conv launchables. I'll look for a fix right now.

iree/turbine/runtime/launch.py
```diff
@@ -4,6 +4,15 @@
...
 
+# HERE BE HACKS
+import ctypes
+import sys
+
+inc_ref = ctypes.pythonapi.Py_IncRef
+inc_ref.argtypes = [ctypes.py_object]
+inc_ref.restype = None
+
+
```
zjgarvey, line 15: What does this do?

zjgarvey (commented):
@stellaraccident would you mind taking a look?

Although the boo tests are using the async path, it might be a good idea to add a test solely for async launchables, but I'm not sure what would be appropriate beyond handwriting some async IR and verifying that `Launchable` works with `entry_point="main$async"` (maybe comparing numerics against the same module with `entry_point="main"`).

zjgarvey (approved):
LGTM

PR #723, opened 2025-04-17
[BOO] allow `BooConv2d` to take advantage of non-contiguous memory formats

zjgarvey:
In particular, this should allow setting `model = model.to(memory_format=torch.channels_last)` to enable `NHWC` convolutions, whereas previously we had been forcing these tensors back to `NHWC` format by calling `contiguous()` on the inputs inside the `Launchable` call. 

Tests are added to check if `BooConv2d` is calling into the expected kernel signatures.

### DIFF
--- a/iree/turbine/kernel/boo/modeling/replace.py
+++ b/iree/turbine/kernel/boo/modeling/replace.py
@@ -29,7 +29,9 @@ class BooConv2d(torch.nn.Module):
         super(BooConv2d, self).__init__()
         self.in_channels = in_channels
         self.out_channels = out_channels
-        self.kernel_size = kernel_size
+        self.kernel_size = (
+            [kernel_size] * 2 if isinstance(kernel_size, int) else kernel_size
+        )
         self.stride = stride
         self.padding = padding
         self.dilation = dilation
@@ -37,7 +39,7 @@ class BooConv2d(torch.nn.Module):
         self._bias = bias
 
         self.weight = torch.nn.Parameter(
-            torch.randn(out_channels, in_channels // groups, *kernel_size)
+            torch.randn(out_channels, in_channels // groups, *self.kernel_size)
         )
         if bias:
             self.bias = torch.nn.Parameter(torch.randn(out_channels))
--- a/iree/turbine/kernel/boo/modeling/replace.py
+++ b/iree/turbine/kernel/boo/modeling/replace.py
@@ -50,6 +50,9 @@ class BooConv2d(torch.nn.Module):
         input_layout = "NCHW"
         kernel_layout = "NCHW"
         output_layout = "NCHW"
+        no_batch = len(x.shape) == 3
+        if no_batch:
+            x = x.unsqueeze(0)
         if x.is_contiguous(memory_format=torch.channels_last):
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/boo/modeling/replace.py
```diff
@@ -29,34 +29,52 @@ def __init__(
...
             self.bias = torch.nn.Parameter(torch.randn(out_channels))
         else:
             self.register_parameter("bias", None)
 
     def forward(self, x):
+        input_layout = "NCHW"
+        kernel_layout = "NCHW"
+        output_layout = "NCHW"
+        if x.is_contiguous(memory_format=torch.channels_last):
+            x = x.permute([0, 2, 3, 1])
```
rkayaith, line 54: is there any case where the input may not be 4D? e.g. does torch's conv2d support a variable number of batch dims?

iree/turbine/kernel/boo/modeling/replace.py
```diff
@@ -29,34 +29,52 @@ def __init__(
...
+            input_layout = "NHWC"
+        w = self.weight
+        if w.is_contiguous(memory_format=torch.channels_last):
+            w = w.permute([0, 2, 3, 1])
+            kernel_layout = "NHWC"
+            output_layout = "NHWC"
```
rkayaith, line 60: is there any particular reason to match the output layout to the weights layout rather than the input layout?

rkayaith (approved):
nice, LGTM! just some minor questions

iree/turbine/kernel/boo/modeling/replace.py
zjgarvey: That's a good point. I'll look into adding support for this case.

zjgarvey: Their conv2d allows rank 3 or 4 input tensors and returns a tensor of the same rank.

iree/turbine/kernel/boo/modeling/replace.py
zjgarvey: Yeah, my reasoning for this is:

It's possible the initial model inputs aren't in `channels_last` format. If the weights are in `channels_last` format, I'm expecting that the rest of the model is going to be using `channels_last`, and we should try to produce outputs that are also `channels_last` to speed-up subsequent convolutions.

iree/turbine/kernel/boo/modeling/replace.py
```diff
@@ -29,34 +29,55 @@ def __init__(
         super(BooConv2d, self).__init__()
         self.in_channels = in_channels
         self.out_channels = out_channels
-        self.kernel_size = kernel_size
+        self.kernel_size = (
```
MaheshRavishankar, line 32: I dont think I follow this change?

MaheshRavishankar (commented):
Just have one question otherwise it looks good to me.

iree/turbine/kernel/boo/modeling/replace.py
zjgarvey: @MaheshRavishankar Does the above comment address your concern? 

I apologize for merging without hearing back from you on this comment.

PR #724, opened 2025-04-17
[Wave] Add atomic min op lowering

nithinsubbiah:
Adds atomic min op lowering
- Supports int data types. Float type needs support in MLIR to be able to lower


iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1132,6 +1136,58 @@ class SchedulingBarrier(CustomOp):
     operations: list[Operation]
 
 
+@define_op("atomic_min")
+@dataclass
+class AtomicMin(CustomOp):
```
harsh-nod, line 1141: Can this inherit from BinaryPyOp?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -375,6 +376,44 @@ def handle_shuffle(emitter: WaveEmitter, node: fx.Node):
...
+                    f"{op}\nGot\n"
+                    f"rhs: {rhs.ir_value.type}\n")
+            if lhs_data_type != rhs_data_type:
+                op = get_custom(node)
+                raise ValidationError(
+                    f"Expected lhs and rhs to have same data type for\n"
+                    f"{op}\nGot\n"
+                    f"lhs: {lhs_data_type} vs rhs: {rhs_data_type}\n"
+                )
+            if _is_float_type(lhs_data_type):
```
harsh-nod, line 403: Could you add a TODO above this with your learnings.

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -601,6 +640,24 @@ def handle_minimum(lhs: Value, rhs: Value, options: WaveCompileOptions) -> OpRes
...
+    value_type = val.type
+    value_element_type = get_type_or_element_type(value_type)
+    atomic_kind = None
+    if _is_float_type(value_element_type):
+        atomic_kind = arith_d.AtomicRMWKind.minimumf
+    elif _is_integer_like_type(value_element_type) and _is_signed_or_signless_type(
+        value_element_type):
+        atomic_kind = arith_d.AtomicRMWKind.mins
+
+    a = vector_d.extract(val, static_position=[0], dynamic_position=[])
```
harsh-nod, line 655: Could you add an assert that this is a vector of rank 1 and size 1?

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1623,3 +1624,59 @@ def test(
...
+    constraints: list[tkw.Constraint] = [
+        tkw.HardwareConstraint(
+            threads_per_wave=wave_size,
+            waves_per_block=(1, 1, 1),
+            vector_shapes={M: BLOCK_M, N: BLOCK_N},
+        )
+    ]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 0)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
```
harsh-nod, line 1652: have more than 1 wave?

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1623,3 +1624,59 @@ def test(
...
+    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+
+    @tkw.wave(constraints)
+    def test(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.i32],
+    ):
+        res = tkw.read(a)
+        neg_shmem = tkw.allocate(shape=(M,N), distributed_shape=(M,N), dtype=tkl.i32, address_space=tkl.AddressSpace.SHARED_MEMORY)
```
harsh-nod, line 1660: Can you also try with global memory (if it works)?

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1623,3 +1624,60 @@ def test(
...
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 0)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
+    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+
+    @tkw.wave(constraints)
+    def test(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.i32],
+    ):
+        res = tkw.read(a, elements_per_thread=4)
+        shmem = tkw.allocate(shape=(M,N), distributed_shape=(BLOCK_M, BLOCK_N), dtype=tkl.i32, address_space=tkl.AddressSpace.SHARED_MEMORY)
```
raikonenfnu, line 1660: can we make default to shared_memory, since we cannot allocate in other memory (that we expose of)

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1623,3 +1624,60 @@ def test(
...
+        zero_reg = tkl.Register[M, N, tkl.i32](55555)
```
raikonenfnu, line 1661: let's put this as 1e6 and call it inf_reg or something since it is not zeros

iree/turbine/kernel/ops/wave_ops.py
nithinsubbiah: done

iree/turbine/kernel/wave/codegen/handlers.py
nithinsubbiah: done

iree/turbine/kernel/wave/codegen/handlers.py
nithinsubbiah: I've added check for rank and size in the handler

tests/kernel/wave/wave_e2e_test.py
nithinsubbiah: I've added SHARED_MEMORY as default arg

tests/kernel/wave/wave_e2e_test.py
nithinsubbiah: done

tests/kernel/wave/wave_e2e_test.py
nithinsubbiah: I'll follow up in the next PR if that's alright, global memory probably needs some work on interface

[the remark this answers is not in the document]
nithinsubbiah: We can use hardware constraint (through emitter) to infer the shape however there's another use case in promotion where shape is queried as part of the allocate node's property and at which point we don't have access to emitter (https://github.com/iree-org/iree-turbine/blob/main/iree/turbine/kernel/wave/promotion.py#L163).

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1147,6 +1155,20 @@ class SchedulingBarrier(CustomOp):
     operations: list[Operation]
 
 
+@define_op("atomic_min")
+@dataclass
+class AtomicMin(BinaryOpBase, ABC):
+    """
+    Represents an atomic operation in the graph.
+    Takes in value and buffer to perform the operation on.
```
harsh-nod, line 1163: Can you also specify what it returns because I think there's some confusion there about whether it returns the original value or the updated value

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -377,6 +380,68 @@ def handle_shuffle(emitter: WaveEmitter, node: fx.Node):
...
+                    f"{op}\nGot\n"
+                    f"lhs: {lhs_data_type} vs rhs: {rhs_data_type}\n"
+                )
+            if _is_float_type(lhs_data_type):
+                # TODO: To support float types, MLIR LLVM dialect needs to be updated with
+                # float types. LLVM already supports fmin and fmax (https://llvm.org/docs/LangRef.html#atomicrmw-instruction)
+                # and thus atomicrmw operation in MLIR dialect needs to target those instructions with "workgroup" scope.
+                raise NotImplementedError(f"Atomic ops don't support float types yet\n")
+            if elements_per_thread > 1:
+                raise ValidationError(
```
harsh-nod, line 416: I thought multiple elements per thread works so why are you throwing an error here?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -377,6 +380,68 @@ def handle_shuffle(emitter: WaveEmitter, node: fx.Node):
...
+            keys = list(local_index.keys())
+            fastest_dim = get_fastest_index(node.index)
+            for i in range(elements_per_thread - 1):
+                new_index = copy.deepcopy(local_index)
+                key = keys[fastest_dim]
+                new_index[key].start += i + 1
+                start_idx = _build_start_indices(emitter, new_index)
+                start_indices.append(start_idx)
+
+            result = binary_fn(lhs, rhs, start_indices, emitter.options)
```
harsh-nod, line 439: Isn't build start indices used for gathers as well? How does that fit in here?  If you have to load four elements from shared memory and it's a gather Will this still work?

harsh-nod (changes requested):

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1153,6 +1166,33 @@ class SchedulingBarrier(CustomOp):
     operations: list[Operation]
 
 
+@define_op("atomic_min")
+@dataclass
+class AtomicMin(BinaryOpBase, ABC):
```
raikonenfnu, line 1171: The base class should just be AtomicOp not `AtomicMin`

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1153,6 +1166,33 @@ class SchedulingBarrier(CustomOp):
...
+    """
+    Represents an atomic operation in the graph.
+    Takes in value and buffer to perform the operation on.
```
raikonenfnu, line 1174: Can you add a bit more detail WRT how buffer is a memory type (e.g global/shared) and value is a register

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -622,6 +698,30 @@ def handle_minimum(lhs: Value, rhs: Value, options: WaveCompileOptions) -> OpRes
...
+    atomic_kind = None
+    if _is_float_type(value_element_type):
+        atomic_kind = arith_d.AtomicRMWKind.minimumf
+    elif _is_integer_like_type(value_element_type) and _is_signed_or_signless_type(
+        value_element_type
+    ):
+        atomic_kind = arith_d.AtomicRMWKind.mins
+    num_elements = val.type.shape[0]
+    unroll_result = []
+    for i in range(num_elements):
```
raikonenfnu, line 716: Should add a comment here describing how, memref_d.atomic_rmw only handles scalars

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -622,6 +698,30 @@ def handle_minimum(lhs: Value, rhs: Value, options: WaveCompileOptions) -> OpRes
...
+        a = vector_d.extract(val, static_position=[i], dynamic_position=[])
+        atomic_result = memref_d.atomic_rmw(atomic_kind, a, buffer, idx[i])
+        unroll_result.append(atomic_result)
+
+    result = vector_d.from_elements(value_type, unroll_result)
+    return result
```
raikonenfnu, line 722: Perhaps this should be handled/moved to `handle_atomic_op` S.T we can reuse this unrolling for future atomic ops.

lit_tests/kernel/wave/codegen.py
```diff
@@ -2338,3 +2339,58 @@ def scanop_cumsum(
...
+        canonicalize=True,
+        use_buffer_load_ops=False,
+        use_buffer_store_ops=False,
+        compile_to_mlir=True,
+    )
+
+    atomic_min_codegen = wave_compile(options, atomic_min_codegen)
+    print(atomic_min_codegen.asm)
+
+    # CHECK-LABEL: @atomic_min_codegen
```
raikonenfnu, line 2392: Can we add a LIT test here for atomic_min with mapping

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1834,3 +1835,81 @@ def test(
...
+    num_waves = 2
+
+    constraints: list[tkw.Constraint] = [
+        tkw.HardwareConstraint(
+            threads_per_wave=wave_size,
+            waves_per_block=(1, 2, 1),
+            vector_shapes={
+                M: int(BLOCK_M / num_waves),
+                N: BLOCK_N,
+                SHMEM_DIM: int(BLOCK_M / num_waves),
```
raikonenfnu, line 1864: if this is being reduced, this will always be 1. We shouldn't need to introduce a new shmem_dim. Introducing a new shmem_dim would be quite confusing because it doesn't really map to any tensor level dimensions

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1834,3 +1835,81 @@ def test(
...
+
+    @tkw.wave(constraints)
+    def test(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.i32],
+    ):
+        res = tkw.read(a, elements_per_thread=4)
+        shmem = tkw.allocate(
+            shape=(SHMEM_DIM, N),
+            distributed_shape=(int(BLOCK_M / num_waves), BLOCK_N),
+            dtype=tkl.i32,
```
raikonenfnu, line 1887: Can we make `shape=(M, N), distributed_shape=(1, BLOCK_N)`?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1153,6 +1166,33 @@ class SchedulingBarrier(CustomOp):
     operations: list[Operation]
 
 
+@define_op("atomic_min")
+@dataclass
+class AtomicOp(BinaryOpBase, ABC):
+    """
+    Represents an atomic operation in the graph.
+    Takes in value (register) and buffer (shared/global) as inputs,
+    and writes the modified value back on to the buffer.
```
raikonenfnu, line 1175: Can we add a comment on how the mapping is used for the shared memory only

iree/turbine/kernel/wave/promotion.py
```diff
@@ -122,6 +122,11 @@ def promote_node(
     symbolic_shape = node.type.symbolic_shape
     with node.graph.inserting_after(node.graph._root):
         constrained_shape = get_constrained_shape(symbolic_shape, constraints)
+        # If the read/write involves shared memory, then use the distributed shape
```
raikonenfnu, line 125: read/write here would most likely involves shared memory. Should detail that if already has existing shared memory with pre-specified distributed_shape (e.g defined from kernel level), then we use the distribtued shape if not we use constrained/block shape deduced from constraint

lit_tests/kernel/wave/codegen.py
```diff
@@ -2338,3 +2339,96 @@ def scanop_cumsum(
...
+            M: shape[0],
+            N: shape[1],
+            SHMEM_DIM: shape[0] / num_waves,
+            ADDRESS_SPACE: tkl.AddressSpace.GLOBAL_MEMORY.value,
+        },
+        canonicalize=True,
+        use_buffer_load_ops=False,
+        use_buffer_store_ops=False,
+        compile_to_mlir=True,
+    )
```
raikonenfnu, line 2409: Update LIT test please

iree/turbine/kernel/wave/shared_memory_indexing.py
```diff
@@ -5,29 +5,29 @@
...
 ):
     """
-    This function removes global indexing from shared memory reads and writes.
+    This function removes global indexing from ops that deal with shared memory.
     Global indexing is an indexing that arises from Workgroup constraints
     and Tiling constraints.
     """
 
-    def is_shared_memory_read_or_write(node: fx.Node):
+    def shared_memory_ops(node: fx.Node):
```
raikonenfnu, line 24: can we use `is_shared_memory_ops`

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1834,3 +1835,82 @@ def test(
...
+        },
+        canonicalize=True,
+        run_bench=run_bench,
+        use_buffer_load_ops=use_buffer_ops,
+        use_buffer_store_ops=use_buffer_ops,
+    )
+    options = set_default_run_config(options)
+    test = wave_compile(options, test)
+    test(a, c)
+    assert_close(c[0, :], b)
```
raikonenfnu, line 1916: can you also assert on c[1,:] ?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -622,6 +706,23 @@ def handle_minimum(lhs: Value, rhs: Value, options: WaveCompileOptions) -> OpRes
...
+) -> OpResult:
+    value_element_type = get_type_or_element_type(val.type)
+    atomic_kind = None
+    # Only scalars are supported currently
+    if _is_float_type(value_element_type):
+        atomic_kind = arith_d.AtomicRMWKind.minimumf
+    elif _is_integer_like_type(value_element_type) and _is_signed_or_signless_type(
+        value_element_type
+    ):
+        atomic_kind = arith_d.AtomicRMWKind.mins
```
raikonenfnu, line 721: should add another else to fail/error out if it's neither type

iree/turbine/kernel/lang/wave_types.py
```diff
@@ -213,6 +213,16 @@ def __init__(
             ), f"Iterator conflict: {current} and {sym}"
             iter_shape[i] = sym
 
+        # If number of iterators is less than the symbol assigned iterators
+        # and the expression in the mapping is an integer, the integer denotes
+        # an index
```
raikonenfnu, line 218: Please give example use case of this

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -1834,3 +1835,88 @@ def test(
...
+    ):
+        res = tkw.read(a, elements_per_thread=4)
+        # We allocate a buffer of (1,BLOCK_N) shape to perform reduction across
+        # waves. Inputs are distributed with (1,BLOCK_N) shape across each wave
+        # and performs atomic min operation on this shared memory space. Mapping
+        # attribute to atomic_min op is utilized to access the same shared memory
+        # from different waves.
+        shmem = tkw.allocate(
+            shape=(M, N),
+            distributed_shape=(1, BLOCK_N),
```
Hardcode84, line 1894: I wonder if can infer `distributed_shape` automatically. I think logical `shape` should be `(M / 2, N)`, then, if we replace M and N with corresponding BLOCK_M, BLOCK_N we will get `distributed_shape` we want.

iree/turbine/kernel/lang/wave_types.py
```diff
@@ -213,6 +213,18 @@ def __init__(
             ), f"Iterator conflict: {current} and {sym}"
             iter_shape[i] = sym
 
+        # If number of iterators is less than the symbol assigned iterators and the
```
harsh-nod, line 216: Can you explain why this is being added?

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1153,6 +1170,35 @@ class SchedulingBarrier(CustomOp):
     operations: list[Operation]
 
 
+@define_op("atomic_min")
+@dataclass
+class AtomicOp(BinaryOpBase, ABC):
+    """
+    Represents an atomic operation in the graph.
+    Takes in value (register) and buffer (shared/global) as inputs,
```
harsh-nod, line 1178: minor: can you update this to say just Register and Memory as those are the input types.

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -379,6 +381,88 @@ def handle_shuffle(emitter: WaveEmitter, node: fx.Node):
...
+                raise NotImplementedError(f"Atomic ops don't support float types yet\n")
+
+            lhs = lhs.ir_value
+            rhs = rhs.ir_value
+            lhs_type = lhs.type
+            assert (
+                lhs_type.rank == 0 or lhs_type.rank == 1
+            ), f"expected lhs_type.rank == 1 but got {lhs_type.rank}, {node}"
+
+            node_index = node.index
```
harsh-nod, line 424: Could you use a different name? The edit distance between the two variables names is too low and bound to cause confusion.

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -379,6 +381,88 @@ def handle_shuffle(emitter: WaveEmitter, node: fx.Node):
...
+            if mapping:
+                symbolic_shape = get_custom(node).type.symbolic_shape
```
harsh-nod, line 426: Can you pull this out into a utils function? Seems very useful as it seems like it transforms an index based on the provided mapping? I can imagine this function being used throughout the code to unify mapping into the index so that the index is the sole attribute we need to reason about.

iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -40,17 +38,19 @@
 
 from ...ops.wave_ops import get_custom, read, write, CustomOp
 
-from ..utils.general_utils import find_index_bounds, get_fastest_index
+from ..utils.general_utils import (
+    find_index_bounds,
+    get_fastest_index,
+)
```
harsh-nod, line 44: are any of these imports being used? if not please remove.

iree/turbine/kernel/wave/promotion.py
```diff
@@ -122,6 +122,13 @@ def promote_node(
     symbolic_shape = node.type.symbolic_shape
     with node.graph.inserting_after(node.graph._root):
         constrained_shape = get_constrained_shape(symbolic_shape, constraints)
+        # If the read/write operation already has a set distributed shape at the kernel
+        # we use that for allocation. Otherwise deduce the shape from constraints.
+        memory_node = get_custom(node.memory)
+        if isinstance(memory_node, Allocate) and hasattr(
```
harsh-nod, line 128: doesnt allocate always have the distributed_shape attribute? Shouldnt you instead check if it is None?

harsh-nod (changes requested):
overall, looks good. I think this is getting ready to land with just some minor comments.

iree/turbine/kernel/lang/wave_types.py
```diff
@@ -19,7 +19,7 @@
 from .._support.dtype import DataType
 from .._support.indexing import IndexExpr, IndexSymbol, index_symbol
 
-from sympy import Symbol
+from sympy import Integer, Symbol
```
Hardcode84, line 22: Not used?

iree/turbine/kernel/wave/codegen/handlers.py
Hardcode84: This wasn't addressed

lit_tests/kernel/wave/codegen.py
```diff
@@ -2340,3 +2341,97 @@ def scanop_cumsum(
...
+            waves_per_block=(1, num_waves, 1),
+            vector_shapes={
+                M: int(BLOCK_M / num_waves),
+                N: BLOCK_N,
+            },
+        )
+    ]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 0)]
+    constraints += [tkw.WaveConstraint(M, sympy.Integer(1))]
```
Hardcode84, line 2370: Here and later: we can use just `1`

lit_tests/kernel/wave/codegen.py
```diff
@@ -2340,3 +2341,97 @@ def scanop_cumsum(
...
+        res = tkw.read(a, elements_per_thread=4)
+        shmem = tkw.allocate(
+            shape=(M, N),
+            distributed_shape=(1, BLOCK_N),
+            dtype=tkl.i32,
+        )
+        inf_reg = tkl.Register[M, N, tkl.i32](1e6)
+        tkw.write(inf_reg, shmem, elements_per_thread=4)
+        res = tkw.atomic_min(res, shmem, elements_per_thread=4, mapping=mapping)
+        res = tkw.read(shmem, elements_per_thread=4, mapping=read_mapping)
```
Hardcode84, line 2398: Do we need `elements_per_thread`?

Hardcode84 (commented):
Mostly LGTM, just a few small comments

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1104,6 +1117,10 @@ def indexing_dims(self) -> list[IndexSymbol]:
     def type(self) -> "Memory":
         return Memory[(*self.shape, self.address_space, self.dtype)]
 
+    @property
```
harsh-nod, line 1120: Is this needed?

harsh-nod: Also not right type

harsh-nod (approved):
lgtm! please land after addressing comments

Hardcode84 (approved):

PR #725, opened 2025-04-18
[WAVE] support cumsum scan op

xintin:
Implementation of scan op::cumsum
This implementation supports only one element per thread. 
There is another PR, where we support >= 1 elements per thread.


iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -0,0 +1,131 @@
...
+        shuffle = ShuffleOp(init, offset_val, subgroup_size)
+        shuffle_val = get_graph_node(shuffle, graph)
+
+        lane_id = hardware_constraint.linearized_thread_id % subgroup_size
+
+        # we are explicitly adding index because this pass is being
+        # applied after the indexing phase
+        # ToDo (xintin): check if we can replace register with scalar.
+        # No point using register for a scalar. Applies to other objects too.
+        zero_vec = get_register_as_graph_node(shuffle_val, 0.0, graph)
```
xintin, line 65: this is IDENTITY in wave wise scan

iree/turbine/kernel/_support/tracing.py
```diff
@@ -340,6 +338,17 @@ def handle_vector_constant(
             kwargs={},
         )
 
+    ### ========================================================================
+    ### Scan Operations
```
raikonenfnu, line 342: I don't think you should need any of these changes in this file right?

iree/turbine/kernel/lang/prims.py
```diff
@@ -54,3 +55,6 @@ def is_debug() -> bool:
 broadcast = ops.vector_broadcast
 broadcast_in_dim = ops.vector_broadcast_in_dim
 transpose = ops.vector_transpose
+
+# Scan Operations
+cumsum = ops.vector_cumsum
```
raikonenfnu, line 60: same here, I don't think you need any changes from here

iree/turbine/kernel/ops/__init__.py
```diff
@@ -4,3 +4,4 @@
 from .control_flow import *
 from .memory import *
 from .shape_manipulation import *
+from .scan import *
```
raikonenfnu, line 7: ditto with not needing changes from this file

iree/turbine/kernel/ops/scan.py
```diff
@@ -0,0 +1,17 @@
+import typing
+
```
raikonenfnu, line 2: ditto with not needing changes from this file

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -0,0 +1,130 @@
...
+from ..wave.constraints import HardwareConstraint
+from .utils.graph_utils import DCE
+
+
+def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
+    custom.add_to_graph(graph)
+    return custom.fx_node
+
+
+def get_register_as_graph_node(
```
raikonenfnu, line 30: NIT: I think for now may be more compact/straight forward if we use `get_graph_node(NewRegister(shape, dtype, value), graph)` as opposed to having this helper function

raikonenfnu (commented):
Overall looks good! just some NITs and what it seems like dead code

raikonenfnu (approved):
lgtm! :)

PR #733, opened 2025-04-21
[Wave] Add support for iterating over an arbitrary range parameterized

harsh-nod:
by an init value, next value and terminating condition.


iree/turbine/kernel/wave/constraints.py
```diff
@@ -445,13 +445,28 @@ class TilingConstraint(DistributionConstraint):
...
 
     dim: IndexExpr
-    tile_size: IndexExpr
+    tile_size: Optional[IndexExpr] = None
     induction_var: Optional[IndexExpr] = None
     iters: Optional[IndexExpr] = None
     start: IndexExpr = Integer(0)
+    init_symbol: Optional[IndexSymbol] = None
+    next_symbol: Optional[IndexSymbol] = None
+    condition: Optional[Callable] = None
```
raikonenfnu, line 462: We can add this for next PR, but leaving a comment as a TODO/documenation. Would be good to have the init_symbol, next_symbol, and condition to be directly inserted inside the kernel within the `tkw.iterate` function call. This will allow us to iterate over the same dimension but different conditions

raikonenfnu (approved):
Overall looks good aside from the design which we decide to iterate on

iree/turbine/kernel/wave/constraints.py
harsh-nod: I have a new version now that should address all of these.

xintin (approved):

PR #740, opened 2025-04-22
[TKW] Use uptream `gpu.shuffle` repacking

Hardcode84:
AMDGPU shuffles only support i32 internally and all other datatypes need to be bitcasted/zexted/repacked to multiple values. Instead of doing this on Wave side (which was incomplete anyway), rely on upstream repacking during GPU->ROCDL lowering, which was recently added.

This requires latest IREE so it's a breaking change.

### DIFF
--- a/iree/turbine/kernel/wave/codegen/handlers.py
+++ b/iree/turbine/kernel/wave/codegen/handlers.py
@@ -343,37 +343,10 @@ def handle_shuffle(emitter: WaveEmitter, node: fx.Node):
     offset = cast_py_value(emitter, offset, IntegerType.get_signless(32)).ir_value
     width = cast_py_value(emitter, width, IntegerType.get_signless(32)).ir_value
 
-    if not VectorType.isinstance(src.type):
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -343,37 +343,10 @@ def handle_shuffle(emitter: WaveEmitter, node: fx.Node):
...
-    elif element.type.width < 32:
-        element = arith_d.extf(F32Type.get(), element)
-
     # Shuffle data between other threads in a warp.
-    result = gpu_d.shuffle(element, offset, width, gpu_d.ShuffleMode.XOR)
-
-    # Reconstruct shuffled value to original shape and dtype.
-    shuffled_val = result[0]
-    if element_original_type != shuffled_val.type:
-        shuffled_val = arith_d.truncf(element_original_type, shuffled_val)
```
raikonenfnu, line 373: very nice :)

raikonenfnu (approved):
LGTM

PR #747, opened 2025-04-23
[boo] Fix boo_driver sometimes printing result values

rkayaith:
We don't care about the actual values, so set the print threshold to `0` so they're always hidden.


jerryyin (approved):
LGTM, thanks for tracking it down!

PR #751, opened 2025-04-24
[Wave] fix read/write index to use laneIdx

raikonenfnu:
Currently most non-MMA test cases uses 1 warp, but once we go over 1 warp this will break, as we are using block threadIdx.x instead of using laneIdx. This will be problematic because linearized access pattern looks something like:
```
blockIdx.x * blockDim.x + waveIdx.x * waveDim.x + threadIdx.x
```

In case above threadIdx.x will go over the entire dimension of thread in the blockDimension, but this should already be accounted for from the waveIdx.x * waveDim.x, hence will cause out of bound/unexpected behaviour. This PR essentially changes the threadIdx.x into laneIdx.x which is threadIdx.x // threads_per_wave.


harsh-nod (approved):
thanks!

PR #752, opened 2025-04-24
[Wave] Implement NewScalar op support

raikonenfnu:
- Refactor conversion op as a util/helper function S.T we can reuse for ScalarOp handler.
- Add support for inferType on ComparisonPyOp on Scalars
- refactor dtype of DataType to be self which is what is expected and can support all dtypes.
- Implement NewScalarOp and handlers

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -2055,11 +2055,11 @@ def test_scalar_cond_copy():
         a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
         b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
     ):
-        zero = tkw.scalar(tkl.f16, 0.0)
-        one = tkw.scalar(tkl.f16, 1.0)
+        zero = tkw.scalar(0.0, tkl.f16)
+        one = tkw.scalar(1.0, tkl.f16)
 
-        tid = tkw.scalar(tkl.i32, THREAD_0)
-        thresh = tkw.scalar(tkl.i32, thresh_value)
+        tid = tkw.scalar(THREAD_0, tkl.i32)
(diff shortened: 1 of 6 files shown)


lit_tests/kernel/wave/codegen.py
```diff
@@ -2018,3 +2018,79 @@ def scalar_codegen_i32(
...
+    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+
+    thresh_value = 12
+
+    @tkw.wave(constraints)
+    def scalar_cond_copy(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+        b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+    ):
+        zero = tkw.scalar(tkl.f16, 0.0)
```
harsh-nod, line 2058: minor: can you move the type to the second argument?
so something like
```
 tid = tkw.scalar(THREAD_0, tkl.i32)
```
This is similar to other ops like self_index and cast which have the type second.

harsh-nod (approved):
nice! just a minor stylistic comment but looks good otherwise!

lit_tests/kernel/wave/codegen.py
raikonenfnu: done, thanks :)

lit_tests/kernel/wave/codegen.py
```diff
@@ -2018,3 +2018,79 @@ def scalar_codegen_i32(
...
+
+    options = WaveCompileOptions(
+        subs={
+            M: 1,
+            N: 64,
+            ADDRESS_SPACE: tkl.AddressSpace.GLOBAL_MEMORY.value,
+        },
+        canonicalize=True,
+    )
+    scalar_cond_copy = wave_compile(options, scalar_cond_copy)
```
harsh-nod, line 2079: This is a lit test and shouldn't have a wave compile here.

PR #753, opened 2025-04-24
[WAVE] Support wave-wise cumsum scan op::more than one elem per thread

xintin:
ScanOp cumsum to support more than 1 element per thread. 
Logic for single element and multiple is separate for now because of indexing.


iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -0,0 +1,232 @@
...
+    lane_id = (
+        hardware_constraint.linearized_thread_id % hardware_constraint.threads_per_wave
+    )
+
+    target_shape = list(src.type.symbolic_shape)
+    target_shape.pop(target_shape.index(scan_dim))
+
+    thread_incl = offset
+
+    num_steps = int(math.log2(float(subgroup_size)))
```
xintin, line 95: currently, we are looping in a pow of 2 in range of subgroup size which is 64. 
To support, unaligned shapes, this needs to revisited.

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -123,7 +124,8 @@ def handle_register(emitter: WaveEmitter, node: fx.Node):
     except ValueError as e:
         raise ValidationError("Malformed arguments") from e
     get_thread_shape = lambda index: max(subs_idxc(x.size) for x in index.values())
-    shape = [get_thread_shape(get_custom(node).index)]
+    if get_custom(node).index:
```
harsh-nod, line 127: Could you assign an index to the new register?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -34,6 +34,7 @@
     scf_d,
     vector_d,
     llvm_d,
+    ShapedType,
```
harsh-nod, line 37: Unused import?

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,134 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
+    for i in range(1, elements_per_thread):
+        values[i] = get_graph_node(binary_fn(values[i], values[i - 1]), graph)
+
+    return values
+
+
 def emit_global_scan(
     binary_fn: Callable,  # Supports only Add for now.
     src: fx.Node,
+    local_scan: list[fx.Node],
```
harsh-nod, line 58: Could you pass in just the fx.Node instead of accessing the last element of the list when the rest of the elements in the list are no longer required?

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -116,14 +200,56 @@ def decompose_scan_ops(
...
+            else:
+                result = emit_global_scan(
+                    binary_fn,
+                    scan_src,
+                    [scan_src],
+                    custom.graph,
+                    subgroup_size,
+                    hardware_constraint,
+                    local_scan_sizes,
+                    scan_dim,
```
raikonenfnu, line 250: IIUC, only `local_sca` here changes, can we take emit_global_scan out of the if condition, and just modify s.t `local_scan` be modified inside the if `local_scan_sizes > 1`

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,141 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
+        zero_vec2.index = get_custom(local_scan[-1]).index
+
+        excl_offset = get_graph_node(
+            SelectOp(cond=cond_expr2_node, if_true=shuffle2_node, if_false=zero_vec2),
+            graph,
+        )
+
+        final_scalars = []
+        for lane_elem in local_scan:
+            summed = get_graph_node(binary_fn(lane_elem, excl_offset), graph)
```
raikonenfnu, line 155: can we add some comment/doc here on what we are trying to do

raikonenfnu (commented):
Overall looks good, just some styling/structure nits :)

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,141 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
-            SelectOp(cond=cond_node, if_true=shuffle_val, if_false=zero_vec), graph
+            SelectOp(cond=cond_node, if_true=shuffle_val_node, if_false=zero_vec), graph
         )
 
-        init = get_graph_node(binary_fn(init, masked), graph)
+        # perform binary scan op
+        thread_incl = get_graph_node(binary_fn(thread_incl, masked), graph)
 
-    return init
+    scanop_result = thread_incl
```
raikonenfnu, line 124: can we use same variable for this?

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,141 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
+        scanop_result = Reshape(
+            args=final_scalars,
+            target_vector_shape={scan_dim: local_scan_size},
+        ).add_to_graph(graph)
+
+        scanop_result.index = get_custom(src).index
+        scanop_result.expanded_dims = get_custom(src).expanded_dims
+        scanop_result.vector_shapes = get_custom(src).vector_shapes
+
+    return scanop_result
```
raikonenfnu, line 169: nice :)

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,146 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
+    ToDo: Can we make it more efficient?
     """
-    init = src
+    lane_id = (
+        hardware_constraint.linearized_thread_id % hardware_constraint.threads_per_wave
+    )
+
+    scanop_result = last_local_scan_node = local_scan[-1]
+
+    if local_scan_size > 1:
```
raikonenfnu, line 81: NIT: Can you add small description why we need this for local_scan_size > 1

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,146 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
+        That is why we have another shuffle phase.
+
+    ToDo: Can we make it more efficient?
     """
-    init = src
+    lane_id = (
+        hardware_constraint.linearized_thread_id % hardware_constraint.threads_per_wave
+    )
+
+    scanop_result = last_local_scan_node = local_scan[-1]
```
raikonenfnu, line 79: can scan_op_result and last_local_scan_node be the same variable?

xintin: I looked into that but not possible as it gets modified in the code later.

raikonenfnu: but you only need it for the indexing?

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,148 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
```
raikonenfnu, line 79: Seems like we are only using `last_local_scan_node` for it's index. If that is the case, any thoughts on refactoring to something like:
```
scanop_result = last_local_scan_node = local_scan[-1]
local_scan_index = get_custom(local_scan[-1]).index
```
I think it makes each variable clearer on what it is for and added bonus of not needing to do get_custom() + .index at every use

iree/turbine/kernel/wave/decompose_scan_ops.py
```diff
@@ -28,61 +32,148 @@ def get_graph_node(custom: CustomOp, graph: fx.Graph) -> fx.Node:
...
 ) -> fx.Node:
     """
     Emit an intra-warp inclusive scan using butterfly pattern scan and masking.
+    In this algorithm, we are using two-pass approach.
+    If the number of elements per thread is one, then we just need one shuffle pass.
+    If there are more than one elements per thread, we need to calculate the exclusive
+        offset for the next lane elements using the information from the previous lane.
+        That is why we have another shuffle phase.
+
+    ToDo: Can we make it more efficient?
```
raikonenfnu, line 73: Sorry, per https://google.github.io/styleguide/pyguide.html#312-todo-comments, TODO should be all caps.

raikonenfnu (approved):
LGTM, let's wait for runner before merging though

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -41,6 +41,15 @@
...
 )
 
+import torch
+from torch.testing import assert_close
+import pytest
+import sympy
+import os
+import torch
+import json
+from .common.shapes import get_test_shapes as get_common_test_shape
```
raikonenfnu, line 51: I don't think we are using this

raikonenfnu (changes requested):

raikonenfnu (approved):

PR #755, opened 2025-04-24
[Wave] Add support for init args in while loops

harsh-nod:
This PR adds support for init args in the while loop style iterate ops. In order to accomplish this, a couple of changes had to be made
1. Moving add_get_results out from an expansion specific pass to one before the infer types pass
2. Fixing indices of MMA and ACC nodes during index propagation
3. Updates to handler. Specifically since the while loop returns the value of the induction variable while we don't expose this in the DSL, we reorder the arguments.
4. Changes to lit tests to fix this
5. Updates to logic to determine MMA reduction dims and added validation to avoid ambiguities

### DIFF
--- a/iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
+++ b/iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
@@ -9,6 +9,7 @@ from ...ops.wave_ops import (
     BinaryPyOp,
     Broadcast,
     CustomOp,
+    GetResult,
     IterArg,
     MMA,
     NestedRegionOp,
@@ -290,28 +291,30 @@ def populate_mma_source_indices(
             dim, dim_index, node.mma_type
         )
     node.index = combine_indices(node.index, index)
-    return [
-        (
-            get_custom(node.lhs),
-            specialize_index(index, {MMA_LHS: 1, MMA_RHS: 0, MMA_ACC: 0}),
-            node.vector_shapes,
-        ),
-        (
-            get_custom(node.rhs),
-            specialize_index(index, {MMA_LHS: 0, MMA_RHS: 1, MMA_ACC: 0}),
-            node.vector_shapes,
-        ),
-        (
-            get_custom(node.acc),
-            specialize_index(index, {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1}),
-            node.vector_shapes,
-        ),
-        (
-            node,
-            specialize_index(index, {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1}),
-            node.vector_shapes,
-        ),
-    ]
+    lhs_tuple = (
+        get_custom(node.lhs),
+        specialize_index(index, {MMA_LHS: 1, MMA_RHS: 0, MMA_ACC: 0}),
+        node.vector_shapes,
+    )
+    rhs_tuple = (
+        get_custom(node.rhs),
+        specialize_index(index, {MMA_LHS: 0, MMA_RHS: 1, MMA_ACC: 0}),
+        node.vector_shapes,
+    )
+    acc_tuple = (
+        get_custom(node.acc),
+        specialize_index(index, {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1}),
+        node.vector_shapes,
+    )
+    mma_tuple = (
+        node,
+        specialize_index(index, {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1}),
+        node.vector_shapes,
+    )
+    # Remove reduction dim from index of accumulator and mma nodes.
+    del acc_tuple[1][node.reduction_dim]
+    del mma_tuple[1][node.reduction_dim]
+    return [lhs_tuple, rhs_tuple, acc_tuple, mma_tuple]
 
 
 def populate_read_write_source_indices(
@@ -476,15 +479,18 @@ def propagate_indices(
                 source, source_index, source_vector_shapes, symbolic_constraints
             ):
                 continue
-            source_index = source.transform_index(source_index)
-            source.index = combine_indices(source.index, source_index)
+            # GetResults inherit their index from the Iterate node
+            # and hence we don't need to update their index.
+            if not isinstance(source, GetResult):
+                source_index = source.transform_index(source_index)
+                source.index = combine_indices(source.index, source_index)
             source.vector_shapes = source_vector_shapes
             append_aliased_shapes(source, symbolic_constraints)
         visited.add(source)
         for func in [get_inputs, get_users]:
-            sources, reduction = add_nodes_to_sources(
+            sources, _ = add_nodes_to_sources(
                 source,
-                reduction,
+                None,
                 func,
                 source_index,
                 source_vector_shapes,
--- a/iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
+++ b/iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
@@ -381,18 +381,17 @@ def combine_indices(
 
 def add_nodes_to_sources(
     source: CustomOp,
-    reduction: Iterate,
     fn: Callable,
     source_index: dict[IndexSymbol, IndexSequence],
     source_vector_shapes: dict[IndexSymbol, int],
     sources: list[
         tuple[CustomOp, dict[IndexSymbol, IndexSequence], dict[IndexSymbol, int]]
     ],
-) -> tuple[list[CustomOp], Iterate]:
+) -> list[CustomOp]:
     """
     Populate the sources with the inputs and users of the source node.
     """
-    for args, reduction in [fn(source.fx_node, reduction)]:
+    for args, reduction in [fn(source.fx_node, None)]:
         logger.debug(f"{source.fx_node} -> {args}")
         if not args:
             break
@@ -406,7 +405,7 @@ def add_nodes_to_sources(
                 custom.vector_shapes if custom.vector_shapes else source_vector_shapes
             )
             sources.append((custom, source_index, vector_shapes))
-    return sources, reduction
+    return sources
 
 
 def should_update_index(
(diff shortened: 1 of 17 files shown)


iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -290,28 +291,30 @@ def populate_mma_source_indices(
...
+        get_custom(node.acc),
+        specialize_index(index, {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1}),
+        node.vector_shapes,
+    )
+    mma_tuple = (
+        node,
+        specialize_index(index, {MMA_LHS: 0, MMA_RHS: 0, MMA_ACC: 1}),
+        node.vector_shapes,
+    )
+    # Remove reduction dim from index of accumulator and mma nodes.
```
raikonenfnu, line 314: nice! will this fix Nicolas's issue as well?

iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -476,15 +479,18 @@ def propagate_indices(
...
+            # GetResults inherit their index from the Iterate node
+            # and hence we don't need to update their index.
+            if not isinstance(source, GetResult):
+                source_index = source.transform_index(source_index)
+                source.index = combine_indices(source.index, source_index)
             source.vector_shapes = source_vector_shapes
             append_aliased_shapes(source, symbolic_constraints)
         visited.add(source)
         for func in [get_inputs, get_users]:
-            sources, reduction = add_nodes_to_sources(
```
raikonenfnu, line 485: seems like we don't need reductio in add_nodes_to_sources at all?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -935,49 +935,75 @@ def handle_iterate(emitter: WaveEmitter, node: fx.Node):
...
 
 def handle_iterate_while(emitter: WaveEmitter, node: fx.Node):
-    # TODO: Add support for init args.
     try:
         axis, init_args, subgraph, implicit_capture, start, condition = node.args
     except ValueError as e:
         raise ValidationError("Malformed arguments") from e
 
+    # Flatten init_args and get IR values for each of them.
+    flat_init_args, _ = pytree.tree_flatten((init_args))
```
raikonenfnu, line 944: Just curious, are init_args not already flat by default?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -935,49 +935,75 @@ def handle_iterate(emitter: WaveEmitter, node: fx.Node):
...
     # Before block: condition.
-    current_value = whileOp.before.blocks[0].arguments[0]
+    current_values = whileOp.before.blocks[0].arguments
     with InsertionPoint(whileOp.before.blocks[0]):
         # Replace the axis with a temporary variable when generating the condition
         # to avoid conflicts with the actual value of the axis.
         condition = condition.subs({axis: index_symbol("$TMP")})
         subs = add_emitter_subs(emitter)
-        subs[index_symbol("$TMP")] = current_value
+        subs[index_symbol("$TMP")] = current_values[-1]
```
raikonenfnu, line 972: any reasons why we index on -1? (as opposed to something like $TMP{i} = current_values[i])

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -935,49 +935,75 @@ def handle_iterate(emitter: WaveEmitter, node: fx.Node):
...
         condition = gen_sympy_index(subs, condition)
-        scf_d.ConditionOp(condition, [current_value])
+        scf_d.ConditionOp(condition, current_values)
 
     # After block: body.
-    current_value = whileOp.after.blocks[0].arguments[0]
+    current_values = whileOp.after.blocks[0].arguments
```
raikonenfnu, line 977: can we reuse the previous definition of `current_values`?

iree/turbine/kernel/wave/codegen/handlers.py
raikonenfnu: is current_values[-1] the condition and the rest are iter_args?

iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -935,49 +935,75 @@ def handle_iterate(emitter: WaveEmitter, node: fx.Node):
...
+
+        # Map the iteration arguments
+        iter_args = get_custom(node).iter_args(subgraph)
+        for i, arg in enumerate(iter_args):
+            if i < len(current_values) - 1:
+                emitter.bind_node_proxy(arg, IRProxyValue(current_values[i]))
+
+        # Emit the subgraph
+        return_values = emitter._emit_graph(subgraph)
+        if all(x is None for x in return_values):
```
raikonenfnu, line 997: Is this the case where there is no init_args aside from the condition variable? Can we add some comments/docs here?

iree/turbine/kernel/wave/expansion/expansion.py
```diff
@@ -401,7 +401,6 @@ def add_get_results(trace: CapturedTrace):
             get_result = get_custom(
                 GetResult(reduction.fx_node, 0).add_to_graph(reduction.graph)
             )
-            get_result.vector_shapes = reduction.init_args[0].vector_shapes
```
raikonenfnu, line 404: nice! this is because we are now doing getResult before index analysis and expansion, S.T we can let index analysis set the vector shapes?

iree/turbine/kernel/wave/utils/mma_utils.py
```diff
@@ -67,7 +67,18 @@ def is_mma(node):
         acc_shape = custom.acc_type.symbolic_shape
 
         try:
-            k = ((set(lhs_shape) & set(rhs_shape)) - set(acc_shape)).pop()
+            common_dims = (set(lhs_shape) & set(rhs_shape)) - set(acc_shape)
```
raikonenfnu, line 70: wouldn't `- set(acc_shape)` get rid of the batch dims?
```
(B, M, K) & (B, K, N) - (B, M, N)
= (B,K) - (B, M ,N)
= (K)
```

raikonenfnu (commented):
Overall lgtm just some minor questions

iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
harsh-nod: Possibly, but this was fixing a bug which is that since we are propagating mma/acc to init/users of mma, the index they use should not have the reduction dim in it.

iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
harsh-nod: yes I can remove it.

iree/turbine/kernel/wave/codegen/handlers.py
harsh-nod: Yes, unlike scf.for, scf.while also returns the value of the induction variable. Since we don't expose this, we need to reconcile the difference between what we expose through tkw.iterate and what scf.while does. I am resolving that by keeping the induction variable at the end, so that 0 to num_init_args - 1 is consistent between tkw.iterate and scf.while.

iree/turbine/kernel/wave/codegen/handlers.py
harsh-nod: yeah because we don't need that in the new block.

iree/turbine/kernel/wave/codegen/handlers.py
harsh-nod: yeah this is just a pure while with no init args. Sure will add a comment.

iree/turbine/kernel/wave/expansion/expansion.py
harsh-nod: yup that's right.

iree/turbine/kernel/wave/utils/mma_utils.py
harsh-nod: I think the name is the problem here, it should be 
```
reduction_dim_candidates = (set(lhs_shape) & set(rhs_shape)) - set(
                acc_shape
            )
```
In this case it does the right thing in your example above. But also for the case where
```
(B, M, K) & (B, K, N) - (M, N) -> (B, K)
```

iree/turbine/kernel/wave/analysis/index_sequence_analysis.py
raikonenfnu: nice! :)

iree/turbine/kernel/wave/utils/mma_utils.py
raikonenfnu: In what cases do we expect the accumulator to be (M, N) as opposed to (B, M, N)?

iree/turbine/kernel/wave/codegen/handlers.py
raikonenfnu: makes sense, thanks!

iree/turbine/kernel/wave/codegen/handlers.py
raikonenfnu: sorry, is this yeah we can reuse or yeah we need different definitions? haha

iree/turbine/kernel/wave/codegen/handlers.py
raikonenfnu: can we document/add a comment for this somewhere actually?

raikonenfnu (approved):
Overall looks good(er) now, so approving to unblock landing, but have some minor follow on questions :)

iree/turbine/kernel/wave/utils/mma_utils.py
harsh-nod: The test has this as an example. In general, not sure how frequent this might be as a use case.

iree/turbine/kernel/wave/codegen/handlers.py
harsh-nod: got it, its a yeah we need different definitions

iree/turbine/kernel/wave/codegen/handlers.py
harsh-nod: sure will add in subsequent pr

PR #756, opened 2025-04-25
[Wave] Implement operator.ne lowering and mask p_reg in spec dec

raikonenfnu:


iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -583,6 +583,20 @@ def handle_eq(lhs: Value, rhs: Value, options: WaveCompileOptions) -> OpResult:
     return result
 
 
+@handle_binary_op([operator.ne])
+def handle_ne(lhs: Value, rhs: Value, options: WaveCompileOptions) -> OpResult:
+    element_type = get_type_or_element_type(lhs.type)
+    if _is_float_type(element_type):
+        result = arith_d.cmpf(arith_d.CmpFPredicate.ONE, lhs, rhs)
```
Hardcode84, line 590: It took me some time to realize `CmpFPredicate.ONE` is `OrderedNE` )

Hardcode84 (approved):
Can we just land lowering and lit test? I think they are generally useful

PR #763, opened 2025-04-26
[Wave] Implement different shuffleMode + Handle unaligned Scan

raikonenfnu:
- Implement different shuffleMode for ShuffleOp with enums + mapping
- Use gpu shuffle up on scanOp to enable unaligned shapes

### DIFF
--- a/iree/turbine/kernel/wave/utils/classes.py
+++ b/iree/turbine/kernel/wave/utils/classes.py
@@ -3,6 +3,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+from enum import Enum
 from dataclasses import dataclass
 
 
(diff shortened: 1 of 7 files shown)


iree/turbine/kernel/wave/utils/classes.py
```diff
@@ -13,3 +14,15 @@ class KernelLaunchInfo:
...
     func_name: str = ""
     grid_str: str = ""
+
+
+############################################
+# Wave Ops related Utils
+############################################
+
+# GPU shuffle modes
+class ShuffleMode(Enum):
```
xintin, line 24: nit: can we have the same order as in handler?
idx, xor, up, down or this order there.

xintin:
Awesome! 
LGTM. :) 
Thanks.

xintin (approved):

iree/turbine/kernel/wave/utils/classes.py
raikonenfnu: ahh nice catch! :)

raikonenfnu: switched it up to the enum ordering of gpu dialect in MLIR which is xor, down, up, idx! :)

raikonenfnu: https://github.com/llvm/llvm-project/blob/8ba3a232d1985a75cf771f8832e911b6f8abe86b/mlir/include/mlir/Dialect/GPU/IR/GPUOps.td#L1284-L1287

PR #764, opened 2025-04-27
[Wave] Implement CDF and spec dec lower kernel in Wave

raikonenfnu:
- Fix bugs and add test for float cmps
- Fix bug and add test for RegisterOp when symbol is not same type as vector type/Register type
- Implement "lower" speculative decoding kernel in Wave

### DIFF
--- a/iree/turbine/kernel/wave/codegen/handlers.py
+++ b/iree/turbine/kernel/wave/codegen/handlers.py
@@ -130,6 +130,9 @@ def handle_register(emitter: WaveEmitter, node: fx.Node):
 
     if isinstance(value, sympy.Basic):
         value = gen_sympy_index(add_emitter_subs(emitter), value)
+        if value.type != vector_type.element_type:
+            conversion_op = get_conversion_op(value.type, vector_type.element_type)
+            value = conversion_op(vector_type.element_type, value)
         register = vector_d.splat(vector_type, value)
     else:
         register = arith_d.ConstantOp(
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -130,6 +130,9 @@ def handle_register(emitter: WaveEmitter, node: fx.Node):
 
     if isinstance(value, sympy.Basic):
         value = gen_sympy_index(add_emitter_subs(emitter), value)
+        if value.type != vector_type.element_type:
+            conversion_op = get_conversion_op(value.type, vector_type.element_type)
```
harsh-nod, line 134: why do we need this? Can we instead add an explicit cast in the kernel and maybe assert if the types are not the same?

raikonenfnu: > Can we instead add an explicit cast in the kernel and maybe assert if the types are not the same

Yes and no, the main problem we are facing right now is currently it is breaking in the tkl.Register initialization/ newRegister op initialization, so we won't have the chance to get to an explicit cast.
```python
> breaks here:
token_idx = tkl.Register[B, N, D, tkl.i32](THREAD_0) 
```
Since THREAD_0 output from `gen_sympy_index` is IndexType. 

Alternatively, I can add an assert that if we add a sympy expression as an input to tkl.Register/newRegister, it needs to be an IndexType, and then kernel author can cast it later on. This method may be safer but will result in slightly longer code though. What do you think?

harsh-nod: Sure, let's go with that then. I think it makes sense not to expose the index type to users as it has no mapping to anything in other languages.

harsh-nod (approved):
lgtm! thanks!

PR #768, opened 2025-04-29
[Wave] Add first speculative sampling kernel

harsh-nod:
This PR adds the first part of the speculative sampling kernel implemented in Wave. Multiple changes had to be made to produce a functional kernel, such as 
- Handling of multiple iterates 
- Adding support for sympy.Eq, invert operator
- Improvements to expansion 
- Making hoisting aware of setsymbols so that if a read has a mapping that uses a symbol that is being set, dont hoist the read before the setsymbol


nithinsubbiah (approved):
nice, this looks good to me

PR #769, opened 2025-04-29
[Wave] Plumb func_name into compile.py for wave asm

aviator19941:
Plumbs func_name through to the IR creation of `isolated_test_call` instead of hard-coding it to "isolated_benchmark".


raikonenfnu (approved):
noice!

PR #773, opened 2025-04-29
[Wave] Add wave_sdpa API for f16 attention

harsh-nod:
This PR adds a wave_sdpa API that is similar to the Pytorch API. Internally, it utilizes the vanilla attention template to get the kernel. 

The template was modified to do the scaling inside the kernel so that the query matrix no longer has to be scaled outside the kernel. This was modified in all the users of the vanilla template.

### DIFF
--- a/tests/kernel/wave/attention/vanilla_attention_test.py
+++ b/tests/kernel/wave/attention/vanilla_attention_test.py
@@ -119,10 +119,7 @@ def testTransposedVAttentionPure(
     k = device_randn(k_shape, dtype=torch.float16)
     v = device_randn(v_shape, dtype=torch.float16)
     output = device_zeros(o_shape, dtype=torch.float32)
-    log2e = 1.44269504089
-    dk_sqrt = math.sqrt(1.0 / shape.head_size)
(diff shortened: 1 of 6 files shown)


tests/kernel/wave/attention/vanilla_attention_test.py
```diff
@@ -202,10 +199,8 @@ def testAttentionPure(
     k = device_randn(k_shape, dtype=torch.float16)
     v = device_randn(v_shape, dtype=torch.float16)
     output = device_zeros(o_shape, dtype=torch.float32)
-    log2e = 1.44269504089
-    dk_sqrt = math.sqrt(1.0 / shape.head_size)
     # TODO: Add scaling of QK as part of kernel.
```
nithinsubbiah, line 202: nit: remove todo?

nithinsubbiah (approved):

tests/kernel/wave/attention/vanilla_attention_test.py
harsh-nod: Thanks will remove in next PR.

PR #775, opened 2025-04-30
[Wave] Add mlir style printing

harsh-nod:
This PR adds the ability to print fx.graphs in a manner similar to mlir graphs, with the caveat that not all the information about the fx.nodes is printed yet.


Hardcode84 (approved):
This is generally useful, we merge it even without updating any tests

PR #779, opened 2025-04-30
[TKW] Paged decode MHA using `GenericDot`

Hardcode84:


harsh-nod:
Can you add a description of what this commit is doing?

iree/turbine/kernel/_support/tracing.py
```diff
@@ -129,6 +132,8 @@ def create_arg(self, a):
             return a
         if isinstance(a, IndexMapping):
             return a
+        if isinstance(a, GenericDot):
```
harsh-nod, line 135: Could you describe the weird behavior that happens when you don't do this and add it as a TODO here? So this is not needed for GEMMs, but needed for Attention. TODO investigate further.

iree/turbine/kernel/wave/decompose_dot_mma.py
```diff
@@ -55,7 +57,7 @@ def is_dot_mma(node: fx.Node) -> bool:
             lhs.index = copy(lhs_index)
             rhs.index = copy(rhs_index)
 
-            k_sym = get_custom(lhs).indexing_dims[1]
+            k_sym = get_custom(lhs).type.symbolic_shape[-1]
```
harsh-nod, line 60: how did this work before?

iree/turbine/kernel/wave/decompose_reduce_ops.py
```diff
@@ -66,9 +66,9 @@ def determine_shuffle_config(
         WORKGROUP_1,
         WORKGROUP_2,
     ] + induction_vars
-    offset = access_pattern.start.subs({k: 0 for k in ignore})
+    offset = safe_subs(access_pattern.start, {k: 0 for k in ignore})
```
harsh-nod, line 69: why safe_subs vs subs?

harsh-nod (changes requested):
overall looks good, just request for some more documentation

iree/turbine/kernel/wave/decompose_dot_mma.py
Hardcode84: It should be `-1` from the start but instead I wrote `1` so it breaks for non 2D shapes

iree/turbine/kernel/_support/tracing.py
Hardcode84: This code is needed anyway as we are capturing something which is not a `fx.Node`, not sure why gemm worked without it, but I don't think we need any deep invesigation.

iree/turbine/kernel/wave/decompose_reduce_ops.py
Hardcode84: I was getting errors `start` is `int` and not a sympy object during my experiments.

harsh-nod (approved):
lgtm!

PR #783, opened 2025-05-01
[BOO] enable tuning specs for boo_convs

zjgarvey:
This PR adds support for a few options:

1. Specify an environment variable via `BOO_TUNING_SPEC_PATH`, in which case `get_launchable` will create a jit-compiling launchable that calls iree-compile through a subprocess and adds a compile flag for specifying a tuning spec file. 
2. Enables running the preload script with any compiler flags and logs stderr and stdout (to enable things like `--mlir-print-ir-after-all` or `--debug --debug-only=...`). 
3. In the future, we could also specify some additional compile flags for `get_launchable` if necessary. 

I'm opting not to modify the `device.compile_target_flags` because the in-process compiler is missing some necessary compiler flags.


iree/turbine/kernel/boo/conv_exports/preload.py
```diff
@@ -147,6 +152,36 @@ def get_cache_status(
...
+                    curr_dir / f"{get_device_from_torch(d).get_type_key_hash()}.vmfb"
+                )
+                if vmfb_path.is_file() and vmfb_path.stat().st_size > 0:
+                    continue
+                vmfb_path.unlink(missing_ok=True)
+                failed_devices.add(str(d))
+            if len(failed_devices) == 0:
+                continue
+            failed_funcs[fn] = f"failed compilation for devices: {failed_devices}"
+        return failed_funcs
```
rkayaith, line 183: shouldn't `run`/`_out_of_process_compile` just return whether it succeeded, instead of trying to recover the info later?

rkayaith (commented):
If the tuning spec path is going to use out of process compiles, is there any point to keeping the in-process path? There's a lot of paths through the compilation code, I think it'd be best to merge them as much as possible.
(imo it would be nice if the tuning spec path could be done in-process, but I don't have a strong opinion about that)

Can you also document the env vars in the README.

iree/turbine/kernel/boo/conv_exports/launch.py
```diff
@@ -55,19 +59,71 @@ def _out_of_process_compile(func_name, key_hashes_and_flags):
...
+        if vmfb_path.is_file():
+            logger.debug("found vmfb in cache: %s", str(vmfb_path))
+            continue
+        logger.debug("Compiling vmfb to cache: %s", str(vmfb_path))
+        cl_list = (
+            ["iree-compile"] + list(flags) + [f"'{mlir_path}'", "-o", f"'{vmfb_path}'"]
+        )
+        command = subprocess.list2cmdline(cl_list)
+        if len(flags) > 2:
+            (CACHE_BASE_DIR / func_name / f"compile_command.txt").write_text(command)
```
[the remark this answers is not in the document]
zjgarvey: I can always dump the compile command, but there are always two flags (backend + target) for default, and it felt wasteful to print the default command.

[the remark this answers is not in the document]
rkayaith: I see, is there an easy way to avoid the magic number? e.g. get the length of the default compile flags from device

if not, I'd just keep it simple and always dump the file

iree/turbine/kernel/boo/conv_exports/launch.py
```diff
@@ -48,26 +55,80 @@ def clear_cache_dir():
     shutil.rmtree(CACHE_BASE_DIR)
 
 
-def _out_of_process_compile(func_name, key_hashes_and_flags):
+def _out_of_process_compile(func_name, key_hashes_and_flags) -> Tuple[str, Tuple[bool]]:
```
rkayaith, line 58: could you type hint the arguments, here and elsewhere

iree/turbine/kernel/boo/conv_exports/launch.py
```diff
@@ -173,6 +234,16 @@ def get_launchable(
             parameter_providers=(),
             entry_point=f"{func_name}$async",
         )
+    elif Path(BOO_TUNING_SPEC_PATH).is_file():
```
rkayaith, line 237: I would explicitly check for empty string here, so file name typos don't get silently ignored
```suggestion
    elif BOO_TUNING_SPEC_PATH != "":
```

iree/turbine/kernel/boo/conv_exports/preload.py
```diff
@@ -243,8 +276,20 @@ def _get_preload_args():
...
         help="Specify a maximum number of concurrent processes.",
     )
-    return parser.parse_args()
+    # parser.add_argument(
+    #     "--extra-compile-flags",
+    #     "-f",
+    #     nargs="*",
+    #     type=str,
+    #     help="Specify additional compile flags to use when generating vmfbs.",
+    # )
```
rkayaith, line 285: should this be removed?

PR #787, opened 2025-05-02
Initial plumbing for Wave kernel integration with sharktank

aviator19941:
This is the initial plumbing for integrating Wave kernels into sharktank. The [wave IR](https://gist.github.com/aviator19941/062c502e1b05f2eadcf514feb8a76413) contains a func.func that calls a function in the stream.executable and this PR merges the wave IR and creates a target_function that calls the wave IR stream executable.

The Wave integration into sharktank makes use of the `mlir_kernel` decorator in sharktank by passing in the MLIR from `wave_compile` along with a `util.func` to call the FuncOp in the Wave asm. We need to add an `inline` pass [here](https://github.com/iree-org/iree-turbine/pull/787/files#diff-56777d4d58cb01861a15563fbccba4e55df2c83db374b5f3bca3c43d564a4097R143) in order to remove the private `func.func` and `util.func` calls in the MLIR to avoid having a `func.func @wave_kernel` that has the same body of the `func.func @main` function. The inline pass removes the duplicate `func.func @wave_kernel` and keeps the `func.func @main`.


tests/kernel/wave/common/utils.py
```diff
@@ -42,63 +39,3 @@ def param_bool(name, shortname=None, values=None):
     values = values or [False, True]
     ids = [f"{shortname}" if v else f"no_{shortname}" for v in values]
     return pytest.mark.parametrize(name, [pytest.param(v) for v in values], ids=ids)
-
-
-def scaled_dot_product_attention_bhsd(
```
raikonenfnu, line 47: Can we move this migration in a different PR? Would be nice to actually have one util file for all the reference kernels :)

raikonenfnu (commented):
Is it possible to test the wave kernel integration with sharktank?

tests/kernel/wave/common/utils.py
aviator19941: yeah I can move it to a different PR.

aviator19941:
> Is it possible to test the wave kernel integration with sharktank?

Yes, the PR is here: https://github.com/nod-ai/shark-ai/pull/1530, but depends on this one to be merged in first.

raikonenfnu:
@aviator19941 are there no way at all we can add a test to ensure these changes remain alive and well? :)

raikonenfnu:
The reason I ask for this is to ensure that we won't run into the case where someone on Wave side accidentally or purposely break this flow and only find out few days later on SharkTank, then we'd have to do a commit bisecting and stuff. So ideally, we have the safeguard here s.t people don't break this path to begin with :D

aviator19941:
> The reason I ask for this is to ensure that we won't run into the case where someone on Wave side accidentally or purposely break this flow and only find out few days later on SharkTank, then we'd have to do a commit bisecting and stuff. So ideally, we have the safeguard here s.t people don't break this path to begin with :D

Yeah that's a good point I will try to add a test

iree/turbine/kernel/wave/templates/vanilla_attention.py
```diff
@@ -648,6 +648,7 @@ def base_attention(
     ):
         base_attention_core(q, k, v, c)
 
+    # TODO: fix the signature and add custom mask test for bhsd case
```
raikonenfnu, line 651: why is there this comment out of nowhere haha

iree/turbine/runtime/op_reg/compiler.py
```diff
@@ -139,7 +139,9 @@ def compile_standalone_kernel(
 
     symb = func_name
     if config.async_invocations:
-        pm = PassManager.parse("builtin.module(torch-iree-func-conversion)", kb.context)
+        pm = PassManager.parse(
+            "builtin.module(inline, torch-iree-func-conversion)", kb.context
```
raikonenfnu, line 143: Can you add detail in the PR description on why we need inline?

lit_tests/kernel/wave/integration.py
```diff
@@ -0,0 +1,309 @@
...
+
+def _get_jinja2_env() -> Environment:
+    global _JINJA2_ENVIRONMENT
+    if _JINJA2_ENVIRONMENT is None:
+        _JINJA2_ENVIRONMENT = Environment(loader=BaseLoader())
+    return _JINJA2_ENVIRONMENT
+
+
+@CustomOp.register()
+class WaveBhsdFlashAttention(CustomOp):
```
raikonenfnu, line 64: Maybe we should call this filename or function name to indicate that this is indeed integration for sharktank

raikonenfnu: I do see `test_wave_aot_integration`, which is the right direction and would be even gooder to call out sharktank :D

iree/turbine/kernel/wave/templates/vanilla_attention.py
aviator19941: The custom mask case is not working for the bhsd case, and there is no test for the bhsd custom mask case, so I added a comment for the future

raikonenfnu (commented):
Overall LGTM, can @aviator19941 rebase and re-run th tests, once it passes, we can land :)

raikonenfnu (approved):

PR #794, opened 2025-05-05
[TKW] Tweak Paged Decode kernel waves/tile sizes

Hardcode84:
sglang decode numbers on large shapes `seq_len=100, (32, 128, 2, 512, 512)`:
```
phase_0: 77-> 35
phase_1: 38 -> 19
```

### DIFF
--- a/iree/turbine/kernel/wave/templates/paged_decode_attention.py
+++ b/iree/turbine/kernel/wave/templates/paged_decode_attention.py
@@ -73,7 +73,7 @@ def get_paged_decode_attention_kernels(
     PHASE_1_BLOCK_B = 64
     PHASE_1_ELEMS_PER_THREAD = PHASE_1_BLOCK_B // THREADS_PER_WAVE
     PHASE_1_BLOCK_N = 1
-    B_WAVES = 2
+    B_WAVES = 4
     HEAD_BLOCK_SIZE = 16 * B_WAVES
     head_ratio = shape.num_query_heads // shape.num_kv_heads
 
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/wave/templates/paged_decode_attention.py
```diff
@@ -69,12 +69,12 @@ class Phase(Enum):
         PHASE_1 = (1,)
 
     THREADS_PER_WAVE = 64
-    PHASE_1_BLOCK_B = 64
-    PHASE_1_ELEMS_PER_THREAD = PHASE_1_BLOCK_B // THREADS_PER_WAVE
-    PHASE_1_BLOCK_N = 1
-    HEAD_BLOCK_SIZE = 16
+    PHASE_1_BLOCK_B_WAVES = 1
+    PHASE_1_BLOCK_B = 64 * PHASE_1_BLOCK_B_WAVES
```
harsh-nod, line 73: Is this 64 because there are 64 threads per wave or just a tile size choice?

harsh-nod (commented):
Can you add the perf change and corresponding shape from this PR in the commit message?

iree/turbine/kernel/wave/templates/paged_decode_attention.py
Hardcode84: yes, because B is distributed along threads and it need to have at least 1 element_per_thread, I've tried larger `BLOCK_B` to have larger `element_per_thread` but it didn't improve the perf.

harsh-nod: makes sense, thanks.

PR #798, opened 2025-05-06
[Wave] Add ability to compare against flash_attn_func

harsh-nod:
This PR adds the ability to benchmark the flash_attn_func against wave kernels.


iree/turbine/kernel/wave/perf/benchmark.py
```diff
@@ -110,25 +125,48 @@ def main():
...
+        if flash_attn_enabled:
+            benchmark_sdpa(
+                shape_config["batch_size"],
+                shape_config["num_heads"],
+                shape_config["seq_len_q"],
+                shape_config["seq_len_k"],
+                shape_config["head_dim"],
+                args.num_warmup,
+                args.num_iterations,
+                flash_attn_func,
```
Hardcode84, line 167: nit: `flash_attn_func if flash_attn_enabled else wave_nn.functional.wave_sdpa` instead of copypasting entire call

Hardcode84 (approved):

iree/turbine/kernel/wave/perf/benchmark.py
harsh-nod: thanks will fix

PR #800, opened 2025-05-06
[BOO] Include tuning specs in pip package

zjgarvey:
Boo use was broken in pip install because the tuning spec file wasn't included.


rkayaith (approved):
thanks!

PR #802, opened 2025-05-06
[runtime] Enable async for `CustomOp` eager execution

zjgarvey:
This PR adds async invocation for `CustomOp` eager execution by applying the pass `--torch-iree-func-conversion` to the `FreeFuncKernelBuilder.module_op` to generate a `main$async` function, and factoring out the async invocation from `Launchable.__call__` into a common file `runtime/invoke.py`. 

In addition, this PR allows manually specifying that a `CustomOp` should be compiled as a single dispatch for eager execution.

Both of these will be useful in transitioning the boo conv launchables into custom ops without losing any functionality.

### DIFF
--- a/iree/turbine/runtime/op_reg/base.py
+++ b/iree/turbine/runtime/op_reg/base.py
@@ -184,6 +184,20 @@ class CustomOp(ABC):
         """
         ...
 
+    @property
+    def single_dispatch(self) -> bool:
+        """Indicates whether the CustomOp should be forced into a single dispatch using a util.func pipeline attribute.
+
+        It is recommended to only use this for more complicated ops which would not automatically get compiled into a single dispatch.
+        E.g. A fused conv + bias-add + relu custom op.
+
+        For eager contexts, this will apply the pipeline attribute to the main$async function.
+
+        For aot contexts, this currently does nothing, but could eventually attempt to apply an `util.inline.never` attribute,
+        in addition to the pipeline attribute, to the function being called by the InlineKernelBuilder.
+        """
+        return False
+
     @abstractmethod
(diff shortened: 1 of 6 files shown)


AWoloszyn (approved):
Looks good on the runtime side.

iree/turbine/runtime/op_reg/base.py
```diff
@@ -184,6 +184,20 @@ def signature(self) -> str:
...
+
+        It is recommended to only use this for more complicated ops which would not automatically get compiled into a single dispatch.
+        E.g. A fused conv + bias-add + relu custom op.
+
+        For eager contexts, this will apply the pipeline attribute to the main$async function.
+
+        For aot contexts, this currently does nothing, but could eventually attempt to apply an `util.inline.never` attribute,
+        in addition to the pipeline attribute, to the function being called by the InlineKernelBuilder.
+        """
+        return False
```
rkayaith, line 199: out of curiosity, what's the reason to default this to `False`?

rkayaith (approved):

iree/turbine/runtime/op_reg/base.py
zjgarvey: Only because I'm not sure if this pass pipeline is appropriate for anything other than our convs right now, and most `CustomOp`'s are simple enough to compile to a single dispatch as-is.

I don't think we use `CustomOp` eager execution in any downstream projects, but I'd like to avoid changing the compilation behavior significantly in case it could cause problems elsewhere. Someone might have a `CustomOp` in sharktank, for example, which really can't form into a single dispatch.

PR #803, opened 2025-05-06
Add support for additional type interop

rsuderman:
Interop for BF16 / FP8 requires some munging to bypass numpy's lack of support. In these cases we bitcast the underlying tensor and use the integer equivalent bitwidths

### DIFF
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -50,6 +50,7 @@ __all__ = [
     "DeviceState",
 ]
 
+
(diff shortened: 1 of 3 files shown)


zjgarvey (approved):
Looks good to me. 

Might be good to test the float8 types on cpu, too.

rsuderman:
> Looks good to me.
> 
> Might be good to test the float8 types on cpu, too.

I agree, sadly `torch` doesn't support `fp8` operations on `cpu`.

PR #805, opened 2025-05-07
[Wave] Enable using buffer ops for contiguous loads

harsh-nod:
This PR enables the use of buffer ops for contiguous load and stores. Previously, buffer ops could only be used for gathers and scatters.


Hardcode84:
Can you check on extend attention? I think that one had the perf issues with buffer ops.

PR #808, opened 2025-05-07
[TKW] Avoid cache folder race conditions when running tests in parallel

Hardcode84:
Even if caching is disabled, we are using cache dir for the wave_runtime and hsaco files.

Assign each worker an unique dir to avoid race conditions.

### DIFF
--- a/iree/turbine/kernel/wave/utils/run_utils.py
+++ b/iree/turbine/kernel/wave/utils/run_utils.py
@@ -242,15 +242,16 @@ def invoke_with_wave_runtime(
     Invokes the kernel with the wave runtime.
     """
     import wave_runtime
-    from ..cache import WAVE_RUNTIME_DIR, CACHE_BASE_DIR
+    from ..cache import get_wave_runtime_dir, get_cache_base_dir
 
     # Get the path to the binary.
     if options.kernel_hash:
         binary = (
-            str(CACHE_BASE_DIR / options.kernel_hash / options.kernel_hash) + ".hsaco"
+            str(get_cache_base_dir() / options.kernel_hash / options.kernel_hash)
+            + ".hsaco"
         )
     else:
-        binary = glob.glob(str(WAVE_RUNTIME_DIR / "*.hsaco"))[0]
+        binary = glob.glob(str(get_wave_runtime_dir() / "*.hsaco"))[0]
 
     dynamic_dims = tuple(options.dynamic_symbols_map.values())
     # Update the grid size as this may vary depending
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,23 +47,42 @@ def pytest_configure(config):
     )
 
 
+def _get_worker_id(config):
+    if not hasattr(config, "workerinput"):
+        return None
+
+    worker_id = config.workerinput["workerid"]
+    if not worker_id.startswith("gw"):
+        return None
+
+    return int(worker_id[2:])
+
+
 def _set_default_device(config):
     distribute = int(config.getoption("--gpu-distribute"))
     if distribute < 1:
         return
 
-    if not hasattr(config, "workerinput"):
+    worker_id = _get_worker_id(config)
+    if worker_id is None:
         return
 
-    worker_id = config.workerinput["workerid"]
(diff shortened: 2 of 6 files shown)


marbre (commented):
Do those test dirs need to / get cleaned up somewhere?

Hardcode84:
> Do those test dirs need to / get cleaned up somewhere?

We are already doing it https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/ci-tk.yaml#L97

tests/conftest.py
```diff
@@ -47,31 +47,59 @@ def pytest_configure(config):
     )
 
 
+def _get_worker_id(config):
+    if not hasattr(config, "workerinput"):
```
harsh-nod, line 51: What does the attribute workerinput specify? Can you add a comment explaining it? When would you not have a workerinput attribute?

tests/conftest.py
```diff
@@ -47,31 +47,59 @@ def pytest_configure(config):
...
+        return None
+
+    worker_id = config.workerinput["workerid"]
+    if not worker_id.startswith("gw"):
```
harsh-nod, line 55: Also what does the prefix gw specify? Can you add a comment?

tests/conftest.py
```diff
@@ -47,31 +47,59 @@ def pytest_configure(config):
...
+        return None
+
+    return int(worker_id[2:])
```
harsh-nod, line 58: Also here, what is the magic number 2 here? Comment would be helpful.

harsh-nod (changes requested):
thanks for fixing this. Overall looks good, just request for comments

tests/conftest.py
Hardcode84: Added some code comments, but it's mostly just internal `pytest` APIs

harsh-nod (approved):
lgtm! please add comments and land

iree/turbine/kernel/wave/utils/run_utils.py
```diff
@@ -242,15 +242,16 @@ def invoke_with_wave_runtime(
...
-    from ..cache import WAVE_RUNTIME_DIR, CACHE_BASE_DIR
+    from ..cache import get_wave_runtime_dir, get_cache_base_dir
 
     # Get the path to the binary.
     if options.kernel_hash:
         binary = (
-            str(CACHE_BASE_DIR / options.kernel_hash / options.kernel_hash) + ".hsaco"
+            str(get_cache_base_dir() / options.kernel_hash / options.kernel_hash)
+            + ".hsaco"
         )
```
ScottTodd, line 252: drive-by: Does Wave _only_ support hsaco? IREE (and iree-turbine) support multiple targets and kernel formats. Could use a generic file extension or a container format like IREE's VMFB (or other flatbuffers: https://github.com/iree-org/iree/tree/main/runtime/src/iree/schemas).

This could also use [`PurePath.with_suffix`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.with_suffix) instead of appending to a string.

tests/conftest.py
```diff
@@ -47,31 +47,66 @@ def pytest_configure(config):
...
 
     import iree.turbine.kernel.wave.utils.general_utils as general_utils
 
     general_utils.DEFAULT_GPU_DEVICE = device_id
 
 
+def _set_cache_dir(config):
+    """
+    Sets the unique cache directory for the current worker to avoid race conditions.
+    """
```
ScottTodd, line 90: What sorts of race conditions? If tests have race conditions, user applications may too. Can the cache be made more multi-thread or multi-process friendly?

iree/turbine/kernel/wave/utils/run_utils.py
Hardcode84: Standalone wave runtime only supports hip/hsaco currently. IREE-based runtime doesn't need this code, but it has perf problems with dlpack

tests/conftest.py
Hardcode84: wave runtime path is writing temp files into the cache dir with potentially colliding names, so you run multiple instances running in parallel you will get into the troubles.

PR #809, opened 2025-05-07
[TKW] Paged Decode dynamic symbols

Hardcode84:
Make `kv_lens` and `num_seqs` args of the Paged Decode kernel dynamic, so they won't trigger the recompilation.


harsh-nod:
Can you add a description in the commit that briefly describes the changes?

harsh-nod (approved):
lgtm!

PR #821, opened 2025-05-11
[TKW] Allow to bind symbols from the scalar arguments

Hardcode84:
I'm planning to retire `dynamic_symbols_map` eventually and extract symbol values directly from the input tensor dimensions. One exception is `extend_attention` which have `MAX_EXTEND_SEQ_LEN` symbol which does not correspond to any input tensor dimension.

As we now support scalar codegen, instead of using `dynamic_symbols_map` we now can just pass `MAX_EXTEND_SEQ_LEN` directly as kernel argument.

Extend attention is using `MAX_EXTEND_SEQ_LEN` for grid dimension calculation so normal scalar + `set_symbol` won't work. Instead, introduce `SymbolBind` annotation, which works exactly like scalar arg, but also immediately binds value to the corresponding symbol.

Most of the codegen changes are to propagate scalar args to the dispatch dims and to generally support scalar args in iree runtime path.

### DIFF
--- a/iree/turbine/kernel/_support/tracing.py
+++ b/iree/turbine/kernel/_support/tracing.py
@@ -31,7 +31,7 @@ from ..lang.grid import Grid
 from ..lang.types import (
     Index,
 )
-from ..lang.wave_types import IndexMapping
+from ..lang.wave_types import IndexMapping, SymbolBind
 from ..ops.wave_ops import CustomOp
 
 from .regions import RegionGraph, SubgraphTracer
@@ -44,6 +44,8 @@ from ..ops.base import (
 
 from . import context
 from .dtype import DataType
+
+import inspect
 
 try:
     from typing import assert_type
@@ -61,12 +63,22 @@ TCallable = TypeVar("TCallable", bound=Callable)
 
 
 class KernelRegionGraph(RegionGraph):
+    func: Optional[Callable] = None
+
+    def __init__(
+        self,
+        location_capture_config: Optional[LocationCaptureConfig] = None,
+        func: Optional[Callable] = None,
+    ):
+        super().__init__(location_capture_config=location_capture_config)
+        self.func = func
+
     def new_subtracer(
         self,
         region_graph: "RegionGraph",
         parent: Optional["SubgraphTracer"] = None,
     ) -> "KernelTracer":
-        return KernelTracer(region_graph, parent=parent)
+        return KernelTracer(region_graph, parent=parent, func=self.func)
 
 
 ###############################################################################
@@ -99,6 +111,18 @@ class KernelBufferProxy(fx.Proxy):
 class KernelTracer(SubgraphTracer):
     """Custom Tracer for generating a trace of a kernel computation."""
 
+    arg_names: list[str] = []
+
+    def __init__(
+        self,
+        region_graph: RegionGraph,
+        parent: Optional["SubgraphTracer"] = None,
+        func: Optional[Callable] = None,
+    ):
+        super().__init__(region_graph, parent)
+        if func is not None:
+            self.arg_names = inspect.getfullargspec(func).args
+
     # Property to keep track of current number of arguments.
     current_arg_id = 0
 
@@ -110,6 +134,12 @@ class KernelTracer(SubgraphTracer):
             if isinstance(t, DataType) and node.op == "placeholder":
                 node.meta["arg_id"] = self.current_arg_id
                 node.meta["dtype"] = t
+                node.meta["symbolic_type"] = []
+                self.current_arg_id += 1
+            elif issubclass(t, SymbolBind) and node.op == "placeholder":
+                node.meta["arg_id"] = self.current_arg_id
+                node.meta["symbol_name"] = self.arg_names[self.current_arg_id]
+                node.meta["dtype"] = t.dtype
                 node.meta["symbolic_type"] = []
                 self.current_arg_id += 1
             elif isinstance(t, KernelBufferMeta):
--- a/iree/turbine/kernel/_support/tracing.py
+++ b/iree/turbine/kernel/_support/tracing.py
@@ -136,7 +136,10 @@ class KernelTracer(SubgraphTracer):
                 node.meta["dtype"] = t
                 node.meta["symbolic_type"] = []
                 self.current_arg_id += 1
-            elif issubclass(t, SymbolBind) and node.op == "placeholder":
+            elif issubclass(t, SymbolBind):
+                assert (
+                    node.op == "placeholder"
+                ), "SymbolBind must be a placeholder, got {node.op}"
                 node.meta["arg_id"] = self.current_arg_id
                 node.meta["symbol_name"] = self.arg_names[self.current_arg_id]
                 node.meta["dtype"] = t.dtype
--- a/iree/turbine/kernel/compiler/dispatch_codegen.py
+++ b/iree/turbine/kernel/compiler/dispatch_codegen.py
@@ -178,6 +178,9 @@ class StreamExecutable:
                         )
                     )
 
+            def is_scalar_symbol(binding: BindingDesc) -> bool:
+                return binding.symbol_type is not None
+
             # Define the export.
             with InsertionPoint.at_block_begin(self._exe_block):
                 index_type = IndexType.get()
@@ -187,7 +190,7 @@ class StreamExecutable:
                         [
                             index_type
                             for b in dynamic_dim_bindings
-                            + [b for b in scalar_bindings if b.symbol_type is not None]
+                            + [b for b in scalar_bindings if is_scalar_symbol(b)]
                         ]
                     )
                 )
--- a/iree/turbine/kernel/compiler/dispatch_codegen.py
+++ b/iree/turbine/kernel/compiler/dispatch_codegen.py
@@ -24,14 +24,15 @@ from .ir import (
     IndexType,
     InsertionPoint,
     IntegerAttr,
+    IntegerType,
     IrType,
     Location,
     StringAttr,
     Value,
     arith_d,
     func_d,
-    stream_d,
     iree_codegen_d,
+    stream_d,
 )
 
 from .kernel_codegen import (
@@ -140,7 +141,6 @@ class StreamExecutable:
 
         with self._loc:
             binding_type = IrType.parse("!stream.binding")
-            index_type = IndexType.get()
 
             # Define the dispatch function.
             def abi_type(binding: BindingDesc):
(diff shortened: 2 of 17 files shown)


iree/turbine/kernel/_support/tracing.py
```diff
@@ -111,6 +131,12 @@ def proxy(self, node: fx.Node) -> fx.Proxy:
                 node.meta["dtype"] = t
                 node.meta["symbolic_type"] = []
                 self.current_arg_id += 1
+            elif issubclass(t, SymbolBind) and node.op == "placeholder":
```
harsh-nod, line 134: Could you instead do
```
elif issubclass(t, SymbolBind):
      assert node.op == "placeholder" ...
```
otherwise SymbolBinds on non-placeholders will silently fail.

iree/turbine/kernel/_support/tracing.py
```diff
@@ -111,6 +131,12 @@ def proxy(self, node: fx.Node) -> fx.Proxy:
...
+                node.meta["arg_id"] = self.current_arg_id
+                node.meta["symbol_name"] = self.arg_names[self.current_arg_id]
```
harsh-nod, line 136: Why do you need the symbol_name? Isn't the current_arg_id a unique identifier for the argument?

iree/turbine/kernel/compiler/dispatch_codegen.py
```diff
@@ -24,14 +24,15 @@
     IndexType,
     InsertionPoint,
     IntegerAttr,
+    IntegerType,
```
harsh-nod, line 27: is this being used?

iree/turbine/kernel/compiler/dispatch_codegen.py
```diff
@@ -181,9 +181,16 @@ def abi_type(binding: BindingDesc):
...
             with InsertionPoint.at_block_begin(self._exe_block):
+                index_type = IndexType.get()
                 export_op = stream_d.ExecutableExportOp(name, name)
                 export_block = export_op.workgroup_count.blocks.append(
-                    *([b.as_mlir_type() for b in dynamic_dim_bindings])
+                    *(
+                        [
+                            index_type
+                            for b in dynamic_dim_bindings
+                            + [b for b in scalar_bindings if b.symbol_type is not None]
```
harsh-nod, line 191: why do you need
```
if b.symbol_type is not None
```
?

iree/turbine/kernel/compiler/dispatch_codegen.py
```diff
@@ -193,16 +200,22 @@ def abi_type(binding: BindingDesc):
...
             }
 
             with InsertionPoint(workgroup_builder.entry_block):
-                result_type = IndexType.get()
+                for i, s in enumerate(scalar_bindings):
+                    if s.symbol_type is None:
+                        continue
+
+                    offset = len(dynamic_dim_bindings) + i
+                    dynamic_symbols_mapping[s.symbol_type] = arguments[offset]
```
harsh-nod, line 216: Why not create a separate mapping for scalar_bindings? This seems to be using dynamic_symbols_mapping to hold both dynamic and scalar bindings?

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -6,21 +6,24 @@
...
+    IntegerType,
     IrType,
     Location,
-    ArrayAttr,
-    SymbolRefAttr,
     MemRefType,
     RankedTensorType,
+    SymbolRefAttr,
+    Value,
+    arith_d,
```
harsh-nod, line 24: thanks!

iree/turbine/kernel/compiler/host_codegen.py
```diff
@@ -84,18 +89,40 @@ def isolated_test_call(
...
         entry_block = func_op.add_entry_block(arg_locs)
-        offset = len(sig.kernel_buffer_bindings)
-        dynamic_argument_map = {
-            k: v for k, v in zip(dynamic_symbols, entry_block.arguments[offset:])
-        }
+        scalars_offset = len(sig.kernel_buffer_bindings)
+        scalars_count = len(scalar_bindings)
+        dynamic_offset = scalars_offset + scalars_count
+
+        def to_index(v: Value) -> Value:
```
harsh-nod, line 104: Can you pull this nested function out? breaks the flow when reading the code. Also might I suggest renaming to maybe_cast_to_index ? Might be worthwhile to add this to a utils file since it seems handy for other tasks as well.

iree/turbine/kernel/wave/compile.py
```diff
@@ -176,4 +187,6 @@ def get_binary_path():
     if options.wave_runtime:
         binary_path = get_binary_path()
 
-    return WaveKernel(options, compiled_wave_vmfb, asm, binary_path)
+    return WaveKernel(
+        options, compiled_wave_vmfb, asm, binary_path, bound_scalar_symbols
+    )
```
harsh-nod, line 192: Do we need to add this to the cache? What do we need to change to caching with this change?

iree/turbine/kernel/wave/utils/run_utils.py
```diff
@@ -50,24 +52,25 @@ def get_device_uuid(device_list: list[str], device_str: str) -> tuple[int, str]:
...
-    for input in inputs:
+    for input in chain(inputs, scalar_args, dynamic_dims):
         if isinstance(input, torch.Tensor):
             input_cpu = input.cpu().contiguous()
             device_array = rt.asdevicearray(device, input_cpu)
             arg_list.push_ref(device_array._buffer_view)
+        elif isinstance(input, int):
+            arg_list.push_int(input)
+        elif isinstance(input, float):
+            warn("Currently, `push_float` is not working on the iree side")
```
harsh-nod, line 69: Why not throw an error here?

iree/turbine/kernel/wave/wave.py
```diff
@@ -175,6 +174,16 @@ def __init__(
 
         self.grid_type = Grid[tuple(get_grid_shape(self.workgroup_constraints))]
 
+        # TODO: needed for the wave_runtime grid calculations, we should really
+        # just generate host wrapper suitable for wave_runtime instead of doing
+        # it in python (and it will be faster as well).
+        hints = get_type_hints(eager_function)
```
harsh-nod, line 180: Wait why do we need type hints? Seems brittle to depend on type hints to get name and arg?

harsh-nod (changes requested):
thanks. Had some questions about how this works with caching and the need for type hints.

iree/turbine/kernel/wave/wave.py
Hardcode84: See my prev comment about `MAX_EXTEND_SEQ_LEN: tkl.SymbolBind[tkl.i32]`.

iree/turbine/kernel/wave/utils/run_utils.py
Hardcode84: To be able to write negative test and actually detect if it suddenly starts working.

iree/turbine/kernel/wave/compile.py
Hardcode84: It's constructed in `LaunchableWave` init and always available so we don't need to cache it.

iree/turbine/kernel/_support/tracing.py
Hardcode84: done

iree/turbine/kernel/compiler/dispatch_codegen.py
Hardcode84: removed

iree/turbine/kernel/compiler/dispatch_codegen.py
Hardcode84: `b.symbol_type is None` means it's just a normal scalar arg without any attached symbol and they cannot participate in grid calculations.

iree/turbine/kernel/compiler/dispatch_codegen.py
Hardcode84: It's used in `gen_sympy_index` call in few lines so even if we had a separate mapping we'd need to merge them before the call.

iree/turbine/kernel/compiler/host_codegen.py
Hardcode84: Moved func outside for now. I don't think we have a good common codegen utils module yet.

iree/turbine/kernel/compiler/dispatch_codegen.py
harsh-nod: Could you create a function with your explanation above:
```
def _is_scalar_and_not_symbol_bound(b: ...):
       """
        ..... b.symbol_type is None means it's just a normal scalar arg without any attached symbol and they cannot participate in grid calculations ....
       """
       return b.symbol_type is not None
```

which would make it more clear?

iree/turbine/kernel/compiler/dispatch_codegen.py
harsh-nod: okay makes sense, then I would just rename this variable to make it clear that it is not just dynamic symbols.

iree/turbine/kernel/wave/wave.py
harsh-nod: Makes sense. Are there any performance slowdowns caused by this function vs using __annotations__ ?

iree/turbine/kernel/wave/compile.py
harsh-nod: great

iree/turbine/kernel/wave/utils/graph_utils.py
```diff
@@ -59,6 +59,11 @@ def is_removable_operator(node: fx.Node) -> bool:
         if custom.users or custom.has_side_effects or is_global_write(node):
             return False
 
+        if isinstance(custom, Placeholder) and custom.graph == trace.get_root_graph():
+            # Do not remove root placeholders as they correspond to kernel
+            # arguments and removind them will change the kernel signature.
```
harsh-nod, line 64: minor typo: removing

iree/turbine/kernel/wave/utils/run_utils.py
harsh-nod: Got it, I see this is tested in test_scalar_codegen.

harsh-nod (changes requested):
I think this is mostly good to go. My biggest question had to do with any performance implications with get_type_hints and whether annotations accomplishes the same.The others are mostly refactor requests.

iree/turbine/kernel/wave/wave.py
Hardcode84: Nothing I immediately noticed. Also, this code is only called once when we create `LaunchableWave` object from `@tkw.wave` decorator, not when we actually invoking kernel, so it isn't on the hot path anyway.

iree/turbine/kernel/compiler/dispatch_codegen.py
Hardcode84: Renamed to `symbols_mapping`, but not sure if it's much better.

iree/turbine/kernel/compiler/dispatch_codegen.py
Hardcode84: done

iree/turbine/kernel/compiler/host_codegen.py
harsh-nod: thanks

harsh-nod (approved):
thanks! should be a nice addition to the language :)

PR #822, opened 2025-05-12
[Wave] Minimize shared allocs

harsh-nod:
This PR adds a more comprehensive approach to optimizing shared memory allocations. Shared memory allocations are derived from a single 1D source and offsets from where they are read is determined based on live range analysis of the allocations. More detailed information is provided in the documentation.


iree/turbine/kernel/wave/memory_analysis/minimize_shared_allocs.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    live_intervals = {}
+    for alloc in allocs:
+        live_intervals[alloc] = LiveInterval()
+        users, _ = get_users(alloc, None)
+        for user in users:
+            if user._sort_key < live_intervals[alloc].start:
+                live_intervals[alloc].start = user._sort_key
+            if user._sort_key > live_intervals[alloc].end:
+                live_intervals[alloc].end = user._sort_key
```
raikonenfnu, line 75: Doesn't the `_sort_key` run into the nested graph issue we were talking about?

harsh-nod: No because we renumber all the nodes first, see update_sort_keys above.

iree/turbine/kernel/wave/memory_analysis/minimize_shared_allocs.py
```diff
@@ -0,0 +1,179 @@
...
+    return live_intervals
+
+
+def get_shared_memory_allocation_size(alloc: fx.Node) -> tuple[int, int, int]:
```
Hardcode84, line 79: nit: result type should be just `int`?

Hardcode84:
I think this code relies on allocs having the same dtype. Can you add a test when you have both f16 and f32 allocs?

harsh-nod:
> I think this code relies on allocs having the same dtype. Can you add a test when you have both f16 and f32 allocs?

Good catch. For now, I will add an assert that disallows this, but I can fix in a future commit.

iree/turbine/kernel/wave/memory_analysis/minimize_shared_allocs.py
harsh-nod: thanks, will fix.

harsh-nod:
> > I think this code relies on allocs having the same dtype. Can you add a test when you have both f16 and f32 allocs?
> 
> Good catch. For now, I will add an assert that disallows this, but I can fix in a future commit.

Based on our discussion, I have moved to the view op which should allow this functionality and help us generalize.

iree/turbine/kernel/wave/memory_analysis/minimize_shared_allocs.py
```diff
@@ -0,0 +1,187 @@
...
+def minimize_shared_allocs(trace: CapturedTrace, minimize_shared_allocs: bool):
+    """
+    Minimize the number of shared allocs by reusing them.
+    See: iree/turbine/docs/kernel/memory_analysis.md for more details.
+    """
+    if not minimize_shared_allocs:
+        return
+    update_sort_keys(trace, trace.get_root_graph())
+
+    allocs = trace.walk(lambda x: isinstance(get_custom(x), Allocate))
```
Hardcode84, line 132: Probably better to check `address_space==SHARED_ADDRESS_SPACE` too, even if it can't be anything else currently.

Hardcode84 (approved):
LGTM. Any perf numbers?

iree/turbine/kernel/wave/memory_analysis/minimize_shared_allocs.py
harsh-nod: okay will add in subsequent PR.

harsh-nod:
> LGTM. Any perf numbers?

Nothing yet, but will look into that next.

PR #827, opened 2025-05-15
[BOO] Add autocast functionality to `boo_conv`

zjgarvey:
This enables running lower-precision forward and backward kernels when performing the forward pass of a model in a `torch.amp.autocast` context.

### DIFF
--- a/iree/turbine/kernel/boo/ops/conv.py
+++ b/iree/turbine/kernel/boo/ops/conv.py
@@ -19,14 +19,13 @@ class _BooConv(torch.autograd.Function):
     def forward(ctx, x, w, b=None, kwargs={}):
         x = x.detach()
         w = w.detach()
+        device_type = x.device.type
+
+        do_autocast = device_type == "cuda" and torch.is_autocast_enabled("cuda")
 
         # set original dtype and autocast dtype
         _og_dtype = x.dtype
-        _dtype = (
-            w.dtype
-            if not torch.is_autocast_enabled()
-            else torch.get_autocast_gpu_dtype()
-        )
+        _dtype = w.dtype if not do_autocast else torch.get_autocast_dtype("cuda")
 
         # convert everything to w.dtype or autocast dtype
         x = x.to(dtype=_dtype)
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/boo/ops/conv.py
```diff
@@ -15,37 +15,77 @@
...
         x, w = ctx.saved_tensors
         args = (x, w)
         kwargs = ctx.kwargs
 
+        # get original and autocast dtypes
+        _og_dtype = grad_output.dtype
+        _dtype = (
+            _og_dtype
+            if not torch.is_autocast_enabled()
+            else torch.get_autocast_gpu_dtype()
```
rkayaith, line 71: should `torch.get_autocast_dtype` be used here instead of hardcoding to `gpu`?

tests/kernel/boo/ops/boo_conv_test.py
```diff
@@ -51,6 +57,45 @@ def testBooConvBackwardsWithBias(self):
             b = torch.ones([1], dtype=torch.float32, device=device, requires_grad=True)
             torch.autograd.gradcheck(boo_conv, (x, w, b), atol=1e-5, eps=1e-3)
 
+    def testBooConvBackwardsAmpContext(self):
+        with tempfile.TemporaryDirectory() as td:
+            set_boo_cache(Path(td))
+            device = "cuda:0" if torch.cuda.is_available() else None
```
rkayaith, line 63: should this be two separate tests? with the gpu one skipped if cuda isn't available

rkayaith (approved):
just some minor comments

tests/kernel/boo/ops/boo_conv_test.py
zjgarvey: I wanted to make sure the autocast context didn't apply when it wasn't relevant. I.e., on CPU we should still be performing float32 kernel launches.

zjgarvey: Maybe separating them is a good idea though.

zjgarvey:
Having some trouble getting this to do what I expect on cpu vs. gpu.

Might need to figure something more sophisticated out for this to work, so I'll just have this only work with gpu autocast for now.

tests/kernel/boo/ops/boo_conv_test.py
zjgarvey: I separated them. Interestingly enough, if `cuda.is_available()` and you do the cpu test, it will actually still do the autocasting for cpu tensors and launch a bfloat16 cpu kernel. I fixed this behavior by also verifying the device type of the input args before casting.

zjgarvey:
I tried registering with `amp.custom_fwd`/`amp.custom_bwd` for multiple device types and it fails (which is rather odd).

I'm going to only enable for gpu for now, and when I go about the `CustomOp` implementation, as opposed to the `Launchable` implementation, I'll see if it makes more sense how to do this cleanly.

rkayaith (approved):
still looks good 😁

PR #828, opened 2025-05-15
[BOO] Enable bad kernel launch fix + kwarg filtered replacement

zjgarvey:
Adds a compiler flag to fix certain bad kernel launches.
Allows replacing only those `Conv2d` modules which match provided kwargs.


zjgarvey:
This flag needs a newer version of IREE. Will try again after the next version bump.

MaheshRavishankar (approved):

PR #829, opened 2025-05-15
[Wave] Introduce softsign kernel to replace tanh_approx

yichiche:
Add a new `softsign` kernel variant that replaces the previous `tanh_approx` implementation. Benchmarks show a 10~15% speedup over `tanh_approx`, but with a modest accuracy drop (e.g. Grok performance falls from ~83% to ~79%). This commit provides the softsign implementation alongside existing approximations for users to choose the appropriate tradeoff.

### DIFF
--- a/iree/turbine/kernel/wave/codegen/handlers.py
+++ b/iree/turbine/kernel/wave/codegen/handlers.py
@@ -85,6 +85,7 @@ from ...ops.wave_ops import (
     tanh,
     tanh_approx,
     workgroup_barrier,
+    softsign,
 )
 from ...compiler.base import CodegenError, ValidationError, NDEBUG
 from ...compiler.builder import IRProxyValue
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -816,6 +817,55 @@ def handle_tanh_approx(source: Value, options: WaveCompileOptions) -> OpResult:
...
+    """
+    element_type = get_type_or_element_type(source.type)
+
+    if not _is_float_type(element_type):
+        raise ValidationError(
+            f"Unhandled operand type for logit_softsign: {element_type}"
+        )
+
+    # Constants
+    scale_s = 128 ** (-0.5)  # for Grok: scale factor based on shape
```
raikonenfnu, line 836: any thoughts on making 128 an argument? (it doesn't have to be compile time, can just literally be a python int as well)

raikonenfnu: So in fact, I was thinking the opposite of using constant_scale. I think it would make it head_dim as an function argument to softsign e.g
`tkw.softsign(logit, head_dim=128)`

raikonenfnu: but the more I think about it, isn't the scaling of `sqrt(head_dim)` and logit_cap not part of the softsign formulation? does it make more sense to move this and the logit_cap(30) to be out of this op?

yichiche: Agreed. We could introduce a flag to control whether to apply scaling by the constant if the input hasn’t been normalized yet before computing the actual softsign. For example:
if apply_scaling:
    # assert head_dim is not None, "head_dim must be set if apply_scaling=True"
    scale = head_dim ** -0.5
    x = arith_d.mulf(x, ConstantOp(scale), fastmath=...)
# softsign starts
...

raikonenfnu: sounds good! let's do that! :)

raikonenfnu (approved):

PR #832, opened 2025-05-15
[Wave] Add scheduling/synchronization related ops

raikonenfnu:
- Implement missing scheduling ops such as s_barrier and set_prio
- Made changes to index sequence analysis so can be exposed to users.

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -70,7 +70,15 @@ def extract_slice(
     ...
 
 
+def set_prio(priority: int):
+    ...
+
+
 def shared_memory_barrier():
+    ...
+
+
+def workgroup_barrier():
     ...
 
 
@@ -1121,6 +1129,18 @@ class SelfIndex(CustomOp):
         return Register[(self.dim, self.dtype)]
 
 
+@define_op("set_prio")
+@dataclass
+class SetPrio(CustomOp):
+    """
+    An op that sets/tells hardware what level of priority certain instructions/region is.
+    This is useful for ping-pong or general case where two Waves share the same SIMD, but
+    we want to tell the SIMD to prioritize on wave or the other.
+    """
+
+    priority: int
+
+
 @define_op("shared_memory_barrier")
 @dataclass
 class SharedMemoryBarrier(CustomOp):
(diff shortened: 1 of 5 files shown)


iree/turbine/kernel/wave/utils/graph_utils.py
```diff
@@ -59,7 +61,10 @@ def is_removable_operator(node: fx.Node) -> bool:
 
         if (
             custom.users
-            or isinstance(custom, (Output, SetSymbol, SharedMemoryBarrier))
+            or isinstance(
+                custom,
+                (Output, SetSymbol, SetPrio, SharedMemoryBarrier, WorkgroupBarrier),
```
Hardcode84, line 66: We really need some trait for it :)

Hardcode84 (approved):

iree/turbine/kernel/wave/utils/graph_utils.py
nithinsubbiah, line 66: set as an op property

raikonenfnu: will add in this PR and end it once and for all :D

nithinsubbiah (approved):
apart from setting skipping DCE as a property, looks good to me

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -1121,6 +1129,18 @@ def type(self) -> "Register":
         return Register[(self.dim, self.dtype)]
 
 
+@define_op("set_prio")
+@dataclass
+class SetPrio(CustomOp):
```
harsh-nod, line 1134: Can I suggest you rename this to SetWavePriority? We dont have to stick with the instruction name and this really clarifies what is happening.

iree/turbine/kernel/wave/utils/graph_utils.py
raikonenfnu: done

raikonenfnu: done

iree/turbine/kernel/ops/wave_ops.py
raikonenfnu: makes sense, done!

iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -721,6 +729,10 @@ def type(self) -> Any:
     def type(self, value: Any):
         self.fx_node.type = value
 
+    @property
+    def skip_dce(self) -> bool:
```
Hardcode84, line 733: Maybe something like `has_side_effects`? Also I think it will be safer to set `has_side_effects` to True by default and then set to False for the ops we know have no effects.

raikonenfnu: that will be all the ops then hahaha

raikonenfnu: I agree with the rename, but I think we should keep default to false. The reasoning behind it being most ops we work with has no side effects and people adding ops are usually compiler author who should/would know to add this attribute when necessary. :)

PR #839, opened 2025-05-18
[Wave] Generalize Barriers to work across iterations

raikonenfnu:
This PR has two main purpose.

First and most importantly, it teaches our compiler to only insert shared memory barriers when there is truly no barriers in between ops. This was assumed before but with the soon-coming PR of ping-pong where we re-order and want barriers to be precisely inserted, this PR help handles those case and to not accidentally insert too many barriers when not needed.

Secondly, and complimentarily, this PR teaches compiler to insert barrier between ops from different iterations. Most commonly, shared_read at time=t, and shared_write at time=t+1. This allows us to also remove the hardcoded insertion of barrier at the start of every Iterate subgraph. For pingpong, this will allow the PingPong pass to manually insert barriers for next iteration, and this pass can analyze properly to not introduce redundant barriers.

To achieve those two goals we set above, below are listed the changes in code and their motivation:

- Fix bugs in decompose_reduce_ops where subgraph of Conditional is not properly registered to subgraphs and parent_op. Changes can be seen in `tracing.py` and `decompose_reduce_ops.py`

- Fix bug in pipelined_reduction where minimize_alloc cannot handle pipelined_reduce because it did not have a proper parent_op. Previously we have a workaround in tracing.py which is now removed. Changes can be seen in loop_reconstruction.py and tracing.py.

- Generalize insertion of barrier at top of Iterate into something that checks if barrier is required between ops from different iterations. This is done by adding support in `is_barrier_between` for case where the src do not dominate dst. We handle this by doing a check from src to end of loop, and then from start of loop to dst. We also updated implementation of `is_barrier_between_same_graph` since it was not handling these cases above properly.

-  We also made changes into `barriers.py` to do one more round of `add_shared_memory_barriers`, after the first round to mimic the next step of iteration, and see if there are any inter-iteration dependencies. This naturally removes the hardcoded version of inserting barrier at top of loop.

- Many LIT tests are updated because of the two points above, instead of having lds_barrier at top of every iterate, we now have barrier inserted only right before it's actually used which is the first shared_write in a loop (protecting shared read of previous iter).

### DIFF
--- a/iree/turbine/kernel/_support/tracing.py
+++ b/iree/turbine/kernel/_support/tracing.py
@@ -145,14 +145,9 @@ class CapturedTrace:
         self.region_graph = region_graph
         self.root_graph = root_graph
         self.region_graph.subgraphs[root_graph].subgraphs = {}
-        # Allow graphs to access each other.
-        for name, graph in self.region_graph.subgraphs.items():
-            if not hasattr(graph, "subgraphs"):
-                graph.subgraphs = {}
-            # For each graph, add all other graphs as potential subgraphs.
-            for subname, subgraph in self.region_graph.subgraphs.items():
-                if subname != name:
-                    graph.subgraphs[subname] = subgraph
+        for name, subgraph in self.region_graph.subgraphs.items():
+            if name != root_graph:
+                self.region_graph.subgraphs[root_graph].subgraphs[name] = subgraph
 
     def get_subgraph(self, name: str) -> fx.Graph:
         return self.region_graph.subgraphs[name]
--- a/iree/turbine/kernel/_support/tracing.py
+++ b/iree/turbine/kernel/_support/tracing.py
@@ -158,8 +158,6 @@ class CapturedTrace:
         return self.region_graph.subgraphs[name]
 
     def add_subgraph(self, name: str, graph: fx.Graph):
-        for _, subgraph in self.region_graph.subgraphs.items():
-            subgraph.subgraphs[name] = graph
         self.region_graph.subgraphs[name] = graph
 
     def get_root_graph(self) -> fx.Graph:
--- a/iree/turbine/kernel/wave/barriers.py
+++ b/iree/turbine/kernel/wave/barriers.py
@@ -4,7 +4,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-from .utils.graph_utils import is_reduction_subgraph
+from .utils.graph_utils import is_reduction_subgraph, is_barrier_between
 from .._support.tracing import CapturedTrace
 from ..ops.wave_ops import get_custom, Read, SharedMemoryBarrier, Write, NestedRegionOp
 from ..lang.global_symbols import SHARED_ADDRESS_SPACE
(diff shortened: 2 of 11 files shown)


iree/turbine/kernel/wave/barriers.py
```diff
@@ -49,10 +52,9 @@ def add_shared_memory_barriers(
...
     # Synchronize before the write to shared memory to avoid stepping over
-    # reads in the previous iteration of a loop.
-    if is_reduction_subgraph(graph) and last_node:
-        # Insert barrier at start of block, if load from shared memory exist.
-        with graph.inserting_after(graph._root):
-            SharedMemoryBarrier().add_to_graph(graph)
+    # shared reads in the previous iteration of a loop.
+    if is_reduction_subgraph(graph) and last_node and not check_cycle:
+        # Add barriers between ops from different iterations in the same loop.
+        add_shared_memory_barriers(trace, graph, last_node, check_cycle=True)
```
harsh-nod, line 58: Why do we need check_cycle? Doesnt seem to be used anywhere above?

iree/turbine/kernel/_support/tracing.py
```diff
@@ -158,8 +158,6 @@ def get_subgraph(self, name: str) -> fx.Graph:
         return self.region_graph.subgraphs[name]
 
     def add_subgraph(self, name: str, graph: fx.Graph):
-        for _, subgraph in self.region_graph.subgraphs.items():
```
harsh-nod, line 161: So one question here, do we need lines 148-155 in __init__ above? If we need each graph to access all other subgraphs then I think we still need this, otherwise you can get rid of both?

iree/turbine/kernel/wave/utils/graph_utils.py
```diff
@@ -368,23 +369,35 @@ def is_barrier_between_same_graph(src: fx.Node, dst: fx.Node) -> bool:
...
-    prev_node, next_node = src.prev, dst.next
-    found_src, found_dst = prev_node == src, next_node == dst
-    while prev_node.prev.op != "root" and not found_src:
-        prev_node, found_src = prev_node.prev, prev_node == src
-    if not found_src:
-        return False
-    while next_node.next.op != "root" and not found_dst:
-        next_node, found_dst = next_node.next, next_node == dst
-    return found_dst
+    next_node = src.next
```
harsh-nod, line 372: Nice! I dont remember why the original code didn't do this.

harsh-nod (changes requested):
overall looks good! thanks! Just some comments on check_cycle and whether we need to keep the mapping of each graph have access to all the other graphs.

iree/turbine/kernel/_support/tracing.py
raikonenfnu: Will try remove and see if everything is still alive, looking at the commit that checked that in, we may not need that

iree/turbine/kernel/wave/barriers.py
raikonenfnu: discussed offline, it was to check for next iteration, renamed to make it more intuitive :)

iree/turbine/kernel/_support/tracing.py
raikonenfnu: removed, needed to standardize search for subgraphs by using `get_root_graph` :)

harsh-nod: nice!

harsh-nod (approved):
lgtm! thanks!

PR #843, opened 2025-05-19
[runtime] Rename dltensor capsule to avoid memory leak

zjgarvey:
Adds code similar to <https://github.com/iree-org/iree-turbine/blob/d71316265c73c2d0c8487e1d98ca69a94d7a107c/iree/turbine/kernel/wave/utils/run_utils.py#L104>

There seems to be an issue with memory not getting freed when the py capsule for a dlpack tensor gets renamed. 

I filed an issue in iree: <https://github.com/iree-org/iree/issues/20849>, but in the meantime, we need this fix in order to do basically anything for BOO without running out of memory.

### DIFF
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -510,6 +510,9 @@ def _device_export_torch_tensor_cpu(
 # CUDA and HIP import/export
 ################################################################################
 
+_dl_tensor_name = ctypes.create_string_buffer(b"dltensor")
+_set_capsule_name = ctypes.pythonapi.PyCapsule_SetName
+
 
 def _device_import_torch_tensor_cuda_hip(
     device: Device, t: torch.Tensor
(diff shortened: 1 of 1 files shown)


Hardcode84:
Just FYI, apparently, there are also other leaks when running wave using iree-runtime, which are not fixed by renaming capsule, but I didn't investigated further and don't have any more details.

raikonenfnu:
How is this different from https://github.com/iree-org/iree-turbine/commit/79f61f395c5454ee6f99bee7e081a72c48c3a432 ?

raikonenfnu:
NVM, I see, this is a different path from that commit

raikonenfnu (approved):

stellaraccident:
This looks quite wrong. Please revert. 

In the future, please do not land code to such a delicate part of the runtime without a review from me or @AWoloszyn. This is quite wrong in how this went in.

PR #844, opened 2025-05-19
[TKW] Improve wave runtime

Hardcode84:
There are multiple changes here:
* Previously calling to `wave_compile` multiple times with `wave_runtime=True` was broken as  `wave_compile`  were overwriting serialized gpu binary on disk. Fixed by moving binary reading into `WaveKernel` init from the invoke.
* Simplified `wave_runtime.cpp` and `invoke_with_wave_runtime` and removed all the caching logic from there as it moved to `WaveKernel`.
* Use `tempfile.TemporaryDirectory()` for the temp bin path so it will be unique for each process to reduce race conditions when Wave called from multiple processes (races are still possible, though, when copying binary into cache dir).
* HIP module lifetime is now bound to `WaveKernel` lifetime.
* `get_wave_runtime_dir()` is no more

### DIFF
--- a/iree/turbine/kernel/wave/runtime/runtime.cpp
+++ b/iree/turbine/kernel/wave/runtime/runtime.cpp
@@ -94,10 +94,13 @@ static int launch(const KernelLaunchInfo &info, const Int64Vector &tensors,
     return hipSuccess;
 }
 
-static void unload_binary(void *ptr)
+static void unload_binary(void *ptr) noexcept
 {
     auto module = reinterpret_cast<hipModule_t>(ptr);
-    HIP_CHECK_EXC(hipModuleUnload(module));
+
+    // Assume it will never fail as we cant do anything meaningful about it
+    // anyway.
+    (void)hipModuleUnload(module);
 }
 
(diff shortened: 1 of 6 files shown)


iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -101,26 +86,40 @@ int launch(const KernelLaunchInfo &info, const Int64Vector &tensors,
...
     return hipSuccess;
 }
 
+static void unload_binary(void *ptr) noexcept
+{
+    auto module = reinterpret_cast<hipModule_t>(ptr);
+
+    // Assume it will never fail as we cant do anything meaningful about it
+    // anyway.
+    (void)hipModuleUnload(module);
```
nithinsubbiah, line 103: Can we add a break or even a print saying that the `free` was not successful if that's the case?

Hardcode84: done

nithinsubbiah (approved):

PR #850, opened 2025-05-21
[BOO] Enable `torch.compile` for models with boo convs

zjgarvey:
This change uses a `torch.library.custom_op` to define `boo_conv` instead of `autograd.Function`, which enables `torch.compile` for models with boo convolution. In addition, this commit factors out the autocasting logic since the decorators don't behave very well with respect to `custom_op`. It also changes the behavior slightly to mimic the functionality of `torch.convolution` under autocast contexts. 

These changes are a temporary solution, since it will be beneficial to take a more systematic approach of implementing autograd custom ops. I think this will make the most sense via implementing `iree.turbine.runtime.CustomOp` autograd support. Ideally, boo convolutions can simply be an application of `CustomOp` eager execution, in such a way that this is compatible with backward, `torch.compile`, and `torch.amp.autocast` contexts.

### DIFF
--- a/iree/turbine/kernel/boo/ops/conv.py
+++ b/iree/turbine/kernel/boo/ops/conv.py
@@ -4,106 +4,236 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+from typing import Sequence, Tuple
+
 import torch
 
-from ..conv_exports import ConvSignature, get_launchable
+from ..conv_exports import ConvSignature, get_launchable, DEFAULT_LAYOUTS
 
 __all__ = [
     "boo_conv",
 ]
 
 
-class _BooConv(torch.autograd.Function):
-    @staticmethod
-    @torch.amp.custom_fwd(device_type="cuda")
-    def forward(ctx, x, w, b=None, kwargs={}):
-        x = x.detach()
-        w = w.detach()
-        device_type = x.device.type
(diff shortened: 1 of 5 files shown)


iree/turbine/kernel/boo/ops/conv.py
```diff
@@ -127,5 +255,35 @@ def boo_conv(
...
+    num_spatial_dims = len(weight.shape) - 2
+
+    def listify(value) -> Sequence[int]:
+        if isinstance(value, Sequence):
+            return value
+        if isinstance(value, int):
+            return [value] * num_spatial_dims
+        return list(value)
+
+    _infer = lambda layout: shared_layout or layout or DEFAULT_LAYOUTS[num_spatial_dims]
```
rkayaith, line 267: should `layout` take precedence over `shared_layout`?

rkayaith (approved):
I don't fully follow the torch integration details, but from a high level this looks good!

iree/turbine/kernel/boo/ops/conv.py
zjgarvey: Hmmm. I did this to be consistent with an earlier choice, but I think you are right that it would be more convenient to be able to specify a shared layout, and also specify one layout which differs. 

I think I'll make that change in a different patch though, if you are okay with that.

rkayaith: yup thats fine :+1:

PR #851, opened 2025-05-22
[TKW] Add patterns to sink vector.extract/splat ops

Hardcode84:


iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -1332,9 +1329,21 @@ def handle_reshape(emitter: WaveEmitter, node: fx.Node):
 
     # Determine whether to extract or combine.
     if len(args) > 1:
+        vectors = [cast_vector(emitter, arg) for arg in args]
+        shape = vectors[0].type.shape[0]
+        if shape == 1:
```
raikonenfnu, line 1334: Can we add a small doc on why this is special case

Hardcode84: done

PR #853, opened 2025-05-22
Pin IREE version to 3.5.0rc20250516

Hardcode84:
We have a bad regression in Wave between `3.5.0rc20250516` and `3.5.0rc20250517`. 40% perf drop in paged extend attention kernel and crash in paged decode (which is not reproducible in our unit tests).

I'm locking to the last known good version until further investigation.

### DIFF
--- a/requirements-iree-pinned.txt
+++ b/requirements-iree-pinned.txt
@@ -7,5 +7,7 @@
 # Uncomment to skip versions from PyPI (so _only_ nightly versions).
 # --no-index
 
-iree-base-compiler==3.5.0rc20250522
-iree-base-runtime==3.5.0rc20250522
+# TODO: We have a pretty bad regression in Wave between 3.5.0rc20250516 and
+# 3.5.0rc20250517. I'm locking to 3.5.0rc20250516 until further investigation.
+iree-base-compiler==3.5.0rc20250516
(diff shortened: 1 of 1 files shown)


raikonenfnu (approved):

requirements-iree-pinned.txt
```diff
@@ -7,5 +7,7 @@
 # Uncomment to skip versions from PyPI (so _only_ nightly versions).
 # --no-index
 
-iree-base-compiler==3.5.0rc20250522
-iree-base-runtime==3.5.0rc20250522
+# TODO: We have a pretty bad regression in Wave between 3.5.0rc20250516 and
+# 3.5.0rc20250517. I'm locking to 3.5.0rc20250516 until further investigation.
```
ScottTodd, line 11: Oof. Please follow up with increases in test coverage once you learn more.

pashu123:
This could be the suspect: https://github.com/iree-org/iree/pull/20895

Hardcode84:
Crash was caused by IREE dymanic shared mem change and related runtime ABI changes, we will need to update `Wave_runtime` to handle it, test https://github.com/iree-org/iree-turbine/pull/855 

Slowdown on the other hand is due to some changes in AMDGPU LLVM backed pulled by IREE llvm integrate.

PR #857, opened 2025-05-23
[TKW](WIP) Switch Wave iree runtime path to `turbine.runtime`

Hardcode84:
An experiment to switch Wave to `turbine.runtime` instead of half-baked homemade iree wrappers. Surprisingly, mostly worked out of the box. Didn't test async path yet. Didn't test call overhead yet either, but amount of python code doesn't give much optimism.

### DIFF
--- a/iree/turbine/runtime/launch.py
+++ b/iree/turbine/runtime/launch.py
@@ -199,7 +199,9 @@ class Launchable:
             f"Could not load a target binary for device {turbine_device}"
         )
 
-    def __call__(self, *args, device: Optional[torch.device] = None):
+    def __call__(
+        self, *args, device: Optional[torch.device] = None, outputs: Sequence[Any] = ()
+    ):
         turbine_device: Optional[Device] = (
             None if device is None else get_device_from_torch(device)
         )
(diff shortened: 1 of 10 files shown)


Hardcode84:
Will split into multiple PRs, `iree_ref` can go to the `turbine.runtime` unconditionally as we doesn't actually care for the ref code overhead.

iree/turbine/runtime/launch.py
```diff
@@ -238,7 +240,13 @@ def __call__(self, *args, device: Optional[torch.device] = None):
...
         vm_context, vm_function = self._resolve_target_binary(turbine_device)
 
-        ret_list = VmVariantList(1)
+        ret_list = VmVariantList(len(outputs))
+        for output in outputs:
+            if isinstance(output, Tensor):
+                assert output.is_contiguous(), "Outputs must be contiguous"
+                ret_list.push_ref(turbine_device.import_torch_tensor(arg))
+            else:
+                raise ValueError(f"Unsupported output type: {type(output)}")
```
zjgarvey, line 249: When we do this, I'm pretty sure IREE will still make internal allocations for the output tensors and then copy them over to these pytorch-initialized buffers. 

I'm not sure if this is desirable since the DtoD copy will take some time. Might be good to double check whether this is happening or not, and if so, just use inplace invocation when that is desirable.

PR #861, opened 2025-05-27
Use `turbine.runtime` for `generate_iree_ref`

Hardcode84:
`generate_iree_ref` only used to get reference values for Wave tests and it also the last user of `inplace=False` flag in Wave.


stellaraccident (approved):

stellaraccident:
@AWoloszyn another case like the other we were looking at which needs to be fast pathed.

Hardcode84:
Just to be clear, I don't care about perf much for the `generate_iree_ref` as it is only used to get ref values for Wave tests but I want to use it for the actual Wave kernels too and I do care about launch overhead there.

PR #840, opened 2025-05-19
[Wave] Make APLP faster

harsh-nod:
This PR adds an implementation of APLP
for software pipelining in Rust to speed up
the scheduling computation.

PR #868, opened 2025-05-28
added cos op

bodhisaha:
needed to pip install dependencies in forked repo on machine.


iree/turbine/kernel/wave/scheduling/aplp/target/maturin/libaplp_lib.so
ScottTodd, line 1: At a minimum, these files should be excluded via a .gitignore.

Pre-reviewing your own changes before pushing to github and creating a pull request should also catch such issues.

I've also commented on https://github.com/iree-org/iree-turbine/pull/840 about the new Rust dependency.

PR #878, opened 2025-05-30
[WAVE] Correcting predicts dim in the first speculative decode kernel 

xintin:
1. Fixed `predicts` dim in the first kernel. 
2. updated a few variable names.


nithinsubbiah (approved):
Please change the `GLOBAL_ADDRESS_SPACE` variable name to avoid name conflict. Looks good otherwise, nice find

PR #880, opened 2025-05-31
Update IREE deps

Hardcode84:
IREE deps were previously pinned to `3.5.0rc20250516` as `3.5.0rc20250517` introduced regressions. There were 2 issues:
* Shared memory handling changes in IREE were causing the crash as we are were trying to allocate too much shared mem. IREE doesn't use dynamic shared mem anymore so we need to pass 0 now in wave runtime (test https://github.com/iree-org/iree-turbine/pull/855).
* 40% perf regression in extend attention caused by some changes in llvm AMDGPU backend. These seems to went away after the recent IREE llvm intergrate. I didn't dig too deep on them but I suspect https://github.com/llvm/llvm-project/pull/137807 was the issue (fixed by https://github.com/llvm/llvm-project/pull/140921).

### DIFF
--- a/iree/turbine/kernel/wave/runtime/runtime.cpp
+++ b/iree/turbine/kernel/wave/runtime/runtime.cpp
@@ -88,7 +88,7 @@ static int launch(const KernelLaunchInfo &info, const Int64Vector &tensors,
         HIP_LAUNCH_PARAM_END};
 
     HIP_CHECK_RETURN(hipModuleLaunchKernel(function, info.gridX, info.gridY, info.gridZ,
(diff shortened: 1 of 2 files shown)


ScottTodd (approved):
Thanks! You'll want to close https://github.com/iree-org/iree-turbine/pull/856 and delete the `integrates/iree` branch after merging this so the https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/update_iree_requirement_pins.yml action starts running again 

See for example the latest run: https://github.com/iree-org/iree-turbine/actions/runs/15390641246 
image

PR #883, opened 2025-06-02
Allow stream sharing between iree and iree-turbine.

AWoloszyn:
This lets us use the new iree external hip stream functionality, which allows us to have less latency between launches in pytorch and iree.


AWoloszyn:
> Can you fix the WIP commit to be signed off? Or rebase all of this and make sure it passes DCO?

Done

zjgarvey (commented):
Some questions:

1. Do we want `DeviceState.from_uri` to also use stream sharing? Currently it doesn't, but we aren't using this path with either `Launchable` or `CustomOp`, so I didn't think to add this functionality in.
2. Are the tests here sufficient in your opinion? Should we add any other runtime interop tests to make sure things work properly with stream sharing (e.g., interleaving with actual torch ops)?

zjgarvey (approved):
I'm putting in an approval, since the questions I asked about could likely be addressed later.

I'll open up an issue to track the `DeviceState.from_uri` uses in our stack and see if any of that code needs to be iterated on. 

The test coverage for async interop should be made robust and I'm not sure how to do this. I'll make an issue for getting better test coverage and assign you, @AWoloszyn. I don't want to have breaking changes like #843 land again in the future (from me or others).

PR #885, opened 2025-06-03
[BOO] Add Conv1d and Conv3d replacement capability

zjgarvey:


iree/turbine/kernel/boo/modeling/replace.py
```diff
@@ -80,6 +148,156 @@ def forward(self, x):
         return result.squeeze(0) if no_batch else result
 
 
+class BooConv3d(torch.nn.Module):
+    def __init__(
+        self,
+        in_channels,
+        out_channels,
```
rkayaith, line 155: conv3d got left out of the type hinting

zjgarvey:
Actually, I need a second to work out the fact that `torch.channels_last` doesn't apply to `1d`/`3d` cases...

zjgarvey:
Okay, @rkayaith , I think this is good for a re-review. 

It looks like there isn't a 1d channels last memory format, so I'm just going to always do a pytorch layout conv in the 1d case (not something we needed anyway, but the 3d channels_last we do need).

PR #890, opened 2025-06-03
[Wave] Speculative decode - default predict token to last token in vocab

nithinsubbiah:
- Fixes `uniform_sampling` dimensions
- Default predict token to n-1 from vocabulary

### DIFF
--- a/iree/turbine/kernel/wave/templates/speculative_decoding.py
+++ b/iree/turbine/kernel/wave/templates/speculative_decoding.py
@@ -124,6 +124,8 @@ def get_speculative_decoding_kernel(
             coin * sum_relu, target_shape=[BATCH_SIZE, NUM_DRAFT_TOKENS, VOCAB_SIZE]
         )
         greater_than_u = cdf > threshold_dist_u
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/wave/templates/speculative_decoding.py
```diff
@@ -124,14 +124,16 @@ def tree_speculative_sampling(
             coin * sum_relu, target_shape=[BATCH_SIZE, NUM_DRAFT_TOKENS, VOCAB_SIZE]
         )
         greater_than_u = cdf > threshold_dist_u
-        pad_token = tkl.Register[BATCH_SIZE, NUM_DRAFT_TOKENS, VOCAB_SIZE, tkl.i32](1e6)
+        pad_token = tkl.Register[BATCH_SIZE, NUM_DRAFT_TOKENS, VOCAB_SIZE, tkl.i32](
```
xintin, line 127: nice find. can we add a comment why it should be `VOCAB_SOZE - 1`?

nithinsubbiah: I've added it right below

PR #896, opened 2025-06-04
Tweak schedule

harsh-nod:
This PR adds code to tweak schedules programatically.
Changes to schedules will attempt to preserve
dependencies in the graph as well as resource
constraints.


xintin:
Awesome! LGTM.

iree/turbine/kernel/wave/scheduling/verifier.py
```diff
@@ -0,0 +1,432 @@
...
+                    adj_map[edge._to].append(edge._from)
+        return adj_map
+
+    def get_successors(self, node: fx.Node) -> List[fx.Node]:
+        return self._adj.get(node, [])
+
+    def get_predecessors(self, node: fx.Node) -> List[fx.Node]:
+        return self._pred_adj.get(node, [])
+
+    def has_edge(self, pred: fx.Node, succ: fx.Node) -> bool:
```
xintin, line 104: I think, here we are doing linear search every time. 
Will it be useful in the future PR to convert it into a set, if edges are of higher order?

iree/turbine/kernel/wave/scheduling/verifier.py
```diff
@@ -0,0 +1,432 @@
...
+import numpy as np
+import torch
+import torch.fx as fx
+from .resources import get_available_resources
+from .graph_utils import Edge
+from ...ops.wave_ops import get_custom, IterArg
+
+Schedule = Dict[fx.Node, int]
+RawEdgesList = Optional[List[Edge]]
+NodeRRTGetter = Callable[[fx.Node], np.ndarray]
```
Hardcode84, line 11:
```
Schedule: TypeAlias = Dict[fx.Node, int]
RawEdgesList: TypeAlias = Optional[List[Edge]]
NodeRRTGetter: TypeAlias = Callable[[fx.Node], np.ndarray]
```

iree/turbine/kernel/wave/scheduling/verifier.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def get_successors(self, node: fx.Node) -> List[fx.Node]:
+        return self._adj.get(node, [])
+
+    def get_predecessors(self, node: fx.Node) -> List[fx.Node]:
+        return self._pred_adj.get(node, [])
+
+    def has_edge(self, pred: fx.Node, succ: fx.Node) -> bool:
+        return (
+            any(edge._from == pred and edge._to == succ for edge in self.edges)
```
Hardcode84, line 106: nit: `self.edges and any(edge._from == pred and edge._to == succ for edge in self.edges)`

iree/turbine/kernel/wave/scheduling/verifier.py
```diff
@@ -0,0 +1,432 @@
...
+                # Forward repair: enforce predecessor constraints
+                if pred_cycles and repaired_schedule[node] <= max(pred_cycles):
+                    return True, max(pred_cycles) + 1
+            else:
+                # Backward repair: enforce successor constraints
+                if succ_cycles and repaired_schedule[node] >= min(succ_cycles):
+                    return True, min(succ_cycles) - 1
+            return False, 0
+
+        for _ in range(self.MAX_REPAIR_ITERATIONS):
```
Hardcode84, line 189: If we exceed `MAX_REPAIR_ITERATIONS` this code will fallthrough and return `True` even if schedule wasn't fully repaired.

iree/turbine/kernel/wave/scheduling/verifier.py
```diff
@@ -0,0 +1,432 @@
...
+            node_rrt_getter: Function to get resource requirements for nodes
+            forward: If True, try cycles forward; if False, try cycles backward
+
+        Returns:
+            bool: True if the move was successful, False otherwise
+        """
+        original_cycle = schedule[node]
+        resource_tracker.remove_node(node, original_cycle, node_rrt_getter)
+
+        if forward:
```
Hardcode84, line 263: `range(...) if forward else range(...)` instead of copypasting entire loop

iree/turbine/kernel/wave/scheduling/verifier.py
harsh-nod: Yes, good point this can be converted into a set to make it faster. I will change it.

iree/turbine/kernel/wave/scheduling/verifier.py
harsh-nod: Good catch. I have modified the code so that if we exceed MAX_REPAIR_ITERATIONS, we only return True if the final schedule satisfies the dependency constraints.

PR #900, opened 2025-06-05
[BOO] Reduce CPU overhead

zjgarvey:
This PR. is meant to cut out some CPU overhead from boo convolution launches. 

With these changes, we are pretty close to removing most of the non-essential fluff.

### Changes

1. Any `nn.Module`-level information is pre-processed during module initialization (and not during runtime).
2. `boo_conv` is left as-is for convenience use, but the `BooConvNd` modules will now directly invoke the function `boo_convolution`, which essentially is the same as `boo_conv` but without arg pre-processing.
3. The move to using `boo_convolution` meant we had to duplicate some `amp.autocast` context management into the module forward calls. This could be something to look at improving.
4. `boo_convolution` checks if grads are required and applies either a raw forward function or an `autograd.Function.apply`.
5. Converts all `torch.library.custom_op` decorations to direct library registrations. This is by far the biggest time save. Note: we needed to register these launchable calls as custom ops otherwise they wouldn't work with `torch.compile`. 
6. Directly computes `func_name` (though `get_func_name` itself is cached) and queries the launchable cache without getting a `ConvSignature` first.

### Non-Performance Changes

1. Improves profiling in the `resnet_18_backward.py` example.

### Some Time Comparisons (without this PR -> with this PR)

```
nn.Module: BooConv2d_0
1269.214us -> 713.932us
nn.Module: BooConv2d_1
585.749us -> 291.247us
nn.Module: BooConv2d_2
614.772us -> 269.555us
nn.Module: BooConv2d_3
522.955us -> 257.086us
nn.Module: BooConv2d_4
588.232us -> 352.178us
nn.Module: BooConv2d_5
531.868us -> 362.764us
nn.Module: BooConv2d_6
490.025us -> 248.613us
nn.Module: BooConv2d_7
496.745us -> 273.63us
nn.Module: BooConv2d_8
531.867us -> 303.576us
nn.Module: BooConv2d_9
581.793us -> 303.566us
nn.Module: BooConv2d_10
553.761us -> 363.676us
nn.Module: BooConv2d_11
589.795us -> 325.017us
nn.Module: BooConv2d_12
496.074us -> 243.215us
```

### Remaining Overhead

1. `amp.autocast` is *always* doing a down-cast of the weight tensor on forward calls. This is the same with/without BOO, but we might be able to just bring this into IREE. 
2. `torch.channels_last` handing: We currently conjugate `boo_convolution` with permutations to contiguous format. This makes sense if we are doing backward boo convs, but since we are doing pytorch backward convs right now, this means we need to permute the context-saved input and weight tensors back to `torch.channels_last`, pass to the pytorch backward conv, and then permute the results of the backward conv (input_grad, weight_grad) to contiguous to match the expected output format and shape of `boo_convolution`. To be clear, all of this is metadata manipulation on the CPU; however, these microseconds are adding up. Note: with `torch.compile`, these seem to be less of a problem. 
3. `unwrap_dead_wrappers` is a solid 7-8 us. This gets optimized out with `torch.compile`.
4. Anything that occurs inside the `Launchable.__call__`. Important culprits: dlpack handoff, `torch.cuda.current_stream`.

### Future TODOs:

1. quick cache retrieval for backward boo convs. 
2. Instead of direct library definition, use `iree.turbine.runtime.op_reg.CustomOp` and enable backward. I tried this out in another branch and it works just as-well, but I need more time to clean it up, and would like to work that out in a subsequent patch. We could continue to define future op fusions by assembling the graph with a custom op and other torch code.

### DIFF
--- a/iree/turbine/kernel/wave/runtime/runtime.cpp
+++ b/iree/turbine/kernel/wave/runtime/runtime.cpp
@@ -88,7 +88,7 @@ static int launch(const KernelLaunchInfo &info, const Int64Vector &tensors,
         HIP_LAUNCH_PARAM_END};
 
     HIP_CHECK_RETURN(hipModuleLaunchKernel(function, info.gridX, info.gridY, info.gridZ,
-                                           info.blockX, info.blockY, info.blockZ, 0,
+                                           info.blockX, info.blockY, info.blockZ, info.sharedMemoryBytes,
                                            stream, nullptr, (void **)&hipLaunchParams));
 
     return hipSuccess;
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -613,6 +613,8 @@ def lookup_device_from_torch(
     torch_device: torch.device, *, create: bool = True
 ) -> Optional[Device]:
     """Gets a shared Device corresponding to the given torch.device.
+    For cuda torch.device types, the returned Device will be tied to
+    the current cuda stream.
 
     This will return None if the device is wholly unsupported or if
     create=False. Otherwise, faults in setting up the device are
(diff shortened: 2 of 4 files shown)


zjgarvey:
I've gotten the `get_device_from_torch` CPU time down from 15us to 5us with some modifications (i.e. directly getting the stream pointer instead of getting a `torch.cuda.Stream` then grabbing a pointer), but I'd like to land this and do a standalone pr for any core runtime infra changes.

iree/turbine/kernel/boo/ops/conv.py
```diff
@@ -32,8 +45,72 @@ def disable_backward():
     BOO_USE_BACKWARD_KERNELS = 0
 
 
-@torch.library.custom_op("iree_turbine::boo_convolution", mutates_args=())
-def boo_convolution(
+# Utilities #
+
+
+def make_tuple(a: Any, size: int) -> Tuple:
```
rkayaith, line 51: `a` could be type hinted as `Iterable[int] | int`

PR #902, opened 2025-06-05
[WAVE] Update ci tk yaml to have updated rocm installation 

xintin:


Hardcode84 (approved):
Non-self-hosted github runners seems to have unrelated incident but this change seems to fixed our runners. LGTM

Issue #895, opened 2025-06-04
Tests outside TK/WAVE not running on GPU


zjgarvey:
Currently, none of the core infra tests in the CI are running on GPU. 

All device-related tests, BOO, etc. are currently untested for GPU in the CI.

I'd like to have the following:

1. Setup a Unit Tests and Type Checking CI job that installs from `pytorch-rocm-requirements.txt` on an Mi300 machine.
2. Add more robust GPU-specific interop tests (`pytorch <---> iree` memory interop, async interop etc.) to avoid catastrophic modifications to runtime code.

PR #903, opened 2025-06-05
Add a python3.11 mi300 unit tests and type checking job

zjgarvey:
Partly addresses <https://github.com/iree-org/iree-turbine/issues/895>, but we should still add some more iree/pytorch interop tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,20 +23,22 @@ concurrency:
 
 jobs:
   test:
-    name: "${{ matrix.os }} :: ${{ matrix.version }} :: Unit Tests and Type Checking"
+    name: "${{ matrix.name }} :: ${{ matrix.version }} :: Unit Tests and Type Checking"
     strategy:
       fail-fast: false
       matrix:
         # Support for Python 3.13 depends on https://github.com/pytorch/pytorch/issues/130249
+        name: [ubuntu-24.04]
         version: ["3.10", "3.11", "3.12"]
-        os: [ubuntu-24.04]
+        runs-on: [ubuntu-24.04]
         pytorch_requirements: ["pytorch-cpu-requirements.txt"]
 
         include:
-          - version: "3.11"
-            os: linux-mi300-1gpu-ossci-iree-org
+          - name: linux-mi300
+            version: "3.11"
+            runs-on: linux-mi300-1gpu-ossci-iree-org
             pytorch_requirements: "pytorch-rocm-requirements.txt"
-    runs-on: ${{matrix.os}}
+    runs-on: ${{matrix.runs-on}}
     env:
       VENV_DIR: ${{ github.workspace }}/.turbine-venv
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -55,7 +61,7 @@ jobs:
           # Note: We install in three steps in order to satisfy requirements
           # from non default locations first. Installing the PyTorch CPU
           # wheels saves multiple minutes and a lot of bandwidth on runner setup.
-          pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -r ${{ matrix.pytorch_requirements }}
```
ScottTodd, line 64: This comment about "Installing the PyTorch CPU wheels" is now outdated. The ordering here interacts with this code: https://github.com/iree-org/iree-turbine/blob/26ff1d28d1e7db58778159a99a435fb22758795a/requirements.txt#L18-L22

In the case of AMDGPU, we want that first so the default pytorch install doesn't pull in the default of CUDA (since we want ROCm instead). On CPU, we want the CPU requirements first because the default of CUDA would work but is slow.

---

I also want to keep it clear via docs and ideally enforced in code that:

* Export from PyTorch to MLIR has no GPU requirement - NVIDIA, AMD, or otherwise
* After export, IREE can run on GPU with no PyTorch requirement
* Interop between PyTorch on GPU and Turbine/Wave/IREE on GPU will need a compatible PyTorch install

.github/workflows/ci.yaml
```diff
@@ -30,6 +30,12 @@ jobs:
         # Support for Python 3.13 depends on https://github.com/pytorch/pytorch/issues/130249
         version: ["3.10", "3.11", "3.12"]
         os: [ubuntu-24.04]
+        pytorch_requirements: ["pytorch-cpu-requirements.txt"]
+
+        include:
+          - version: "3.11"
+            os: linux-mi300-1gpu-ossci-iree-org
```
ScottTodd, line 37: This isn't an operating system, so we could rename to "runs-on", similar to here: https://github.com/iree-org/iree/blob/e43c555623122ad3be55673e8caa042dd1505c0c/.github/workflows/pkgci_test_onnx.yml#L28-L38

We could also set a matrix name that is more condensed in the UI: https://github.com/iree-org/iree-turbine/actions/runs/15475532636/job/43569829445?pr=903 (note how the full name gets truncated on the left)
image

See https://github.com/iree-org/iree/blob/e43c555623122ad3be55673e8caa042dd1505c0c/.github/workflows/pkgci_test_onnx.yml#L22 up a bit in that file for how to set a name

.github/workflows/ci.yaml
ScottTodd, line 1: FYI from looking at the logs, there is a lot of stdout spam: https://github.com/iree-org/iree-turbine/actions/runs/15475532636/job/43569829445?pr=903#step:6:3460 . Some tests printing full MLIR modules even on success?

That happens for the CPU test jobs too, so preexisting but would be good to fix.

ScottTodd (approved):

.github/workflows/ci.yaml
zjgarvey: I'll open an issue to address this later. I agree it's a bit too verbose.

zjgarvey:
Thanks @ScottTodd for the comments and review!

ScottTodd:
Oh, I overlooked something here.

We might want a separate job or even workflow file for GPU tests on self-hosted runners. Our runners are more stable now than they have been in the past, but https://iree.dev/developers/general/github-actions/#using-github-hosted-and-self-hosted-runners can still apply:

> Due to how difficult it can be to keep self-hosted runners operating reliably, and how access is limited to the group that maintains the hardware, any jobs using these self-hosted runners must be optional and easy to disable.

What I want to avoid is having a pull request be blocked on merging due to a queue for self-hosted GPU runners, or those runners having some issue that is outside of project contributors control.

PR #904, opened 2025-06-06
[WAVE] Merge two kernels into one in speculative decode

xintin:
- merged two kernels into one kernel
- updated lit test

### DIFF
--- a/iree/turbine/kernel/wave/templates/speculative_decoding.py
+++ b/iree/turbine/kernel/wave/templates/speculative_decoding.py
@@ -10,139 +10,6 @@ from iree.turbine.kernel.lang.global_symbols import *
 import sympy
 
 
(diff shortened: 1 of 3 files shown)


iree/turbine/kernel/wave/templates/speculative_decoding.py
```diff
@@ -199,16 +68,24 @@ def get_speculative_sampling_kernel(
     constraints += [tkw.TilingConstraint(CUR_INDEX)]
     constraints += [tkw.TilingConstraint(J)]
 
+    constraints += [
```
nithinsubbiah, line 71: nit: reorder similar constraints together

nithinsubbiah (changes requested):

PR #905, opened 2025-06-06
Reduce CPU overhead in `lookup_device_from_torch`

zjgarvey:
Calling `torch.cuda.current_stream(torch_device)` was taking between 15 and 50 us on the CPU for every launchable call. 

This is because it checks whether device is an `int`, or a `torch.device`, or a `str`, etc., lazy initializing cuda for no reason, and creating a `torch.cuda.Stream` just so we could get the raw stream pointer int.

This PR also fixes some tests for Device that were not running due to duplicate class name, and adds a few more checks on the GPU device tests.

### DIFF
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -613,6 +613,8 @@ def lookup_device_from_torch(
     torch_device: torch.device, *, create: bool = True
 ) -> Optional[Device]:
     """Gets a shared Device corresponding to the given torch.device.
+    For cuda torch.device types, the returned Device will be tied to
+    the current cuda stream.
 
     This will return None if the device is wholly unsupported or if
     create=False. Otherwise, faults in setting up the device are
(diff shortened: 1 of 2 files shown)


zjgarvey:
Actually, `torch.cuda._is_in_bad_fork()` only takes around 500 ns, so maybe we just add an `assert not torch.cuda._is_in_bad_fork()`.

AWoloszyn:
> I'm not sure how necessary it is to check `torch.cuda.is_initialized()`. The raw stream pointer function gives the correct result even if `torch.cuda.is_initialized()` is False, but maybe there is something about checking edge cases where `torch._C._cuda_isInBadFork()` that I don't understand. The `torch.cuda._lazy_init` call takes less than 5us, so if this is strictly necessary, we can add it in.

I suspect this is a quirk of cuda(hip) in that the default stream is 0. I assume that before torch.cuda.is_initialized() if we get a 0 stream, if we then lazy initialize, we ALSO have a zero stream, as no-one has updated the stream yet.

AWoloszyn (approved):
This looks good to me. I can't think of any edge-cases that we would ever hit,  but maybe we do want to add the assert that we are not in the bad fork.

PR #906, opened 2025-06-07
[Wave] Add support for WG/Wave tiles smaller than vector sizes

Hardcode84:
Handle `tile_size < vector_size` case by doing masking on shared mem. For the case when we have `BLOCK_M x BLOCK_N` tile size and `VEC_M x VEC_N` vector shape, it will read `BLOCK_M x BLOCK_N` tile from global to shared mem and then read   `VEC_M x VEC_N` from shared mem, masking values outside of `BLOCK_M x BLOCK_N`  to zeros.

* Refactor `DistributionConstraint` to have `get_index_bound` which will return combined global bound and tile bound and use it to determine bounds.
* Remove old hacky `align_index_sizes` pass, handle shared mem logic directly in `_build_mask` during MLIR lowering.
* Fix `sympy.Min/Max` lowering when the have more then 2 args.

### DIFF
--- a/iree/turbine/kernel/wave/codegen/read_write.py
+++ b/iree/turbine/kernel/wave/codegen/read_write.py
@@ -147,9 +147,13 @@ def _build_mask(
     if is_shared_mem and bounds:
         bounds = remove_global_indexing(bounds, emitter.constraints)
         # Masking against global bounds was already handled when reading from
-        # global mem, but we may still need to handle masking against vector size during shared mem access.
-        # Bound expression for this case will look like `min(global_bound, vector_size)`.
-        # Replace global bound with some known big enough value so it can be simplified to just vector size.
+        # global mem, but we may still need to handle masking against vector
+        # size during shared mem access.
+        # Bound expression for this case will look like
+        # `min(global_bound, vector_size)`.
+        # Replace global bound with `max(tile_size, vector_size)` so the entire
+        # expression `min(max(tile_size, vector_size), vector_size)` can be
+        # simplified to just vector size.
         bounds = {
             k: safe_subs(
                 v, {k: _get_max_tile_size(k, emitter.constraints, vector_shapes)}
--- a/iree/turbine/kernel/wave/codegen/read_write.py
+++ b/iree/turbine/kernel/wave/codegen/read_write.py
@@ -37,6 +37,7 @@ from ...compiler.vector_codegen import (
     cast_vector,
 )
 
+from ..constraints import Constraint, DistributionConstraint
 from ...ops.wave_ops import get_custom, read, write, CustomOp
 
 from ..utils.general_utils import (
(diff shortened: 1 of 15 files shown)


iree/turbine/kernel/wave/constraints.py
```diff
@@ -501,6 +521,25 @@ def apply(self) -> IndexSequence:
     def work_bound(self) -> IndexExpr:
         return self.count * self.tile_size
 
+    @property
+    def dim_bound(self) -> IndexExpr:
+        return self.dim
+
+    def get_index_bound(self, vector_shape: Optional[int]) -> Optional[IndexExpr]:
+        bound = None
+        if subs_idxc(self.work_bound) != subs_idxc(self.dim_bound):
```
Hardcode84, line 530: add divisibility comment

iree/turbine/kernel/wave/expansion/expansion_utils.py
```diff
@@ -87,12 +91,12 @@ def get_dim_scaling(
                 tile_size % wave_count != 0
                 or (tile_size / wave_count) % vector_size != 0
             ):
-                raise ValueError(
```
raikonenfnu, line 90: Can we double check that 
```
                tile_size % wave_count != 0
                or (tile_size / wave_count) % vector_size != 0

```
Indeed work for all cases before removing completely, or perhaps need to fine tune this to error out in untested cases

iree/turbine/kernel/wave/codegen/read_write.py
```diff
@@ -120,11 +123,48 @@ def _get_symbolic_shape(node: fx.Node) -> tuple[IndexExpr]:
...
-    emitter: WaveEmitter, index: Dict[IndexExpr, IndexExpr], elements_per_thread: int
+    emitter: WaveEmitter,
+    index: dict[IndexExpr, IndexExpr],
+    elements_per_thread: int,
+    vector_shapes: dict[IndexSymbol, int],
+    is_shared_mem: bool,
 ) -> Optional[OpResult]:
-    bounds = find_index_bounds(emitter.constraints, index)
-    if bounds is None:
+    bounds = find_index_bounds(emitter.constraints, index, vector_shapes)
```
raikonenfnu, line 146: If we can have some numeric example, that would be nice!

iree/turbine/kernel/wave/utils/general_utils.py
```diff
@@ -126,9 +126,13 @@ def remove_global_indexing(
...
         for constraint in tiling_constraints:
             new_dim = new_index[key]
-            if sympy.sympify(new_dim.start).has(constraint.induction_var):
+            if new_dim.has(constraint.induction_var):
                 new_dim = new_dim.subs({constraint.induction_var: 0})
-                new_dim.start = new_dim.start - constraint.start
+                if isinstance(new_dim, IndexSequence):
+                    new_dim.start = new_dim.start - constraint.start
+                else:
+                    new_dim = new_dim - constraint.start
```
raikonenfnu, line 134: this probably need another isinstance protection, and then we can have an else that raise ValueError

PR #913, opened 2025-06-10
[Wave] Add aplp to gitignore

tgymnich:
ignore all cargo related files


Hardcode84:
wait, no, it's not actually correct as it contains actual sources

PR #922, opened 2025-06-12
[Wave] Add support for direct global load to lds

nithinsubbiah:


iree/turbine/kernel/wave/utils/graph_utils.py
```diff
@@ -438,3 +439,23 @@ def is_barrier_between(src: fx.Node, dst: fx.Node) -> bool:
             return dst_check or root_check
 
         assert False, "Unhandled case when src and dst are in different graphs"
+
+
+def has_write_shared_user(node: Read) -> bool:
```
raikonenfnu, line 444: can you move this closer to https://github.com/iree-org/iree-turbine/blob/71e89616c133dc40cdd50b8a70a1572faa50412c/iree/turbine/kernel/wave/utils/general_utils.py#L364-L376

PR #923, opened 2025-06-12
Add stack tracing capability to Wave

ftynse:
Extend the previously added location information capability to capture complete stack traces when constructing Wave operations. This is particularly useful for debugging kernels integrated into Pytorch or more complex setups with nested function calls or decorators.

Thread the configuration options through the RegionGraph object because it's the only one accessible everywhere during op construction. For non-Wave kernels, the configuration may remain empty.

The location capture configuration options are added in a separate file to avoid transitive import of MLIR components when only configuration is necessary.

### DIFF
--- a/iree/turbine/kernel/_support/tracing.py
+++ b/iree/turbine/kernel/_support/tracing.py
@@ -35,6 +35,7 @@ from ..lang.wave_types import IndexMapping
 from ..ops.wave_ops import CustomOp
 
 from .regions import RegionGraph, SubgraphTracer
+from .location_config import LocationCaptureConfig
 
 from .. import ops
 from ..ops.base import (
@@ -60,6 +61,13 @@ TCallable = TypeVar("TCallable", bound=Callable)
 
 
 class KernelRegionGraph(RegionGraph):
+    def __init__(
+        self, *, location_capture_config: Optional[LocationCaptureConfig] = None
+    ):
+        super(KernelRegionGraph, self).__init__(
+            location_capture_config=location_capture_config
+        )
+
     def new_subtracer(
         self,
         region_graph: "RegionGraph",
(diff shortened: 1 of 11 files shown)


iree/turbine/kernel/ops/wave_ops.py
```diff
@@ -449,7 +449,7 @@ class CustomOp(ABC):
     _tracing_function: Optional[Callable[..., Any]] = field(default=None, init=False)
 
     @property
-    def location(self) -> Optional[FileLineColInfo]:
+    def location(self) -> Optional[Union[FileLineColInfo, StackTraceInfo]]:
```
martin-luecke, line 452: consider using new Python syntax: `FileLineColInfo | StackTraceInfo` instead of `Union[FileLineColInfo, StackTraceInfo]` here and at some other places

lit_tests/kernel/wave/location.py
```diff
@@ -116,3 +124,100 @@ def add_no_loc_info(
...
+        b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+    ):
+        a_reg = tkw.read(a, elements_per_thread=16)
+        res = a_reg + a_reg
+
+    test_stack_trace_dedup = wave_compile(options, test_stack_trace_dedup)
+    print(test_stack_trace_dedup.asm)
+
+    # Ensure that stack trace location is deduplicated to avoid IR size growth.
+    # In particular, check that both the read and athe addition share a common
```
martin-luecke, line 214:
```suggestion
    # In particular, check that both the read and the addition share a common
```

iree/turbine/kernel/_support/tracing.py
```diff
@@ -60,6 +61,13 @@ def assert_type(a, b):
 
 
 class KernelRegionGraph(RegionGraph):
+    def __init__(
+        self, *, location_capture_config: Optional[LocationCaptureConfig] = None
+    ):
+        super(KernelRegionGraph, self).__init__(
+            location_capture_config=location_capture_config
+        )
+
```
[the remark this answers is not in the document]
ftynse: This may not be required assuming I know the Python object model ;)

PR #881, opened 2025-06-01
[Wave] Implement C = A @ B GEMM example

efric:

PR #936, opened 2025-06-13
[Wave] Fix indexing in unrolling and mapping in multi buffering 

martin-luecke:
This adds proper handling of unrolled ops that index using the induction variable. i.e. they now index correctly to `i+unrollfactor-1` instead of `i`.

Fixes the input and output mapping of reads/writes when doing multibuffering. This should fix the error reported in https://github.com/iree-org/iree-turbine/pull/881#issuecomment-2928065878. 
Latter parts of the codegen depend on the order of items in input and output mapping to be unchanged.
They also depend on the similarity of the keys in this mapping to the sizes of the memory locations. As we change the size of the memory locations, the keys in the mapping have to change as well to retain this similarity.
Simply updating the keys by removing the value and re-adding it with the new key changes the order in this mapping. 
Hence, we add a function to create a new mapping dictionary with identical order and an updated key: `get_dict_with_updated_key`

This also fixes the handling of return values in the iterate subgraph during unrolling.

### DIFF
--- a/iree/turbine/kernel/wave/scheduling/multi_buffering.py
+++ b/iree/turbine/kernel/wave/scheduling/multi_buffering.py
@@ -97,12 +97,14 @@ def _multi_buffer_memory_location(
         i for i, dim in enumerate(original_buffer.shape) if dim == reduction_axis
     ]
     induction_var = tkl.IndexSymbol(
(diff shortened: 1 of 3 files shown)


harsh-nod (approved):
lgtm! thanks!

PR #939, opened 2025-06-16
[Wave] NFC: Remove superfluous reduction subgraph after pipelining

martin-luecke:
The pipelining of a reduction graph creates a new reduction node and a corresponding subgraph. 
The old reduction node is erased, but its subgraph stays in the trace and also shows up during printing of the trace.
This removes the old unused subgraph

### DIFF
--- a/iree/turbine/kernel/wave/scheduling/loop_reconstruction.py
+++ b/iree/turbine/kernel/wave/scheduling/loop_reconstruction.py
@@ -667,8 +667,9 @@ def construct_pipelined_loop(
         visualize,
     )
 
-    # Remove the unpipelined reduction.
+    # Remove the unpipelined reduction and the corresponding subgraph
     reduction.graph.erase_node(reduction.fx_node)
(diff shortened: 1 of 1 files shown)


tgymnich (approved):

iree/turbine/kernel/wave/scheduling/loop_reconstruction.py
```diff
@@ -667,8 +667,9 @@ def construct_pipelined_loop(
         visualize,
     )
 
-    # Remove the unpipelined reduction.
+    # Remove the unpipelined reduction and the corresponding subgraph
     reduction.graph.erase_node(reduction.fx_node)
```
Hardcode84, line 671: maybe use `CustomOp.erase` and handle subgraph removel there?

martin-luecke: I would prefer that too, but in the `erase` function of Iterate I do not have access to the trace, which owns the subgraphs.

PR #953, opened 2025-06-19
Adds a generic iree_turbine `torch.compile` backend

zjgarvey:
This simply imports the graph module (and expands custom ops), creates a `Launchable` through `jit_compile`, and preloads it for the provided args. 

It seems like the async call crashes for no-input functions like that in `testImportAtenFull`. I've temporarily disabled async invocation for situations when the function has no inputs (this is rare), but we should address issue #984 promptly and remove the workaround.


iree/turbine/dynamo/backends/base.py
```diff
@@ -0,0 +1,68 @@
...
+    module_op = fx_importer.module_op
+    logger.debug("Successfully imported gm to mlir:\n%s", module_op)
+    expansion_pass = ExpandCustomOpsPass(module_op)
+    expansion_pass.run()
+
+    with module_op.context:
+        pm = PassManager.parse("builtin.module(torch-to-iree)")
+        pm.run(module_op)
+
+    logger.debug("Successfully lowered to iree-input:\n%s", module_op)
```
zjgarvey, line 40: We can remove this step.

iree/turbine/dynamo/backends/base.py
```diff
@@ -0,0 +1,68 @@
...
+)
+
+
+def _backend(gm: torch.fx.GraphModule, example_inputs):
+    """Generic backend which creates and preloads a launchable."""
+    # Export the graph module to mlir.
+    gm = turbine_cpu_pass_pipeline(gm, example_inputs)
+    logger.debug("Traced Graph Module:\n%s", str(gm))
+    fx_importer = FxImporter()
+    # TODO: figure out how to create a backend based on a dynamo `ExportedProgram`.
```
AaronStGeorge, line 29: Is an `ExportedProgram` a separate thing? My mental model `torch.compile` -> `torch.fx.GraphModule` that might have graph breaks, `torch.export` -> `torch.fx.GraphModule` + exception is thrown if there are graph breaks.

zjgarvey:
> LGTM, though I'm a newbie.
> 
> I thought a torch.compile backend already existed `iree-turbine`? If this isn't the only backend, what makes this one different?

We do have a `torch.compile` backend, but it is out-dated and only for cpu. This should work for both cpu and gpu and also has support for CustomOp decomposition. We should probably add a `deprecated` warning for the other backend in this PR, so I'll be sure to add that in.

zjgarvey:
> > We do have a `torch.compile` backend, but it is out-dated and only for cpu. This should work for both cpu and gpu and also has support for CustomOp decomposition. We should probably add a `deprecated` warning for the other backend in this PR, so I'll be sure to add that in.
> 
> Is there a reason we don't update and expand the existing torch.compile backend to allow gpu instead of creating a new backend?

Yeah, the old backend is named `turbine_cpu`, so it wouldn't make sense to use that unless we renamed it. I'm not sure if this is used by anyone, so I didn't want to just throw it away.

setup.py
```diff
@@ -106,6 +106,7 @@ def initialize_options(self):
     entry_points={
         "torch_dynamo_backends": [
             "turbine_cpu = iree.turbine.dynamo.backends.cpu:backend",
+            "iree_turbine = iree.turbine.dynamo.backends.base:backend",
```
IanNod, line 109: > Yeah, the old backend is named `turbine_cpu`, so it wouldn't make sense to use that unless we renamed it. I'm not sure if this is used by anyone, so I didn't want to just throw it away.

The new base backend looks to support cpu as well. If we are concerned about removing `turbine_cpu` for people who may be using it (I don't think there is but better safe than sorry) then I would suggest mapping both `turbine_cpu` and `iree_turbine` to the same ```iree.turbine.dynamo.backends.base:backend```. This will remove duplicated code and make general maintenance easier later down the road.

PR #958, opened 2025-06-20
[wave] make cmake execute_process errors fatal

ftynse:
Otherwise the process just continues with empty output variables and fails later with incomprehensible errors.


Hardcode84 (approved):

ftynse:
The machine running nodai-amgdpu-mi250 tests seems broken or out of disk space.

PR #961, opened 2025-06-23
Proof-of-life connection between Wave and a C++ opt tool

ftynse:
Introduce a minimalist connection between Wave kernels and C++ opt-style tool performing a out-of-bounds access checks using dataflow analyses. The tool is expected to be precompiled and installed on the system.

### DIFF
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -139,6 +139,10 @@ jobs:
           cat ${GITHUB_WORKSPACE}/water/$LLVM_SHA_FILE >> $GITHUB_ENV
           echo 'EOF' >> $GITHUB_ENV
 
+          echo 'WAVE_TEST_WATER<<EOF' >> $GITHUB_ENV
+          echo '1' >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+
       - name: Setup env
         if: "(!contains(toJSON(matrix.os), 'amdgpu') && contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
         run: |
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -22,7 +22,91 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
   cancel-in-progress: true
 
+env:
+  LLVM_SHA_FILE: llvm-sha.txt
+  LLVM_CACHE_NUMBER: 1  # Increase to reset cache
+  WATER_REF: main
+
 jobs:
+  build_llvm_linux:
+    name: Builds Linux LLVM/MLIR
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [linux-mi300-1gpu-ossci-iree-org]
+    runs-on: ${{matrix.os}}
+    defaults:
+      run:
+        shell: bash -el {0}
+    timeout-minutes: 420
+    steps:
+      - name: Checkout Water
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: ftynse/water
+          ref: ${{ env.WATER_REF }}
+          path: water
+          fetch-depth: 0
+
+      - name: Setup Cache Vars
+        run: |
+          echo 'LLVM_SHA<<EOF' >> $GITHUB_ENV
+          cat ${GITHUB_WORKSPACE}/water/$LLVM_SHA_FILE >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+
+      - name: Setup env
+        run: |
+          sudo apt update
+          sudo apt install -y ninja-build cmake clang lld
+
+      - name: Cache LLVM-MLIR
+        id: cache-llvm-mlir
+        uses: actions/cache@v3
+        with:
+          path: llvm-mlir/_mlir_install/**
+          key: ${{ runner.os }}-build-llvm-${{ env.LLVM_CACHE_NUMBER }}-${{ env.LLVM_SHA }}
+
+      - name: Checkout LLVM
+        if: steps.cache-llvm-mlir.outputs.cache-hit != 'true'
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: llvm/llvm-project
+          ref: ${{ env.LLVM_SHA }}
+          path: llvm-mlir/llvm-project
+          fetch-depth: 0
+
+      - name: Build LLVM-MLIR
+        if: steps.cache-llvm-mlir.outputs.cache-hit != 'true'
+        timeout-minutes: 420
+        run: |
+          pushd ${GITHUB_WORKSPACE}/llvm-mlir
+          echo "INFO: Need to rebuild LLVM-MLIR. Previous installation for MLIR not found"
+          np=`nproc`
+          echo "INFO: nproc $np"
+          mkdir _build
+          cd _build
+          export CC=clang
+          export CXX=clang++
+          cmake ../llvm-project/llvm                                           \
+            -GNinja                                                            \
+            -DCMAKE_BUILD_TYPE=Release                                         \
+            -DLLVM_ENABLE_PROJECTS="mlir"                                      \
+            -DLLVM_ENABLE_ASSERTIONS=ON                                        \
+            -DLLVM_INSTALL_UTILS=ON                                            \
+            -DLLVM_TARGETS_TO_BUILD="X86;AMDGPU"                               \
+            -DLLVM_ENABLE_BINDINGS=OFF                                         \
+            -DLLVM_ENABLE_ZSTD=OFF                                             \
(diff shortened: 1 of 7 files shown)


ftynse:
@Hardcode84 @harsh-nod @ScottTodd  please advise on how to better test this. As discussed [previously](https://discord.com/channels/689900678990135345/689906000043573354/1384220851129548910), the C++ code is a separate repo, but we still want to run tests somehow.

Hardcode84:
We can pin Wave to the specific Water commit and then build and cache LLVM/MLIR + Water on CI the same way I did previously https://github.com/ftynse/water/pull/2 so it will only be rebuilt if pinned commit has changed

ftynse:
I expect that to be quite frequent, at least in the initial period. Rebuilding Water itself shouldn't take much time compared to its LLVM dependency. If we can somehow cache that, it may be sufficient. Ideally, we'd share the cache between projects (including IREE proper) somehow. We had this working with buildkite + cloud storage, but I don't think Github actions are powerful to let you have a network mount...

Hardcode84:
These caches are per repo so I don't think they can be shared. But, we can cache LLVM/MLIR separately and it will only be rebuilt when llvm sha is changed, which will be less frequent (~once a week, if we follow IREE integrates)

iree/turbine/kernel/wave/water.py
```diff
@@ -0,0 +1,234 @@
...
+        return int(gpu_d.AddressSpace.Workgroup)
+    if memory_space == Attribute.parse("#gpu.address_space<global>"):
+        return int(gpu_d.AddressSpace.Global)
+    if memory_space == Attribute.parse("#gpu.address_space<private>"):
+        return int(gpu_d.AddressSpace.Private)
+    return None
+
+
+def _deiree(module: Module) -> str:
+    """Return a copy of the module without IREE-specific operations, suitable for MLIR processing."""
```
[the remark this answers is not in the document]
Hardcode84: This code is called before we call the IREE pipeline so we (Wave) have a full control of what dialects kernel contains.
Speaking of dispatch workgroups, I have already switched them to upstream gpu dialect https://github.com/iree-org/iree-turbine/pull/959

The high level idea is to generate Wave kernel using 100% upstream dialects and the generate IREE wrapper later so I hope this code is temporary.

Hardcode84:
undrafting so tests can actually run

ftynse:
@Hardcode84 is there a way to force CI to pick up the bumped version of water/main?

Hardcode84:
> @Hardcode84 is there a way to force CI to pick up the bumped version of water/main?

It should just take current main each tume you rerun the CI, but you can pin the specific commit here https://github.com/iree-org/iree-turbine/pull/961/files#diff-d7a58e7c9e1a8fd59cb063aa92b4deeee3150aa277fc8fed28451ea1bc8a2ddfR28

Hardcode84:
(CI tests seems to be stuck)

Also, it may be a good idea by itself to pin Water commit, to avoid random breakages.

Hardcode84:
@ScottTodd @ftynse I think it's ready for review

.github/workflows/ci-tk.yaml
```diff
@@ -22,7 +22,90 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
   cancel-in-progress: true
 
+env:
+  LLVM_SHA_FILE: llvm-sha.txt
+  LLVM_CACHE_NUMBER: 1  # Increase to reset cache
+  WATER_REF: 1a84aa14ee74d5015ceb41bed438de4389f59b83
```
ScottTodd, line 28: Please add a comment here, linking to https://github.com/ftynse/water
```suggestion
  # Git ref from https://github.com/ftynse/water to checkout
  WATER_REF: 1a84aa14ee74d5015ceb41bed438de4389f59b83
```

.github/workflows/ci-tk.yaml
```diff
@@ -22,7 +22,90 @@ concurrency:
...
+
 jobs:
+  build_llvm_linux:
+    name: Builds Linux LLVM/MLIR
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [linux-mi300-1gpu-ossci-iree-org]
```
ScottTodd, line 36: This is not an "os" (operating system), it is a runner label
```suggestion
        runs-on: [linux-mi300-1gpu-ossci-iree-org]
```

.github/workflows/ci-tk.yaml
```diff
@@ -22,7 +22,90 @@ concurrency:
...
+    runs-on: ${{matrix.os}}
+    defaults:
+      run:
+        shell: bash -el {0}
```
ScottTodd, line 40: Why `-el {0}`? Is that load bearing?

.github/workflows/ci-tk.yaml
```diff
@@ -22,7 +22,90 @@ concurrency:
...
+          echo 'EOF' >> $GITHUB_ENV
+
+      - name: Setup env
+        run: |
+          sudo apt update
+          sudo apt install -y ninja-build cmake clang lld
+
+      - name: Cache LLVM-MLIR
+        id: cache-llvm-mlir
+        uses: actions/cache@v3
```
ScottTodd, line 64: Pin actions to commit hashes. Refs can be changed out from under us and we want to pin all dependencies for security reasons.

.github/workflows/ci-tk.yaml
```diff
@@ -22,7 +22,90 @@ concurrency:
...
+            -DLLVM_ENABLE_BINDINGS=OFF                                         \
+            -DLLVM_ENABLE_ZSTD=OFF                                             \
+            -DMLIR_INCLUDE_TESTS=OFF                                           \
+            -DLLVM_USE_LINKER=lld                                              \
+            -DLLVM_DISTRIBUTION_COMPONENTS="llvm-headers;llvm-libraries;cmake-exports;FileCheck;count;not;mlir-headers;mlir-libraries;mlir-cmake-exports;mlir-tblgen" \
+            -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/llvm-mlir/_mlir_install
+          ninja install-distribution-stripped
+          cp bin/FileCheck ${GITHUB_WORKSPACE}/llvm-mlir/_mlir_install/bin/
+          cp bin/count ${GITHUB_WORKSPACE}/llvm-mlir/_mlir_install/bin/
+          cp bin/not ${GITHUB_WORKSPACE}/llvm-mlir/_mlir_install/bin/
```
ScottTodd, line 105: Why do these specific tools need to be manually copied over?

.github/workflows/ci-tk.yaml
```diff
@@ -22,7 +22,90 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
   cancel-in-progress: true
 
+env:
+  LLVM_SHA_FILE: llvm-sha.txt
+  LLVM_CACHE_NUMBER: 1  # Increase to reset cache
+  WATER_REF: 1a84aa14ee74d5015ceb41bed438de4389f59b83
+
 jobs:
+  build_llvm_linux:
```
ScottTodd, line 31: Can you add a comment here explaining why LLVM needs to be built from source, perhaps discussing alternatives to consider in the future?

.github/workflows/ci-tk.yaml
```diff
@@ -32,12 +115,47 @@ jobs:
...
+      - name: Setup Cache Vars
+        if: "(!contains(toJSON(matrix.os), 'amdgpu') && contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
+        run: |
+          echo 'LLVM_SHA<<EOF' >> $GITHUB_ENV
+          cat ${GITHUB_WORKSPACE}/water/$LLVM_SHA_FILE >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+
+          echo 'WAVE_TEST_WATER<<EOF' >> $GITHUB_ENV
+          echo '1' >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
```
ScottTodd, line 143: What are these EOF values doing here? Could use a comment, or move this into a script.

.github/workflows/ci-tk.yaml
ftynse: I'd love to be able to just link to libMLIR.so and have iree do the same, but I don't think that was ever made to work across platforms.

Hardcode84: Please no, dynamic LLVM/MLIR linking is huge footgun and I would prefer for it to not exist :)

.github/workflows/ci-tk.yaml
Hardcode84: They are needed for LLVM/MLIR lit tests and they are not part of the distribution (or I didn't found the way to include them). I'm not running water lit tests here yet, so I can remove them for now.

.github/workflows/ci-tk.yaml
Hardcode84: Copypasted from elsewhere :) removed

.github/workflows/ci-tk.yaml
Hardcode84: added the comment

ftynse:
Looks like everything passes now, we need an approval

PR #962, opened 2025-06-23
[Wave] miscellaneous Python fixes

ashay:
This PR contains three, largely non-functional changes:

> [Wave] fix assertions so that they're not implicitly true
> 
> In Python, assertions on non-empty tuples always pass.  Prior to this
> change, the assertions included not just the conditions but also the
> error messages, constructed together as a tuple.  This patch decouples
> the error message from the condition so that the assertions actually
> check the condition.

and

> [Wave] nfc: remove duplicate "shuffle" key from dictionary init
> 
> The `op_map` dictionary included the `shuffle` key twice (luckily, both
> times it was initialized with the same value: `wave.shuffle`).  This
> patch removes the duplication, and to prevent such problems from
> reoccurring, it sorts the list is alphabetical order.

and

> [Wave] fix comparisons with `None`
> 
> In Python, `None` comparisons are tricky.  The `==` operator checks for
> equivalence and it can return True under specific conditions like with
> the `__eq__` function is defined for the class whose objects are
> compared.  Instead, the reliable way to check for `None` is using the
> `is` operator, which checks for identity instead of equivalence.


ashay:
Thanks @Hardcode84 for approving the CI run and the changes.  Mind merging it please (I don't have commit access to the repo)?

PR #964, opened 2025-06-24
Cube root (CBRT) op

rohan-tan-bhowmik:
Implemented CBRT unary op from math dialect. Enjoy!


iree/turbine/kernel/wave/codegen/handlers.py
```diff
@@ -55,6 +55,7 @@
     bitcast,
     cast,
     conditional,
+    cbrt,
```
Hardcode84, line 58: please keep this list sorted

rohan-tan-bhowmik: Sorry for bad english it is my first language

Hardcode84 (approved):

PR #967, opened 2025-06-25
[BOO] Add Graph-Based Fusion API

zjgarvey:
Adds some tooling for replacing subgraphs with auto-generated iree launchable calls compatible with autograd and torch.compile.

### Approach

1. Identify root ops we want to offload to iree.
2. Indicate good consumer/producer fusion opportunities for each root op. (See `FusionSchema` and  `OpFusionSpec`).
3. Given a `torch.nn.Module`, use `torch.export.export` to get an `ExportedProgram` (contains a `GraphModule`).
4. Extract all subgraphs consisting of root op fusions.
5. Replace each subgraph with a single node representing an iree-turbine `Launchable` wrapped in an `autograd.Function`.
6. Return the module associated with the `ExportedProgram`, which applies the modified graph.

### Some considerations

This is currently aot-only, as we are swapping subgraphs in an aot-traced dynamo export. We could add support for dynamic shapes in the future, or try to build something which will jit export a graph and do the replacement on the fly. However, I feel like those use-cases might be better served by using our `torch.compile` backend selectively.

The user currently needs to provide a `FusionSchema`, but we should provide defaults for supported fusions.

The auto-generated graph ops in `fusion_transform` do not explicitly support automatic mixed-precision contexts, nor do they have custom-logic for alternative memory formats such as `torch.channels_last`. In order to make this work compatible with `boo_convolution`, which does support alternative memory formats, we need to handle this properly. We should also make `boo_convolution` a turbine `CustomOp` so we can export graphs with `boo_convolution` ops.

We are not attaching the `make_single_dispatch` attribute to the generated mlir. It might be good to enable this for forward fusion graphs by default.

### Some critical issues

1. There may be an issue with subgraph replacement when multiple subgraphs match one pattern (but the first of the two subgraphs had inputs which did not require gradient). It might be a good idea to do something in-house for subgraph replacement, since we know exactly where we want to put the replacements. Alternatively, we could make the backward logic a bit more careful, but I'm not sure how to make this work yet.


iree/turbine/kernel/boo/examples/subgraph_extraction.py
```diff
@@ -0,0 +1,99 @@
...
+        class FakeMod(torch.nn.Module):
+            def forward(self, *args):
+                return custom_op(*args)
+
+        e = torch.export.export(FakeMod(), args=fake_args)
+        subgraph_repl.append(e.graph_module)
+
+    new_gm = replace_subgraphs(gm, subgraphs, subgraph_repl)
+
+    new_gm.print_readable()
```
zjgarvey, line 92: Note: the pytorch subgraph rewriter tool seems to match the first `conv + relu` with the second, so something needs to be done to disambiguate these (otherwise the shapes are all messed up).

iree/turbine/kernel/boo/fusion/auto.py
```diff
@@ -0,0 +1,54 @@
...
+    for idx, arg in enumerate(args):
+        if arg is not None and not isinstance(arg, torch.Tensor):
+            raise TypeError(
+                f"Expected all function arguments to be (optional) tensors. Got {type(arg)} at position {idx}."
+            )
+        name += f"_{tensor_type_str(arg)}"
+    return name
+
+
+def kernel(fn: Callable | torch.nn.Module) -> Callable:
```
zjgarvey, line 36: This is just a useful tool for getting a Launchable from any dynamo-exportable python callable. 

e.g.

```python
from iree.turbine.kernel.boo.fusion import kernel

@kernel
def my_sample_fused_op(*args):
    ...
```

iree/turbine/kernel/boo/fusion/subgraph.py
```diff
@@ -0,0 +1,119 @@
...
+    src_gm: GraphModule, fusion_schema: FusionSchema
+) -> Tuple[Sequence[GraphModule], Sequence[Dict[Node, Node]]]:
+    """Traverses src_gm nodes in order. When a node matches a root op in the fusion_schema,
+    A new subgraph is created with the root op in addition to any adjacent nodes matching the schema.
+
+    All subgraphs found this way are returned in a list.
+    The corresponding mappings from src_gm nodes -> subgraph nodes is also returned as a list.
+    """
+    subgraph_projections: List[Dict[Node, Node]] = []
+    subgraph_inclusions: List[Dict[Node, Node]] = []
```
zjgarvey, line 29: I'm putting these here only because I might need them later on. If I don't end up needing these, I'll remove them.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,150 @@
...
+def mlir_from_gm(gm: GraphModule) -> Any:
+    sample_args = tuple(
+        [n.meta.get("val") for n in gm.graph.nodes if n.op == "placeholder"]
+    )
+    gm = turbine_cpu_pass_pipeline(gm, sample_args)
+    imp = FxImporter()
+    imp.import_graph_module(gm)
+    exp_ops = ExpandCustomOpsPass(imp.module_op)
+    exp_ops.run()
+    return imp.module_op
```
zjgarvey, line 68: Currently unused. Will likely remove.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # module_op = mlir_from_gm(gm)
+    arg_factory = lambda: tuple(
+        [n.meta.get("val") for n in gm.graph.nodes if n.op == "placeholder"]
+    )
+
+    @register_impl(op_name)
+    def _(*args):
+        l = get_launchable(gm, arg_factory=arg_factory, func_name=op_name)
+        return l(*[arg.data for arg in args])
```
zjgarvey, line 90: Remove `arg_factory` and use `args`. Further specialize func name like in `auto.py` above. 

I'm hoping if this change is made, even if the graph module metadata is different from incoming inputs, the launchable will at least still work for the new inputs. Needs to be verified.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,150 @@
...
+        return getattr(torch.ops.boo, op_name)(*args)
+
+    return _f
+
+
+def make_autograd_function(
+    joint_gm: torch.fx.GraphModule, sample_args, num_fwd_outputs
+):
+    """From a joint forward/backward graph module, creates an autograd function for calling iree custom graph ops for forward and backward."""
+    fwd_g, bwd_g = default_partition(
```
zjgarvey, line 125: Could maybe combine this with `@kernel` to automatically generate an autograd function from a python callable.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,150 @@
...
+                torch.empty_strided(
+                    list(o.shape),
+                    stride=o.stride,
+                    dtype=o.dtype,
+                    device="meta",
+                    requires_grad=o.requires_grad,
+                )
+                for o in outputs
+            )
+        )
```
zjgarvey, line 113: This needs to be reworked to be valid for shapes outside the original.

iree/turbine/kernel/boo/fusion/subgraph.py
```diff
@@ -0,0 +1,118 @@
...
+
+def extract_fusion_subgraph_modules(
+    src_gm: GraphModule, fusion_schema: FusionSchema
+) -> Tuple[Sequence[GraphModule], Sequence[Dict[Node, Node]]]:
+    """Traverses src_gm nodes in order. When a node matches a root op in the fusion_schema,
+    A new subgraph is created with the root op in addition to any adjacent nodes matching the schema.
+
+    All subgraphs found this way are returned in a list.
+    The corresponding mappings from src_gm nodes -> subgraph nodes is also returned as a list.
+    """
```
AaronStGeorge, line 27: What prevents us from using the [SubgraphMatcher](https://github.com/pytorch/pytorch/blob/8597d37536ef11bdf6b0a539ab79af876e1c92f6/torch/fx/passes/utils/matcher_utils.py#L51)?

iree/turbine/kernel/boo/examples/subgraph_extraction.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+        class FakeMod(torch.nn.Module):
+            def forward(self, *args):
+                return custom_op(*args)
+
+        e = torch.export.export(FakeMod(), args=fake_args)
+        subgraph_repl.append(e.graph_module)
+
+    new_gm = replace_subgraphs(gm, subgraphs, subgraph_repl)
```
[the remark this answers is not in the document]
zjgarvey: I decided to make a graph directly with the single node. This allows me to call the autograd function directly, otherwise the subgraph replacement "forgets" the backward pass of the single traced custom op.

iree/turbine/kernel/boo/fusion/auto.py
```diff
@@ -0,0 +1,75 @@
...
+        launch = get_launchable(
+            lambda: FakeMod(),
+            args,
+            func_name=func_name,
+            cache_only=False,
+            force_single_dispatch=False,
+        )
+        return launch(*[arg.data for arg in args])
+
+    return wrapped
```
zjgarvey, line 75: Ah, I didn't mean to push this change. I need to rework similar to `ops/graph.py` to allow autograd for this decorator.

iree/turbine/kernel/boo/fusion/apply.py
```diff
@@ -0,0 +1,62 @@
...
+    _log_graph_module,
+)
+from ....support.logging import aot_logger as logger
+
+__all__ = [
+    "fusion_transform",
+]
+
+
+def fusion_transform(
```
nicolasvasilache, line 25: If I understand correctly, this is a FX-level graph partitioning where each subgraph will become a single dispatch region? If this is correct, could you explain what is the expected process if IREE later does not manage to fuse the whole subgraph in a single toplevel scf.forall op?

iree/turbine/kernel/boo/fusion/auto.py
```diff
@@ -0,0 +1,42 @@
...
+    "kernel",
+]
+
+
+def kernel(fn: Callable) -> Callable:
+    """Decorator to convert a dynamo exportable function into a launchable.
+
+    This is not-yet compatible with the following:
+        1. torch.compile (we aren't wrapping in a custom op).
+        2. autograd (we aren't defining backward).
```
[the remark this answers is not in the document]
zjgarvey: Yeah, I think we'd like to eventually put backward ops in this same fusion schema.

Currently, if boo backward kernels are enabled, we are compiling the whole backward graph corresponding to the forward fusion subgraph, but it might be better to select only parts of the backward subgraph to offload to IREE. 

There is a limitation that we can only "see" the part of the backward graph corresponding to the replacement forward subgraph. If we wanted to fuse backward ops with anything outside of that, we would need to start from a joint training graph for the whole model to begin with. 

I'm not sure if starting from a joint training graph is desirable or not, but it could certainly be done, and we wouldn't need to worry so much about defining a backward pass for the forward replacements.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,186 @@
...
+    none_output = []
+    for n in gm.graph.nodes:
+        if n.op == "output":
+            trunc_returns = []
+            for ret in n.args[0]:
+                if ret is None:
+                    none_output.append(True)
+                    continue
+                trunc_returns.append(ret)
+                none_output.append(False)
```
ftynse, line 142: Something along these lines

```python
is_none = [ret is None for ret in n.args[0]]
trunc_returns = [ret for ret in in.args[0] if ret is not None]
```

will be more compact and slightly more efficient.

iree/turbine/kernel/boo/fusion/subgraph.py
```diff
@@ -0,0 +1,181 @@
...
+    subgraph_projections: List[Dict[Node, Node]] = []
+    subgraph_inclusions: List[Dict[Node, Node]] = []
+    subgraphs: List[GraphModule] = []
+    used_nodes: Set[Node] = set()
+    for root in src_gm.graph.nodes:
+        # if this node is included in any subgraph already, we can't use it.
```
ftynse, line 87: We probably want to warn or at least debug log that it is the case here. In future, we may also want a non-greedy approach to deciding which of the potentially overlapping fusions to implement.

iree/turbine/kernel/boo/fusion/subgraph.py
```diff
@@ -0,0 +1,181 @@
...
+            # This results in a breadth-first traversal of producers.
+            curr_node = worklist.pop(0)
+            for producer in curr_node.all_input_nodes:
+                if producer.op != "call_function":
+                    continue
+                if producer.target not in node_spec.producers:
+                    continue
+                if producer in used_nodes:
+                    continue
+                # Insert producers at the front, since we want to preserve at least some weak ordering of nodes.
```
ftynse, line 113: I'd say we want to have producers in the same order as they appear in the graph, and we are currently reversing this order. It looks possible to construct the following scenario:

```python 
producer1 = call(...)
producer2 = call(producer1). # e.g. compute the mean of producer1
root = call(producer1, producer2) # e.g. normalize dividing by mean
```

where inverting the order of producers looks invalid.

What we need is a topological sort, which can be implemented either as DFS + visited set or as BFS + node degree counters. We could either traverse producers directly in this order, or just sort all nodes once after everything is collected.

iree/turbine/kernel/boo/fusion/schema.py
```diff
@@ -0,0 +1,25 @@
...
+from torch.fx.node import Target
+
+# import torch
+
+
+@dataclass
+class OpFusionSpec:
+    recursive: bool = True
+    producers: Sequence[Target] = ()
+    consumers: Sequence[Target] = ()
```
ftynse, line 18: So this spec supports only "homogeneous" fusions, e.g., it will prescribe to always fuse, e.g., elementwise operations (potentially recursively), but will not be able so say "only fuse elementwise transitive producers that feed into the first root producer" or "only fuse elementwise transitive producers that feed into a matmul producer", or anything beyond the type. This is worth documenting clearly. Depending on the complexity of the subgraphs we will want to replace, we may want to think of a more complex scheme, perhaps itself `Graph` with some additions to indicate (un)bounded recursion and other wildcards.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,186 @@
...
+                    for val in node.args[0]
+                ]
+            )
+    return inputs, meta_outputs
+
+
+def get_schema(inputs, outputs):
+    """Generate a schema from the result of `get_io_from_gm`."""
+
+    ret_ty = "Tensor?" if any([o is None for o in outputs]) else "Tensor"
```
ftynse, line 58: Shouldn't we use `Tensor?` only for outputs that may actually be null, not for all outputs if one of them may be null?

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,186 @@
...
+    schema = "("
+    schema += ", ".join([f"Tensor {inp}" for inp in inputs])
+    schema += ") -> ("
+    schema += ", ".join([ret_ty for _ in outputs])
+    schema += ")"
+    return schema
+
+
+def get_mlir_module(gm: GraphModule) -> Any:
```
ftynse, line 67: Nit: we might as well not use typing rather than use `Any` :)

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,186 @@
...
+    """Generates torch-mlir IR from a graph module."""
+    sample_args = tuple(
+        [n.meta.get("val") for n in gm.graph.nodes if n.op == "placeholder"]
```
ftynse, line 70: I see this snippet repeatedly. It's short, but may be worth factoring out into a named function for readability.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -0,0 +1,186 @@
...
+    imp = FxImporter()
+    imp.import_graph_module(gm)
+    exp_ops = ExpandCustomOpsPass(imp.module_op)
+    exp_ops.run()
+    return imp.module_op
+
+
+def get_custom_graph_op(gm: GraphModule) -> Callable[[Any], Any]:
+    """Converts a graph module into a custom operator."""
+    hash = sha1(str(gm).encode(), usedforsecurity=False).hexdigest()
```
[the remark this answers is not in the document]
ftynse: An alternative could be to come up with a naming scheme based on the fusion spec.

iree/turbine/kernel/boo/ops/graph.py
zjgarvey: So my primary concern is that if any one of the outputs is `None`, it's possible that in some other situation, the same op may return `None` for a different output.

The primary example would be something like `convolution_backward`, which returns `None` if the corresponding forward convolution input does not require gradient calculation. Fundamentally, this isn't really robust enough to cover those cases, since if a backward graph op did happen to require gradients for all inputs, we still wouldn't be making the outputs optional as they should be.

I think maybe a better approach would be to look at the producers for each output of the source subgraph, and identify the type for that output from the schema of the producing torch op. 

At the end of the day, this might not be strictly necessary since we are getting an aot-exported graph. I'll need to think about this a bit more carefully.

iree/turbine/kernel/boo/fusion/apply.py
zjgarvey: I've at least included better error messaging for when we get a compiler crash for now. Anyone using boo should at least get a good printout with some instructions for filing an IREE issue with a reproducer.

iree/turbine/kernel/boo/ops/graph.py
zjgarvey: todo for later

iree/turbine/kernel/boo/fusion/subgraph.py
zjgarvey: I'm going to keep them for now, since a resolution to https://github.com/iree-org/iree-turbine/issues/1014 might warrant using these for a custom subgraph replacement.

iree/turbine/kernel/boo/fusion/schema.py
zjgarvey: Opened https://github.com/iree-org/iree-turbine/issues/1015 to track for future updates.

iree/turbine/kernel/boo/ops/graph.py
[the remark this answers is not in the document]
zjgarvey: For now, I'm going to simply debug-log the hash/graph module pair.

iree/turbine/kernel/boo/fusion/subgraph.py
```diff
@@ -0,0 +1,206 @@
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from typing import Sequence, Set, List, Dict
```
ftynse, line 7: Nit: I think all of `set`, `list`, and `dict` are now subscriptable with `[]` and their capitalized equivalences from `typing` are no longer necessary.

ftynse (approved):

Issue #941, opened 2025-06-16
[arch] location for C++ support code


ftynse:
We'd like to be able to have C++ code supporting Wave specifically (though other parts of the Turbine can be added if needed) and available to it. The code in question is currently broadly scoped to:
 - MLIR op definitions, e.g., a potential Wave dialect
 - Passes using upstream MLIR operations, potential Wave dialect operations and optionally IREE operations; e.g. what is currently prototyped in https://github.com/ftynse/llvm-project/blob/sanitizers/mlir/lib/Dialect/MemRef/Transforms/AssertInBounds.cpp. (We don't strictly need IREE operations and can do ugly psuedo-`dyn_cast` via `getOperaitonName` and stringly APIs).
 - Implementations of small runtime calls invoked by IR generated by the passes above. Think upstream CRunnerUtils or libasan.so.
 - (Hopefully temporary) divergence from upstream passes, such as disabling constant folding in canonicalization.

The alternatives we currently see are:
 - Put C++ in IREE
   + Pro: don't shard the codebase further
   + Pro: can reuse CI and other development setup
   - Con: this code isn't needed for IREE itself
   - Con: iteration speed too slow for the initial bringup
 - Put C++ into a new repo
   - Option A: iree-turbine has the new repo as a submodule and it integrates into the build system
   - Option B: the new repo creates Python wheels that get `pip install`ed in iree-turbine like iree itself. Note that this may run into the infamous MLIR linking problem, but can be partially worked around by only interacting with statically linked command-line tools.
   + Pros & Cons: the inverse of above

We can also consider inverting the dependency between IREE and Wave and have Wave only use upstream MLIR + interfaces; the Wave/IREE integration then lives in IREE and IREE-specific IR is modeled via interfaces for Wave.

ftynse:
@stellaraccident @ScottTodd ^

PR #969, opened 2025-06-25
[Wave] Include wave runtime in turbine requirements.txt

harsh-nod:
This PR includes the wave runtime in the requirements.txt so that we don't need to build it separately in downstream projects.


harsh-nod:
@ScottTodd - wanted your thoughts on this as we get ready for upstreaming wave kernels into sglang? Currently, the wave kernels all use the wave runtime and so wanted to minimize the amount of friction downstream users have.

requirements.txt
```diff
@@ -26,3 +26,5 @@ pre-commit
 
 # For scheduling in wave
 ./iree/turbine/kernel/wave/scheduling/aplp
+# Wave runtime
+./iree/turbine/kernel/wave/runtime
```
ScottTodd, line 30: > This PR includes the wave runtime in the requirements.txt so that we don't need to build it separately in downstream projects.

I don't have time for a detailed review here, but what specifically are you trying to accomplish by moving into the root `requirements.txt`? Packages published to pip use https://github.com/iree-org/iree-turbine/blob/main/setup.py and https://github.com/iree-org/iree-turbine/blob/main/pyproject.toml, and adding native requirements (C++ or Rust source code) would require deep changes to those files as well as the release process (https://iree.dev/developers/general/release-management/).

See also https://github.com/iree-org/iree-turbine/issues/941, which is setting up a new repository for native compiler source code.

The bar for inclusion in releases is high - code needs to at a minimum be skipped gracefully on unsupported platforms and configurations (Windows/macOS, CPU/Vulkan, etc.).

harsh-nod: So I am trying to get the code lined up for an PR that we are going to submit to sglang. The kernels that are being used use the wave runtime and so I am trying to avoid an extra step of installing the wave runtime after installing iree-turbine. Do we have a ci for packages being published to pip so I can make sure the change works across all the unsupported platforms and configurations as well?

ScottTodd: You can diff those with other projects that have native deps already to see what sort of changes would be needed:
* https://github.com/nod-ai/shark-ai/blob/main/shortfin/build_tools/build_linux_package.sh
* https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/build_packages.yml

Can you elaborate on "PR that we are going to submit to sglang"? You want sglang to take a dep on the iree-turbine Python packages from pypi? Is there an RFC we can review for that?

iree/turbine/kernel/wave/runtime/CMakeLists.txt
```diff
@@ -13,72 +13,101 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
 endif()
 
+# Initialize the flag to false
+set(WAVE_RUNTIME_DISABLED FALSE)
+
 # Detect the installed nanobind package and import it into CMake
 execute_process(
   COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
   OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE nanobind_ROOT
-  COMMAND_ERROR_IS_FATAL ANY)
```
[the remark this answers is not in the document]
harsh-nod: This is intentional because if nano bind is not installed on the system, then it throws an error but I use that to determine that wave runtime should be disabled.

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -129,3 +130,35 @@ NB_MODULE(wave_runtime, m)
...
+};
+
+using Int64Vector = std::vector<uint64_t>;
+using Int32Vector = std::vector<uint32_t>;
+
+static int launch(const KernelLaunchInfo &info, const Int64Vector &tensors,
+    const Int64Vector &dynamicDims, void* scalarArgs)
+{
+    // Runtime is disabled, do nothing
+    return 0;
```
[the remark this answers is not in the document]
harsh-nod: abort() would crash the Python process as well right? I want the calls still to work but I added a debug statement that indicates what is happening so that the user knows what is happening.

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -129,3 +130,35 @@ NB_MODULE(wave_runtime, m)
     m.def("launch", &launch);
     m.def("load_binary", &load_binary);
 }
+
+#else
```
raikonenfnu, line 134: Can we just skip this file or the build all together in cmake side if `WAVE_RUNTIME_DISABLED=TRUE`?

.github/workflows/ci-tk.yaml
```diff
@@ -78,8 +83,8 @@ jobs:
           # wheels saves multiple minutes and a lot of bandwidth on runner setup.
           pip install --no-compile -r pytorch-rocm-requirements.txt
           pip install --no-cache-dir -r requirements-iree-pinned.txt --upgrade
-          pip install -r requirements.txt -e .
-          pip install -r requirements-wave-runtime.txt -e .
+          pip install -r requirements.txt
+          pip install -e . --no-build-isolation
```
raikonenfnu, line 87: Just curious, what is the main reason we need `-no-build-isolation`?

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -129,3 +130,35 @@ NB_MODULE(wave_runtime, m)
...
+// When WAVE_RUNTIME_DISABLED is defined, provide minimal stubs
+#include <vector>
+#include <string>
+#include <cstdint>
+
+struct KernelLaunchInfo
```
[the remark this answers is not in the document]
harsh-nod: In one case KernelLaunchInfo is a nanobind structure and in the other it is a structure I created for the stub implementations so I don't think moving this structure out would be beneficial.

.github/workflows/ci-tk.yaml
harsh-nod: so that we can reuse the packages we have already installed (nano bind & torch) when building wave runtime (and so we don't have to install them again).

iree/turbine/kernel/wave/runtime/runtime.cpp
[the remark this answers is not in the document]
harsh-nod: not needed anymore

iree/turbine/kernel/wave/runtime/CMakeLists.txt
```diff
@@ -13,72 +13,104 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
+# Only proceed with PyTorch setup and build if runtime is enabled
+if(WAVE_RUNTIME_ENABLED)
+  # Link against torch, using the libraries and include dirs in site-packages.
+  execute_process(
+    COMMAND "${Python_EXECUTABLE}" -c "import torch; print(torch.__path__[0])"
+    OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_PATH
+    COMMAND_ERROR_IS_FATAL ANY)
+
+  # Add all directories in /opt/rocm/lib/cmake to CMAKE_PREFIX_PATH
+  file(GLOB subdirectories "/opt/rocm/lib/cmake/*")
```
ScottTodd, line 63: No hardcoded paths to ROCm. This will only work on an increasingly small footprint of devices and platforms.

iree/turbine/kernel/wave/runtime/CMakeLists.txt
```diff
@@ -13,72 +13,104 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
-set(TORCH_LIBRARIES ${torch_PATH}/lib/libtorch.so ${torch_PATH}/lib/libc10.so)
-target_link_libraries(wave_runtime PRIVATE nanobind hip::host ${TORCH_LIBRARIES})
-set_target_properties(wave_runtime PROPERTIES LINK_WHAT_YOU_USE TRUE)
+# Only proceed with PyTorch setup and build if runtime is enabled
+if(WAVE_RUNTIME_ENABLED)
+  # Link against torch, using the libraries and include dirs in site-packages.
+  execute_process(
+    COMMAND "${Python_EXECUTABLE}" -c "import torch; print(torch.__path__[0])"
+    OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE torch_PATH
+    COMMAND_ERROR_IS_FATAL ANY)
```
ScottTodd, line 60: I have no idea if this style of linking against torch C/C++ sources from site-packages is supported, especially for package distribution. Would need @stellaraccident to weigh in.

Either way, I think this code should be spun out into its own package, not part of the iree-turbine package. Keeping a pure Python interface between pytorch and the IREE compiler is quite valuable.

setup.py
```diff
@@ -24,46 +28,141 @@ def load_version_info(version_file):
...
+
+def setup_cmake_environment():
+    """Setup CMake environment with system Python."""
+    python_executable = sys.executable
+    cmake_env = os.environ.copy()
+    cmake_args = []
+
+    print(f"Using Python: {python_executable}")
+
+    return python_executable, cmake_env, cmake_args
```
ScottTodd, line 87: I don't see much value in this function as it exists now. I was confused seeing `python_executable` below instead of `sys.executable`, so the indirection / encapsulation isn't helping readability.

ScottTodd (changes requested):

iree/turbine/kernel/wave/runtime/CMakeLists.txt
harsh-nod: Any suggestions on what I can use instead?

ScottTodd: You can check what https://github.com/pytorch/pytorch/blob/main/cmake/public/LoadHIP.cmake does, though we've been reworking that. The main sections to look at are around `find_package(hip ...)`

harsh-nod: That does have hardcoded paths to ROCM though?

iree/turbine/kernel/wave/runtime/CMakeLists.txt
harsh-nod: Makes sense to keep a pure Python interface between PyTorch and IREE compiler. So would you think we should move wave + dependencies out of iree-turbine into a separate repo and would you have time to discuss what it would take to enable that sometime this week? (ci, pypi packages, etc.)

ScottTodd: I can make some time. I'd prefer for a broader engineering review with a reasonably sized group of stakeholders (5+) over a 1:1 consult, fwiw.

harsh-nod:
Closing in favor of https://github.com/iree-org/iree-turbine/pull/1065

PR #970, opened 2025-06-26
Enable MLIR debug info when compiling standalone kernels

sogartar:
The outstanding unknown location issue is fixed.
https://github.com/iree-org/iree/issues/17403


sogartar:
This does not seem useful, because I get locations like this:
```
#loc6 = loc("-":36:10)
```

rkayaith:
You're probably using the python function `Module.parse` or `Operation.parse` on a string somewhere, which results in `-` as the file name in locations. `Operation.parse` has a `source_name` argument you can use to set the file name to something else, if you have something sensible to use.

sogartar:
I think there is nothing better to do as the kernel is exported in-memory.

PR #971, opened 2025-06-26
[boo] don't wait for Tracy indefinitely

ftynse:
The logic that was supposed to abort if Tracy pipe failed to communicate was in fact re-launching it with no timeout, making the program never exit. Raise an error instead.


iree/turbine/kernel/boo/conv_exports/boo_driver.py
```diff
@@ -209,9 +209,9 @@ def proc_fn():
             try:
                 # Tracy will never exit if it fails to connect, so kill the process after some time.
                 out, err = tracy.communicate(timeout=5)
-            except subprocess.TimeoutExpired:
+            except subprocess.TimeoutExpired as e:
                 tracy.kill()
-                out, err = tracy.communicate()
```
rkayaith, line 214: this is *supposed* to kill the process and then collect whatever output was made, to help with debugging. not sure why it doesn't work, but this is fine too

PR #973, opened 2025-06-26
[Wave] Add scatter_add operation

adedespirlet:
This PR introduces a new Wave operation: “scatter_add” . The scatter_add  operation performs element-wise accumulation from a source register into shared memory (LDS) using dynamic indices along a specified dimension. It supports both integer and floating-point data types and is lowered to memref.atomic_rmw using either addi or addf, depending on the type.

Input arguments:
-	src: a register containing the values to scatter
-	index a register containing the indices along dim where values should be added
-	dim the dimension along which scattering is performed
-	memory: the target tensor in LDS where results are accumulated
-	mapping: the index mapping representing mapping between sets of indices
-	elements_per_thread: number of elements each thread loads in it's VGPR

Currently scatter_add accepts only static one-to-one index mappings. The lowering logic internally performs dynamic indexing , combining index and dim with the provided mapping to compute final memory access indices. _Potential Future Work:_ rewrite the operation handler and API to align with the write() implementation where in case of a scatter operation, index mapping expressions passed through are fully dynamic.

**Limitations:**
Only intra-workgroup scattering is supported (i.e. within LDS/ shared memory) with up to 4 waves. The operation handles multiple elements per thread, provided that the non-scatter dimension is large enough (> elements_per_thread)

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -256,6 +256,16 @@ def select(
 ) -> "Register": ...
 
 
+def scatter_add(
+    register_src: "Register",
+    register_idx: "Register",
+    dim: IndexExpr,
+    memory: "Memory",
+    mapping: IndexMapping,
+    elements_per_thread: Optional[IndexExpr | int] = None,
+) -> "Register": ...
+
+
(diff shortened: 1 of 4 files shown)


ftynse (approved):
LGTM, but please give other folks a chance to review.

PR #974, opened 2025-06-26
[Wave] Adding scatter_max operation for single wave exec

adedespirlet:
This PR introduces a new Wave operation: “scatter_max” . The scatter_max operation performs an element-wise maximum reduction from a source register into shared memory (LDS) using dynamic indices along a specified dimension. It supports both integer and floating-point data types and is lowered to memref.atomic_rmw using either maxs or maximumf, depending on the type.

**Input arguments:**
-	src: a register containing the values to scatter
-	index a register containing the indices along dim where values should be written
-	dim the dimension along which scattering is performed
-	memory: the target tensor in LDS where results are accumulated
-	mapping: the index mapping representing mapping between sets of indices
-	elements_per_thread: number of elements each thread loads -
Tests are included in wave_e2e_test.py to validate the correctness of scatter_max

**Implementation notes:**
Currently scatter_max accepts only static one-to-one index mappings. The lowering logic internally performs dynamic indexing , combining **index** and **dim** with the provided **mapping** to compute final memory access indices. _Potential Future Work_: rewrite the operation handler and API to align with wave's  write() operation  implementation where in case of a scatter operation, index mapping expressions passed through are fully dynamic.

**Limitations:**
Currently, scatter_max only works with a single wave. Using 2 or 4 waves results in incorrect behavior, especially when thread index to the same location. The emitted code is identical to scatter_add when simply swapping the op, suggesting the bug may lie in the iree lowering to LLVM dialect. This needs further investigation. The operation supports multiple elements per thread provided that the non-scatter dimension is large enough (> elements_per_thread)

### DIFF
--- a/iree/turbine/kernel/ops/wave_ops.py
+++ b/iree/turbine/kernel/ops/wave_ops.py
@@ -256,6 +256,16 @@ def select(
 ) -> "Register": ...
 
 
+def scatter_max(
+    register_src: "Register",
+    register_idx: "Register",
+    dim: IndexExpr,
+    memory: "Memory",
+    mapping: IndexMapping,
+    elements_per_thread: Optional[IndexExpr | int] = None,
+) -> "Register": ...
+
+
 def define_op(op_name: str) -> Callable[[T], T]:
     def decorator(cls: T) -> T:
         cls.tkw_op_name = op_name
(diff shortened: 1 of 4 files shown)


ftynse (commented):
Nice!

PR #975, opened 2025-06-26
Add a test for linear layer

AaronStGeorge:


tests/dynamo/backend_test.py
```diff
@@ -45,6 +45,39 @@ def conv_relu_mul(x, y, z, w):
...
+            ),
+        ),
+    ],
+)
+def test_generic_backend_inference_only_linear(device: torch.device):
+    """Simple inference test for linear layer."""
+    model = LinearNet()
+    compiled_model = torch.compile(model, backend="iree_turbine")
+
+    x = torch.randn(8, 3)
```
zjgarvey, line 74: The device should be set for `x` and and the random generator should be seeded for reproducibility.

E.g.,

```python
gen = torch.Generator(device=device)
gen.manual_seed(49)

x = torch.randn((8,3), device=device, generator=gen)
```

zjgarvey (changes requested):

PR #983, opened 2025-06-27
Set GH actions timeouts

Hardcode84:


.github/workflows/update_iree_requirement_pins.yml
ScottTodd, line 1: The jobs in this workflow are unlikely to need timeouts, and this runs on a schedule using standard github-hosted runners, so the cost of reaching the standard timeout is low.

.github/workflows/pre-commit.yaml
```diff
@@ -15,6 +15,7 @@ on:
 jobs:
   pre-commit:
     runs-on: ubuntu-22.04
+    timeout-minutes: 60
```
ScottTodd, line 18: Pre-commit should never hit a 60 minute timeout. If adding a timeout, 5 or 15 minutes might be more appropriate. I would just remove timeouts from this workflow file though.

.github/workflows/ci.yaml
```diff
@@ -87,6 +88,7 @@ jobs:
   ci_summary:
     if: always()
     runs-on: ubuntu-24.04
+    timeout-minutes: 60
```
ScottTodd, line 91: The summary step is small and unlikely to fail. I would remove this timeout.

ScottTodd (commented):
Job-level timeouts will kill processes and cut off any logging. You may also want to use per-test timeouts with https://pypi.org/project/pytest-timeout/. While that uses Python multi-processes tricks and produces lots of noisy output on timeout (compared to something like Bazel or ctest), it at least lets you see test summaries showing how many tests passed, which failed/timed out, etc.

.github/workflows/update_iree_requirement_pins.yml
Hardcode84: I know, but I decided to add it here (and in pre-commit) anyway for uniformity and in case someone starts copypasting these yaml files codes to create another workflow, as it's usually done here :)

ScottTodd: I disagree there. The workflows are doing different things and timeouts make sense for some but not others. Developers working on workflows should be thinking through those details.

.github/workflows/ci-tk.yaml
```diff
@@ -112,6 +113,7 @@ jobs:
       - name: Run expensive e2e tests on AMD GPU
         if: "(contains(toJSON(matrix.os), 'amdgpu') || contains(toJSON(matrix.os), 'mi300')) && (github.event_name != 'pull_request') && !cancelled()"
         run: |
+          export PYTEST_TIMEOUT=600
           WAVE_CACHE_ON=0 pytest -n 4 --capture=tee-sys -vv --run-e2e --run-expensive-tests --durations=100 ./tests/kernel/wave/
```
ScottTodd, line 117: nit: I'd prefer the `--timeout` argument (local to the command) over an environment variable (global state)

https://pypi.org/project/pytest-timeout/

ScottTodd (approved):

PR #995, opened 2025-07-01
Make B, H, M, K2 dynamic dims in bhsd_attention kernel

aviator19941:
This PR is to make the N (head size kv) dimension static and H (num_heads) dimension dynamic and changes the order of the dynamic dims to B, H, M, K2, so the index arguments for the dynamic dimensions are based on the order of the inputs.


raikonenfnu (approved):
LGTM, but let's add commit message and wait for CI to run

PR #999, opened 2025-07-02
[NFC] Add comment for `get_module_asm`

AaronStGeorge:
I was pretty confused what actually happened under the hood when doing something like the following.
```python
import torch
from iree.turbine.kernel.boo.runtime import get_launchable

# could consider adding a bias-add
class SimpleMatmul(torch.nn.Module):
    def forward(self, x, w):
       return torch.matmul(x,w)

device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")

# decide shapes you want to test
x_shape = ...
w_shape = ...

# make some random inputs
gen = torch.Generator(device=device)
gen.manual_seed(10)
x0 = torch.randn(x_shape, generator=gen, device=device)
w0 = torch.randn(w_shape, generator=gen, device=device)

# get a launchable
launchable = get_launchable(SimpleMatmul, (x0, w0), func_name="sample_boo_matmul", force_single_dispatch=True)
```
This PR adds a rough map of what happens below to `get_module_asm`.  May this be a guide to the next person spelunking through code gen!

### DIFF
--- a/iree/turbine/kernel/boo/runtime/launch.py
+++ b/iree/turbine/kernel/boo/runtime/launch.py
@@ -44,6 +44,46 @@ def get_module_asm(
     func_name: str = "main",
     force_single_dispatch: bool = False,
 ) -> str:
+    """Create MLIR assembly str from input torch.nn.Module.
+
+    Roughly speaking, for an input torch.nn.Module with a CustomOp,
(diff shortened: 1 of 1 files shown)


ftynse (approved):
More docs is always good!

PR #1001, opened 2025-07-02
[BOO] Basic GEMM benchmarking

AaronStGeorge:
Extends the pattern created in https://github.com/iree-org/iree-turbine/pull/992 to include GEMM benchmarking.

### DIFF
--- a/iree/turbine/kernel/boo/runtime/launch.py
+++ b/iree/turbine/kernel/boo/runtime/launch.py
@@ -225,11 +225,6 @@ def get_launchable(
     cache_only=False,
     force_single_dispatch=False,
 ) -> Launchable:
-    if cache_only and not is_cache_enabled():
-        raise ValueError(
-            "cache_only=True requires caching to be enabled. "
-            + "Please enable caching or set cache_only=False"
-        )
     session_cache_key = func_name + cache_only * "_no_jit"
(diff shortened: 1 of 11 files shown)


AaronStGeorge:
FYI: @ftynse here's my in progress GEMM benchmark harness if you're looking for another implementation during refactoring.

The code so far will need to be updated for API changes created in https://github.com/iree-org/iree-turbine/pull/1024 but other than that seems to be functionally complete though missing testing.

My current plan is to leave this as a draft and wait for https://github.com/iree-org/iree-turbine/pull/979 to land before final updates + tests + landing.

iree/turbine/kernel/boo/runtime/launch.py
```diff
@@ -225,6 +225,11 @@ def get_launchable(
     cache_only=False,
     force_single_dispatch=False,
 ) -> Launchable:
+    if cache_only and not is_cache_enabled():
```
zjgarvey, line 228: Would you mind putting this check after the cache retrieval attempt?

Even though this check is very quick, this function is getting called at runtime and we don't want to add any extra CPU overhead there.

zjgarvey (approved):
Just the one request, otherwise looks good to me.

PR #1002, opened 2025-07-02
[Wave] add optimization_level compile option

willghatch:
This gives the option a default level of 2, and disables optimization passes for optimization level 0.

### DIFF
--- a/iree/turbine/kernel/wave/compile_options.py
+++ b/iree/turbine/kernel/wave/compile_options.py
@@ -60,6 +60,7 @@ class WaveCompileOptions:
     use_water_leak_check: bool = False
 
     # === Performance options ===
+    optimization_level: bool = True
     denorm_fp_math_f32: str = None
     waves_per_eu: int = None
     wave_runtime: str = None
--- a/iree/turbine/kernel/wave/wave.py
+++ b/iree/turbine/kernel/wave/wave.py
@@ -569,13 +569,19 @@ class LaunchableWave(Launchable):
             trace, options, print_ir_before, print_ir_after
         )
 
-        # Optimizations.
         graph_passes += [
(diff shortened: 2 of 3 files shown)


willghatch:
A couple of questions for a reviewer:

1. I added use of O0 to a test, but I don't think it's a particularly good one.  I wanted one that would be more likely to trip if I disabled non-optimization passes, but the one that I found that seemed appropriate for that consideration also runs with benchmarks, etc, and I think it would be better to choose one that has no benchmarks, etc.  It works where it is, but I'm sure it increases compile times unnecessarily.
2. Do we use a standard code formatter, eg. black?
3. This is an initial guess at which passes are optimizations vs not.  One block was specifically labeled as optimizations, and I found some others that seemed like optimizations.  But I may be missing others.  If you know of others that are definitely just optimizations and not necessary lowerings, please let me know.

iree/turbine/kernel/wave/compile_options.py
```diff
@@ -58,6 +58,7 @@ class WaveCompileOptions:
     )
     use_local_scope: bool = False
     use_water_leak_check: bool = False
+    optimization_level: int = 2
```
ftynse, line 61: Nit 1: this look like it doesn't belong under the "debug options" comment below.
Nit 2: I'd expect the optimization level for a compiler to be 0. There is also no differentiation in the code between 1 and 2, all conditions are `> 0`, so unclear why 2 specifically is chosen.

harsh-nod: I agree, since we don't have a sense for what the different levels should be, maybe keeping it boolean makes sense. I think the canonicalize change might be a big one, so I am okay with splitting it up into 2 PRs, if that is the case.

Hardcode84 (approved):
LGTM, thanks

PR #1010, opened 2025-07-06
[Wave] Change paged decode attention stage 2 distribution.

Hardcode84:
Distribute across `head_size_kv` instead of `num_query_heads` for low `num_query_heads` to get threads better utilized.

Gave around 85->87 tok/s improvement on e2e bench 
```
NUM_Q_HEADS, NUM_KV_HEADS, HEAD_SIZE, HEAD_SIZE_KV, NUM_SEQS 
(6, 1, 128, 128, 1)
```

### DIFF
--- a/iree/turbine/kernel/wave/templates/paged_decode_attention.py
+++ b/iree/turbine/kernel/wave/templates/paged_decode_attention.py
@@ -217,7 +217,11 @@ def get_paged_decode_attention_kernels(
         vector_shapes = {
             S: 0,
             B: BLOCK_B // PHASE_1_BLOCK_B_WAVES,
-            N: BLOCK_N // PHASE_1_BLOCK_N_WAVES,
+            N: (
+                1
+                if phase_1_distribute_query_heads
+                else BLOCK_N // PHASE_1_BLOCK_N_WAVES
+            ),
             U: 1,
         }
         constraints += [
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/wave/templates/paged_decode_attention.py
```diff
@@ -190,16 +202,26 @@ def phase_0_constraints_mha():
...
 
     def phase_1_constraints() -> list[tkw.Constraint]:
-        constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(B, BLOCK_B, 0)]
+        constraints: list[tkw.Constraint] = []
+        if phase_1_distribute_query_heads:
+            constraints += [tkw.WorkgroupConstraint(B, BLOCK_B, 0)]
+            constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+        else:
+            constraints += [tkw.WorkgroupConstraint(B, BLOCK_B, 1)]
+            constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 0)]
```
harsh-nod, line 211: What if you left B as 0 and N as 1? Does that impact performance?

lit_tests/kernel/wave/attention/decode_attention.py
```diff
@@ -106,6 +106,77 @@ def test_paged_flash_decoding():
     # CHECK-COUNT-2:           arith.mulf
     # CHECK-COUNT-1:           arith.addf
     # CHECK-COUNT-1:      arith.divf
+    # Must be non-masked store.
```
harsh-nod, line 109: Why should this be a non-masked store? Is this part of the masking PR?

harsh-nod (changes requested):
Could add the performance change because of this PR and the shapes you tested it on? Otherwise, looks good overall.

harsh-nod (approved):
thanks!

PR #1024, opened 2025-07-08
[boo] refactor cache prepopulator towards generalization

ftynse:
Move the cache prepopulator to the generic driver directory and generalize it based on the previously introduce OpSignature and OpCLIParser classes in preparation for adding more ops.

Add/refine type annotations and fix typos while I'm here.

The shim in `conv_exports` is preserved because we don't have a parser dispatcher yet.


iree/turbine/kernel/boo/driver/preload.py
```diff
@@ -0,0 +1,301 @@
...
+            raise NotImplementedError(
+                "Downloading optimized kernels is not yet supported."
+            )
+        self.extra_compile_flags = extra_compile_flags
+        self.mlir_cache_status = {}
+        self.vmfb_cache_status = {}
+
+    def _assemble_signatures(self):
+        if self.commands_file:
+            self.commands += load_commands(self.commands_file, self.parser_cls)
```
zjgarvey, line 77: I think we should change `load_commands` so that it checks whether the first item in the line is one of any supported op (e.g., if it starts with `conv` or whatever else we support). That way, we don't need the Prepopulator class to be op-specific, but should infer which `OpCLIParser` to use based on what the first term is. 

Ideally, someone has a list of all the miopen commands they want to pre-load kernels for, and that could be a mix of convs and other ops. We don't want them to need to preload the cache separately for each op type.

iree/turbine/kernel/boo/driver/preload.py
```diff
@@ -0,0 +1,301 @@
...
+        torch_devices.append(torch.device(f"cuda:{i}"))
+    return torch_devices
+
+
+class CachePopulator:
+    def __init__(
+        self,
+        *,
+        parser_cls: type[OpCLIParser],
+        signature_cls: type[OpSignature],
```
zjgarvey, line 49: Maybe for later, but it might be a good idea to package the `OpCLIParser` together with it's corresponding `OpSignature`. E.g., `OpSignature` could have an abstract method for `get_miopen_parser` or `get_cli_parser`.

iree/turbine/kernel/boo/driver/preload.py
```diff
@@ -0,0 +1,301 @@
...
+    parser.add_argument(
+        "--max-processes",
+        "-j",
+        type=int,
+        help="Specify a maximum number of concurrent processes.",
+    )
+    return parser.parse_known_args()
+
+
+# TODO: we need a parser dispatcher to avoid having one preloader per op kind
```
zjgarvey, line 294: Ah, yeah. If you want to merge this as-is to unblock something, we can ignore my other comment about `load_commands` and file an issue for cleaning this up later.

zjgarvey (approved):
Nice. Thanks for updating the documentation as you go. 

I think this can be merged as-is if you have other work blocked on these changes. Otherwise it would be good to resolve the <OpName> dependence as soon as possible to avoid over-complicating the driver and preloader.

iree/turbine/kernel/boo/driver/preload.py
ftynse: I thought about that too, but couldn't decide. On one hand, it makes sense to connect that two, on the other hand, it feels sort of gross to have everything depend on `argparse` and have CLI arguments present deeper than the CLI driver. What's your take on this?

iree/turbine/kernel/boo/driver/preload.py
ftynse: We'd need a second op to properly exercise/test this. I'm about to bring that in with https://github.com/iree-org/iree-turbine/pull/979 so I'd rather land this first and generalize further after that lands if that's okay with you

PR #1036, opened 2025-07-08
[Wave] Reproducer for reported issue

harsh-nod:


harsh-nod:
@adedespirlet - does this PR look correct to you?

PR #1045, opened 2025-07-09
[Wave] change canonicalize default to true

willghatch:


harsh-nod:
I dont think it should make a difference for the e2e tests, since a canonicalize pass is run after we hand off to iree for compilation.

harsh-nod (approved):

Issue #1014, opened 2025-07-07
[BOO] Graph API backward failing when two identical matches differ in `requires_grad`


zjgarvey:
The issue is that when a model has two identical subgraphs, e.g.:

```(input, weight0) -> g -> output0 -> (output0, weight1) -> g -> output1```

where `input.requires_grad` is `False`, `weight0` and `weight1` both require grad, and `g` is a subgraph we are matching on with the replacement tool.

In this case, the AOT backward graph for the first `g` doesn't require gradient computed for the `0th` input, and this hard-codes behavior for the backward pass of our op replacement, which gets applied when substituting the second `g` in the graph.

I'm filing this issue early (before merging https://github.com/iree-org/iree-turbine/pull/967), since I'm writing a test which covers this bug to be addressed later.

PR #1046, opened 2025-07-10
[boo] Integrate boo fusion as a `torch.compile` backend

### LINKED ISSUE #1037: [BOO] Scope out a nestable `torch.compile` backend for graph API
Rather than using a one-off `torch.export.export` for getting a source graph for applying boo fusions, it would make more sense to implement the graph api through a `torch.compile` backend, which will update the graph when necessary, and re-apply the fusions.

Since users may want to apply an additional `torch.compile` backend on the remainder of the model that isn't handled by IREE, so we should test out passing an additional `torch.compile` backend to our backend, which will be applied after our graph modifications.

Note: Naively nesting `torch.compile` backends does not work, as the most recently applied dynamo context controls how the exported graph module gets jit compiled.

rkayaith:
This changes the `boo` fusion code to integrate a `torch.compile` backend that runs after AOTAutograd. This greatly simplifies things on our end, since pytorch now:
- handles capturing the forward and backwards passes, and provides these to us as separate graphs for compilation
- identifies values in the forwards pass that are required in the backwards pass, making them explicit outputs of the forward graph
- recompiles the graph as necessary, e.g. due to input shape changes

closes #1037, #1014

### DIFF
--- a/iree/turbine/dynamo/backends/boo.py
+++ b/iree/turbine/dynamo/backends/boo.py
@@ -12,21 +12,34 @@ from iree.turbine.kernel.boo.fusion.schema import (
 
 
 def backend(
+    *,
+    nested_backend: str | CompilerFn = "eager",
     fusion_schema: FusionSchema = DEFAULT_SUPPORTED_BOO_FUSIONS,
-    backend: str | CompilerFn = "eager",
 ):
-    backend_fn = make_boxed_compiler(torch._dynamo.lookup_backend(backend))
+    """
+    A 'torch.compile' backend that selectively offloads operations to IREE.
+    Alternatively, 'torch.compile(..., backend="iree_boo")' can be used.
+
+    After offloading is performed, the graph is further optimized using
+    'nested_backend'. Any valid 'torch.compile' backend can be specified.
+
+    'fusion_schema' may be used to control which operations are fused and
+    offloaded.
+    """
+    nested_backend_fn = make_boxed_compiler(
+        torch._dynamo.lookup_backend(nested_backend)
+    )
 
     def fw_compiler(model: fx.GraphModule, example_args):
         fusion_transform(model, fusion_schema=fusion_schema)
-        return backend_fn(model, example_args)
+        return nested_backend_fn(model, example_args)
 
     return aot_autograd(
         fw_compiler=fw_compiler,
         # Skip boo compilation for the backwards pass for now.
-        bw_compiler=backend_fn,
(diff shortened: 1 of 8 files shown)


iree/turbine/dynamo/backends/boo.py
```diff
@@ -0,0 +1,32 @@
...
+from iree.turbine.kernel.boo.fusion.apply import fusion_transform
+from iree.turbine.kernel.boo.fusion.schema import (
+    FusionSchema,
+    DEFAULT_SUPPORTED_BOO_FUSIONS,
+)
+
+
+def backend(
+    fusion_schema: FusionSchema = DEFAULT_SUPPORTED_BOO_FUSIONS,
+    backend: str | CompilerFn = "eager",
```
rkayaith, line 16: defaulting to `'eager'` for now, which does no further transformations. we could consider matching the `torch.compile` default of `'inductor'`

iree/turbine/kernel/boo/examples/subgraph_extraction.py
```diff
@@ -70,17 +70,17 @@ def profiler_ctx(enabled: bool = False) -> profile:
     # If any convolution is followed by a relu, it will be fused into the IREE's convolution kernel.
     # If any linear is preceeded or followed by view ops, those will be fused into the linear op.
     schema: FusionSchema = {
-        torch.ops.aten.conv2d.default: OpFusionSpec(
+        torch.ops.aten.convolution.default: OpFusionSpec(
             recursive=True, producers=(), consumers=(torch.ops.aten.relu.default,)
         ),
-        torch.ops.aten.linear.default: OpFusionSpec(
+        torch.ops.aten.addmm.default: OpFusionSpec(
```
rkayaith, line 76: the IR we get from `torch.compile` is slightly different, i believe this is decomposed more

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -166,58 +140,3 @@ def maybe_trim_none_outputs(gm: GraphModule) -> list[bool]:
     gm.graph.lint()
     gm.recompile()
     return none_output
-
-
-def get_autograd_function(
```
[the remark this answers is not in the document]
rkayaith: I think we should be pretty aggressive about removing unused code. Especially since there's a lot of tricky stuff going on here, when trying to understand the code it helps a lot if unused code doesn't distract from the relevant stuff.

PR #1048, opened 2025-07-10
[Wave] demo test of extract failing

willghatch:
@harsh-amd

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -1869,6 +1869,33 @@ def test_broadcast_add():
 
 
 @run_test
+def test_extract_register():
+    constraints: list[tkw.Constraint] = [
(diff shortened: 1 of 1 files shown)


lit_tests/kernel/wave/codegen.py
```diff
@@ -1868,6 +1868,33 @@ def broadcast_add(
...
+    constraints += [tkw.WaveConstraint(N, BLOCK_N / 2)]
+
+    @tkw.wave(constraints)
+    def extract_register(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+        b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+    ):
+        a_reg = tkw.read(a, elements_per_thread=4)
+        # unused result, but calling it crashes
+        extracted = tkw.extract(a_reg, 0)
```
harsh-nod, line 1888: This is a bug. So during the type inference phase, the index of each node has not yet been assigned and so infer_type should not be accessing the index of a node.

PR #1049, opened 2025-07-10
[boo] Always pass output shape to `generic_conv`

rkayaith:
This avoids having to do a try/catch to handle it not being specified.


zjgarvey (approved):
Yeah, sounds good.

zjgarvey:
@rkayaith this broke grouped convolutions since the output shape needs to be unflattened. Let me know if you want to address this yourself, otherwise I'll put up a PR later today.

rkayaith:
I'll take a look

Issue #1032, opened 2025-07-08
[BOO] Handle alternative memory formats for graph ops


zjgarvey:
We should have some default path for handling `torch.channels_last` and `torch.channels_last_3d` memory formats.

PR #1054, opened 2025-07-11
[BOO] Provide default handling of channels-last formats for graph ops

zjgarvey:
Addresses #1032 

This provides default handling of channels_last formats in the following way:

Suppose we have a graph like:

```graph = channels_last_input -> subgraph -> channels_last_output```

Then instead of importing and compiling `subgraph` directly, we factor:

```graph = channels_last_input -> permute -> contig_input -> new_subgraph -> contig_output -> inv_permute -> channels_last_output```

Where

```new_subgraph = contig_input -> inv_permute -> channels_last_input -> subgraph -> channels_last_output -> permute -> contig_output```

When we compile `new_subgraph` with IREE, it will include transposes into the IR. Since it is possible that we might fuse this transpose with the subgraph computation, it is strictly better than forcing a contiguous call in pytorch before launching the IREE kernel.

This PR also changes the hashing and logging of graph modules to include stride metadata, and makes `get_launchable` `spec_name` specialize to the presence of `cl` (channels_last) inputs.

### CODE REVIEW

File: iree/turbine/kernel/boo/ops/utils.py [hunk 1]
```diff
@@ -115,21 +116,92 @@ def get_func_name(
...
     shape_str = "x".join([str(dim) for dim in shape])
     return shape_str + f"x{dtype}"
 
 
+def _is_contiguous(
+    t: torch.Tensor | TensorMetadata, memory_format: torch.memory_format
+) -> bool:
+    if isinstance(t, torch.Tensor):
+        return t.is_contiguous(memory_format=memory_format)
+    metadata_memory_format = getattr(t, "memory_format", None)
```
rkayaith, line 130: why is `getattr` needed, shouldn't `TensorMetadata` always have this attribute?

I'm also curious why we need a case for `TensorMetadata` at all, isn't there always an example `torch.Tensor` value in the node metadata?

iree/turbine/kernel/boo/ops/utils.py
```diff
@@ -115,21 +116,92 @@ def get_func_name(
...
+    """
+    num_dims = num_dims or len(t.shape) - 2
+    cl_mem_format = CHANNELS_LAST_MEMORY_FORMAT.get(num_dims, None)
+    if cl_mem_format is None or not _is_contiguous(t, cl_mem_format):
+        return MemoryFormatInformation()
+
+    cl_contig = CHANNELS_LAST_TO_CONTIGUOUS_PERMUTATION.get(num_dims)
+    contig_cl = CONTIGUOUS_TO_CHANNELS_LAST_PERMUTATION.get(num_dims)
+    if None in [cl_contig, contig_cl]:
+        return MemoryFormatInformation()
```
rkayaith, line 168: I think it would simplify the callers if you just returned an identity permutation in these cases, so they can just unconditionally apply it (unless there's some overhead you're trying to avoid?).

Also since the permutations are always `None` if `is_channels_last` is `False`, it may make more sense for this to be refactored from `MemoryFormatInformation(bool, None | list, None | list)` => `MemoryFormatInformation(list, list) | None` (and if you're always returning a permutation, the `None` case can be removed)

I think there is also an assumption in these cases that if the tensor is not channels last, it _is_ contigous. I think it'd be good to have an explicit check + error for that.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -77,35 +82,108 @@ def _define_custom_graph_op(
...
+    output_mem_format_infos = [
+        get_memory_format_information(t) for t in outputs if t is not None
+    ]
+    logger.debug(
+        "Output TensorMetadata:\n%s\nOutput MemoryFormatInformation:\n%s",
+        str(outputs),
+        str(output_mem_format_infos),
+    )
+
+    class LayoutManagedModule(torch.nn.Module):
```
rkayaith, line 99: I don't think this is really specific to boo, and should probably be handled by the lower level export machinery. but this is good for now, I don't know if there's currently an a better place to do it.

iree/turbine/kernel/boo/ops/utils.py
zjgarvey: Yeah, I'm trying to avoid overhead by not calling `torch.permute` if it isn't necessary.

Agreed in that there is no point using the named tuple if we aren't storing any permutations.

It's not so much an assumption as "if it isn't channels_last, we are still going to need to pay the cost of a `tensor.contiguous()` call anyway, so don't complicate things by also doing metadata permutations". I think adding a warning in tensor import (device.py) makes the most sense, but let's put that in a different patch. 

In the future, we could take a look into inferring permutations for an arbitrarily strided tensor as long as it's not something complex like a proper subview. Or we could try to figure something out within IREE during the dlpack handoff to manage non-contiguous tensors.

iree/turbine/kernel/boo/ops/utils.py
zjgarvey: It's just a bit of extra saftey since I'm not explicitly checking `isinstance(t, TensorMetadata)` yet. I think it makes sense to refactor the error messaging and just call `t.memory_format`.

iree/turbine/kernel/boo/ops/graph.py
zjgarvey: Yeah, we can do that in a follow-up.

I think it will also make sense to infer the input layouts from the graph as well (instead of during runtime), since the graph will presumably recompile if we changed layouts at any point.

iree/turbine/kernel/boo/ops/utils.py
rkayaith: > It's not so much an assumption as "if it isn't channels_last, we are still going to need to pay the cost of a `tensor.contiguous()` call anyway, so don't complicate things by also doing metadata permutations".

On the input side that's true (though I think it'd be slightly better to put an explicit `contiguous()` call instead of relying on the one in the tensor import. maybe the `nested_compiler` can do something with it 🤷). But on the output side, returning a tensor with the wrong layout seems like a bug, as the consuming op may have been specialized to handle the layout from the metadata. So I think we _do_ need to actually error out in that case.

For now I would also error out on non-contiguous inputs, just because I'm curious when they show up, and that would surface things in an obvious way. But I don't feel strongly about that.

iree/turbine/kernel/boo/ops/utils.py
```diff
@@ -115,21 +116,81 @@ def get_func_name(
...
+def _is_contiguous(
+    t: torch.Tensor | TensorMetadata, memory_format: torch.memory_format
+) -> bool:
+    if isinstance(t, torch.Tensor):
+        return t.is_contiguous(memory_format=memory_format)
+    if isinstance(t, TensorMetadata):
+        return t.memory_format == memory_format
+    raise TypeError(
+        f"Unhandled type: {type(t)}. _is_contiguous input 0 must be a torch.Tensor or TensorMetadata object."
+    )
```
rkayaith, line 134: since `t` is typed as `torch.Tensor | TensorMetadata`, you don't really need to check the type here (the type checker should infer `t` is `TensorMetadata` here and be happy)
```suggestion
    return t.memory_format == memory_format
```

rkayaith (approved):
LGTM, just the last couple minor comments

iree/turbine/kernel/boo/ops/utils.py
zjgarvey: I'm fine with that, but the typing isn't a guarantee. In fact, mypy isn't even running in `iree/turbine/kernel` right now. 

This check won't happen at runtime, so I don't see any harm in keeping it.

iree/turbine/kernel/boo/ops/utils.py
zjgarvey: I added a check to raise an error if we know the output shape is going to be mis-handled, and a warning if the inputs are not contiguous which basically says "we are going to force this to be contiguous during the handoff to IREE".

Here is an interesting case to consider. The graph for the example in `examples/subgraph_extraction.py` is:

```
class GraphModule(torch.nn.Module):
    def forward(self, primals_1: "f32[16, 3, 3, 3][27, 9, 3, 1]", primals_2: "f32[16][1]", primals_3: "f32[4, 3, 32, 32][3072, 1024, 32, 1]", primals_4: "f32[3, 16, 1, 1][16, 1, 1, 1]", primals_5: "f32[3][1]", primals_6: "f32[1024, 900][900, 1]", primals_7: "f32[1024][1]"):
         # File: /home/zjgar/code/iree-turbine/iree/turbine/kernel/boo/examples/subgraph_extraction.py:39 in forward, code: x0 = self.layer0(x)
        convolution: "f32[4, 16, 30, 30][14400, 900, 30, 1]" = torch.ops.aten.convolution.default(primals_3, primals_1, primals_2, [1, 1], [0, 0], [1, 1], False, [0, 0], 1);  primals_2 = None
        relu: "f32[4, 16, 30, 30][14400, 900, 30, 1]" = torch.ops.aten.relu.default(convolution);  convolution = None
        
         # File: /home/zjgar/code/iree-turbine/iree/turbine/kernel/boo/examples/subgraph_extraction.py:40 in forward, code: x1 = self.layer1(x0)
        convolution_1: "f32[4, 3, 30, 30][2700, 900, 30, 1]" = torch.ops.aten.convolution.default(relu, primals_4, primals_5, [1, 1], [0, 0], [1, 1], False, [0, 0], 1);  primals_5 = None
        
         # File: /home/zjgar/code/iree-turbine/iree/turbine/kernel/boo/examples/subgraph_extraction.py:41 in forward, code: x2 = self.layer2(x1)
        view: "f32[4, 3, 900][2700, 900, 1]" = torch.ops.aten.view.default(convolution_1, [4, 3, 900]);  convolution_1 = None
        view_1: "f32[12, 900][900, 1]" = torch.ops.aten.view.default(view, [12, 900]);  view = None
        t: "f32[900, 1024][1, 900]" = torch.ops.aten.t.default(primals_6);  primals_6 = None
        addmm: "f32[12, 1024][1024, 1]" = torch.ops.aten.addmm.default(primals_7, view_1, t);  primals_7 = None
        view_2: "f32[4, 3, 1024][3072, 1024, 1]" = torch.ops.aten.view.default(addmm, [4, 3, 1024]);  addmm = None
        view_3: "f32[4, 3, 32, 32][3072, 1024, 32, 1]" = torch.ops.aten.view.default(view_2, [4, 3, 32, 32]);  view_2 = None
        return (view_3, primals_1, primals_3, primals_4, relu, view_1, t)
```

The input to `addmm` is not contiguous, so we should probably fuse in that transpose producer. However, we also need to return the result of that transpose, which feels a bit bad. On an unrelated note, but something I just noticed here, if we include view producers to `addmm`, this is indeed a situation where we are returning aliased inputs, so it's perhaps a good idea to not fuse those.

PR #1057, opened 2025-07-11
[Wave] Drop torch dependency from the `wave_runtime`

Hardcode84:
Pass stream from python.

Also, drop rocm dep while we at it.

### DIFF
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -196,9 +196,6 @@ jobs:
         if: "(!contains(toJSON(matrix.os), 'amdgpu') && contains(toJSON(matrix.os), 'mi300')) && !cancelled()"
         run: |
           python -m pip install --no-compile --upgrade pip
-          # Note: We install in three steps in order to satisfy requirements
-          # from non default locations first. Installing the PyTorch CPU
-          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
           pip install --no-compile -r pytorch-rocm-requirements.txt
           pip install --no-cache-dir -r requirements-iree-pinned.txt --upgrade
           pip install -r requirements.txt -e .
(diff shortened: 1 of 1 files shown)


Hardcode84:
Thanks for the review :), I'd like for @ScottTodd to take a look too

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -1,11 +1,75 @@
...
+    if (hipModuleLaunchKernel && hipGetErrorName && hipGetErrorString &&
+        hipModuleUnload && hipModuleLoad && hipModuleGetFunction)
+        return;
+
+#ifdef __linux__
+#define GET_FUNC(module, name) do { \
+    name = reinterpret_cast<decltype(name)>(dlsym(module, #name)); \
+    if (!name) { \
+        throw std::runtime_error("Failed to load symbol: " + std::string(#name)); \
+    } \
```
ScottTodd, line 53: Are C++ exceptions used in other parts of this code? Can this return a failure status instead?

iree/turbine/kernel/wave/runtime/runtime.cpp
```diff
@@ -1,11 +1,75 @@
...
+    GET_FUNC(module, hipGetErrorName);
+    GET_FUNC(module, hipGetErrorString);
+    GET_FUNC(module, hipModuleUnload);
+    GET_FUNC(module, hipModuleLoad);
+    GET_FUNC(module, hipModuleGetFunction);
+
+#undef GET_FUNC
+#else
+    #error "Unsupported platform"
+#endif
```
ScottTodd, line 71: Windows support wouldn't be too difficult to add here. See how https://github.com/iree-org/iree/blob/main/runtime/src/iree/hal/drivers/hip/dynamic_symbols.c searches for `amdhip64.dll` and uses this cross-platform dynamic library loader: https://github.com/iree-org/iree/blob/main/runtime/src/iree/base/internal/dynamic_library.h

Fine to start with only Linux matching the existing support, given how much of an improvement the rest of this PR is.

Hardcode84: Windows support here is trivial, yeah, the only thing which is stopping me is the we have no way to test it currently.

PR #1059, opened 2025-07-13
[NFC] Add logging

AaronStGeorge:
While working on a recent change I found the logging here very convenient, maybe it should be permanent?


ftynse (approved):
I was adding a breakpoint at this place to inspect the program, so looks helpful.

zjgarvey:
I do think this might be a bit aggressive logging for other use-cases of `aot.export` (e.g. full models with thousands of lines of IR), but I'm fine with this. 

If we want to move away from using `aot.export`/`CompiledModule` in `BOO`, I'm fine with that too. We are essentially only using the `torch.export`, `fx_importer`, and `CustomOpExpansionPass` pieces of this tech, so it might just make more sense to just use those directly and do the logging in `BOO`.

PR #1062, opened 2025-07-14
[Wave] Fix paged decode attention for weird head sizes

Hardcode84:
There were 2 issues:
* Expansion was generating less nodes than needed when dim size is not divisible by vector size
* Masking code wasn't generating properly for non-trivial vector sizes when corresponding dimension is not part of the constraints.


nithinsubbiah:
Decode attention e2e tests are failing - need to check if its because of the flakiness or the test needs an update

PR #1064, opened 2025-07-14
[BOO] Add more fusion tests

zjgarvey:
Adds two tests:

- Test for re-compile on new input shape.
- Test for boo fusion + inductor

Note: for re-compile to work properly, `dynamic=False` needs to be set, otherwise we need to properly handle dynamic shapes.

### DIFF
--- a/tests/kernel/boo/fusion/subgraph_fusion_test.py
+++ b/tests/kernel/boo/fusion/subgraph_fusion_test.py
@@ -6,6 +6,7 @@
 
 import unittest
 import tempfile
+import pytest
 
 from typing import Sequence
 from pathlib import Path
@@ -13,6 +14,7 @@ from pathlib import Path
 import torch
 from torch import fx
 from torch._dynamo.testing import EagerAndRecordGraphs
+from torch.profiler import profile, ProfilerActivity
 
 from iree.turbine.dynamo.backends import boo
 from iree.turbine.kernel.boo.fusion import OpFusionSpec, FusionSchema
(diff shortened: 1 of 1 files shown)


tests/kernel/boo/fusion/subgraph_fusion_test.py
```diff
@@ -266,6 +314,52 @@ def testRepeatedReplacementBackward(self):
...
+
+            @torch.compile(
+                dynamic=False,
+                backend=boo.backend(nested_backend="inductor", fusion_schema=schema),
+            )
+            def f(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+                x1 = mx(x)
+                y1 = my(y)
+                z = (x1 + y1 + 2.0) / 16.0
+                return z
```
AaronStGeorge, line 342: It might be convenient to have a backend for "apply all" boo fusion patterns then send the rest through inductor.

tests/kernel/boo/fusion/subgraph_fusion_test.py
```diff
@@ -266,6 +314,52 @@ def testRepeatedReplacementBackward(self):
 
             self.assertFalse(found_zero_grad, msg=err_msg)
 
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="Test requires GPU.")
```
rkayaith, line 317: do we have CI runs without a GPU?

tests/kernel/boo/fusion/subgraph_fusion_test.py
rkayaith: `compile(..., backend="iree_boo_inductor")` will do this (once we actually have default fusion patterns 😅)

rkayaith (approved):
nice, thanks!

PR #1065, opened 2025-07-14
[Wave] Include wave runtime in turbine requirements.txt

harsh-nod:
This PR includes the build of wave runtime in iree-turbine. With the recent removal of torch and hip deps in wave runtime, it became easier to build wave runtime as part of iree-turbine.

Here we move the setup.py from iree/turbine/kernel/wave/runtime into the top-level setup.py and update the pyproject.toml build deps accordingly.


setup.py
```diff
@@ -126,5 +162,7 @@ def initialize_options(self):
             f"parameterized{get_version_spec('parameterized')}",
         ],
     },
-    cmdclass={"build": BuildCommand},
+    cmdclass={"build": BuildCommand, "build_ext": CMakeBuild},
```
ScottTodd, line 165: Structurally this looks about right. Let's hold off on this until wave has its own separate `setup.py` though. That way, the iree-turbine package remains a pure Python source package and the new wave package can be a bdist wheel.

setup.py
```diff
@@ -8,13 +8,49 @@
...
+        ext_fullpath = Path.cwd() / self.get_ext_fullpath(ext.name)
+        extdir = ext_fullpath.parent.resolve()
+
+        # Configure CMake
+        cmake_args = [
+            f"-DPython_EXECUTABLE={sys.executable}",
+            f"-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={extdir}{os.sep}",
+            f"-DCMAKE_BUILD_TYPE={'Debug' if self.debug else 'Release'}",
+        ]
+        subprocess.check_call(["cmake", ext.sourcedir, *cmake_args], cwd=build_dir)
```
ScottTodd, line 48: We'll still also need to add a release build script and workflow that use manylinux. We can build within either:

* (upstream image with no additional deps) `quay.io/pypa/manylinux_2_28_x86_64`
* (iree-org image with a few extra deps) https://github.com/iree-org/base-docker-images/blob/main/dockerfiles/manylinux_x86_64.Dockerfile
* (nod-ai image with a few extra deps) https://github.com/nod-ai/base-docker-images/blob/main/dockerfiles/manylinux_x86_64.Dockerfile

A build script can look like https://github.com/nod-ai/shark-ai/blob/main/shortfin/build_tools/build_linux_package.sh and a workflow could look like https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/build_packages.yml.

The key lines in that workflow are:
* Running the build script: https://github.com/nod-ai/shark-ai/blob/74774d07800225b97ec1dd5eadd171a889b8beb3/.github/workflows/build_packages.yml#L181-L187
* Building across a matrix of operating systems and python versions: https://github.com/nod-ai/shark-ai/blob/74774d07800225b97ec1dd5eadd171a889b8beb3/.github/workflows/build_packages.yml#L93-L129 (we can limit to only Linux)
* Uploading wheels somewhere: https://github.com/nod-ai/shark-ai/blob/74774d07800225b97ec1dd5eadd171a889b8beb3/.github/workflows/build_packages.yml#L189-L195 . Note that the workflows in IREE have a job at the end that waits on prior jobs before sanity checking and promoting from a draft release to a prerelease, while the shark-ai workflow just pushes/publishes as wheels are built: https://github.com/iree-org/iree/blob/42df9b8d50e824cd83f9faa827dc67cf6e6c6eed/.github/workflows/build_package.yml#L366-L368

harsh-nod: Thanks, that's very helpful. I will add those 2 files to this PR for you to take a look at. Where should the wheel files be uploaded? I see actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2 but not sure where that SHA comes from?

ScottTodd: That `uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2` is using https://github.com/actions/upload-artifact, pinned to a commit hash. We pin all dependencies to commit hashes instead of tags (like `v4.6.2`) for supply chain security, since commit hashes are stable. Someone could redirect the `v4.6.2` tag to point at a different commit and then we would be running different code, like on https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised.

In iree-org/iree we upload one GitHub release per day: https://github.com/iree-org/iree/releases. In iree-org/iree-turbine and nod-ai/shark-ai we upload continuously to a fixed "dev wheels" GitHub release: https://github.com/iree-org/iree-turbine/releases/tag/dev-wheels. The separate releases are useful where they are lots of files per release but they are quite noisy in the releases view (drowning out stable releases without a filter like https://github.com/iree-org/iree/releases?q=prerelease%3Afalse). In both cases those are nightly/dev releases maintained with only weak guarantees about stability and retention (we could delete old releases after a few months, as we just did in shark-ai).

Nightly/dev release can be pip installed with a `--find-links` or `--extra-index-url` style argument (https://github.com/nod-ai/shark-ai/blob/main/docs/nightly_releases.md). As part of the stable release process (https://iree.dev/developers/general/release-management/), we "promote" a release candidate from the nightly/dev releases to stable and push to https://pypi.org/. Once pushed to pypi, users can install without any extra flags to `pip install`.

ScottTodd: > Also how can I test these?

* General (outdated) docs: https://iree.dev/developers/debugging/releases/
* Each `build_linux_package.sh` script is designed to be runnable locally and has detailed comments at the top for how to test. You can either run the script on its own to have it launch Docker or launch Docker yourself and run the commands that it would run. Either way, the output should be a whl file that you can `pip install` from outside of Docker on any machine.
* The `.yml` workflow files can be triggered with `workflow_dispatch` from a fork (so they don't "push to prod" in the common repo). See for example my testing at https://github.com/ScottTodd/shark-ai/actions/workflows/build_packages.yml?query=event%3Aworkflow_dispatch. In newer repositories like https://github.com/ROCm/TheRock, we've been configuring workflows to select between "dev" and "nightly" releases, allowing for testing using self-hosted github actions runners in the common repo, but that shouldn't be needed for these fast package builds that don't have unusual hardware requirements.

build_tools/build_linux_package.sh
```diff
@@ -0,0 +1,178 @@
+# Copyright 2025 The IREE Authors
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Modified version of https://github.com/nod-ai/shark-ai/blob/main/shortfin/build_tools/build_linux_package.sh
```
marbre, line 5: If this is forked from github.com/nod-ai/shark-ai we should retain the original file header.

build_tools/build_linux_package.sh
```diff
@@ -0,0 +1,178 @@
...
+REPO_ROOT="$(cd "$THIS_DIR"/../../ && pwd)"
+SCRIPT_NAME="$(basename $0)"
+ARCH="$(uname -m)"
+
+PYTHON_VERSIONS="${OVERRIDE_PYTHON_VERSIONS:-cp311-cp311 cp312-cp312 cp313-cp313}"
+OUTPUT_DIR="${OUTPUT_DIR:-${THIS_DIR}/wheelhouse}"
+CACHE_DIR="${CACHE_DIR:-}"
+
+if [[ "${ARCH}" == "x86_64" ]]; then
+  MANYLINUX_DOCKER_IMAGE="${MANYLINUX_DOCKER_IMAGE:-ghcr.io/nod-ai/manylinux_x86_64@sha256:4acf83343706d1e37252d6001ded3c97a73bc38620580f855b4e65e35ddc5681}"
```
marbre, line 42: In IREE-org, we shouldn't pull Docker images from nod-ai and use one of the [iree-org/base-docker-images](https://github.com/iree-org/base-docker-images).

build_tools/build_linux_package.sh
marbre: As Scott mentioned

> A build script can look like https://github.com/nod-ai/shark-ai/blob/main/shortfin/build_tools/build_linux_package.sh and a workflow could look like https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/build_packages.yml.

but you can also look into the build script used in IREE: https://github.com/iree-org/iree/blob/main/build_tools/python_deploy/build_linux_packages.sh

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools", "wheel", "cmake", "nanobind"]
```
marbre, line 2: These should be added to a `pyproject.toml` specific for the wave package, not to the one of iree-turbine.

setup.py
marbre: In addition to what Scott commented:
I think the above needs to be moved to a `setup.py` that is explicitly for wave and the changes should not be merged into the one for iree-turbine.

marbre (commented):
Some further comments. PR title will need an update as well as it is doesn't match the changes.

setup.py
harsh-nod: Yes makes sense. Thanks for the pointers. After the new repo is created, could we do a session together to make sure I have this all setup correctly for the new repo?

harsh-nod: I was able to run build_linux_package.sh, generate the wheels, install the wheels (in a new venv) and import wave_runtime succeeded.

harsh-nod:
Closing this PR as it has been migrated to the new repo.

PR #1066, opened 2025-07-15
[boo] deduplicate get_func_name for convolutions

ftynse:
Almost the same logic for generating function names was present in two different locations, creating room for divergence and bugs.


zjgarvey (changes requested):
I'm fine de-duplicating, but I want to get the func name without creating a `ConvSignature`. It takes many microseconds to create a signature and call `sig.get_func_name`, and anything within the custom operator function body is executed at runtime.

Let's preserve the caching `get_func_name` function, move it to `conv_exports`, and have `ConvSignature.get_func_name` call into the caching version.

PR #1069, opened 2025-07-15
[BOO] Allow inlining IR for layout customizable convolution

zjgarvey:
Boo ops now use `iree.turbine.runtime.op_reg.CustomOp` for library registration.

`layout_customizable_convolution` now has a `generate` method, which will allow inlining that IR for the fusion API. 

The next step is going to be replacing `aten::convolution` with a function which permutes channels_last inputs and applies layout customizable conv. These artificial permutations should ideally fold easily with the permutations introduced via `get_custom_graph_op`. Note: We need to manually do the replacement with `layout_customizable_convolution`, since once we are in mlir, we cannot specialize based on input layouts.

### DIFF
--- a/iree/turbine/kernel/boo/ops/layout_customizable_conv.py
+++ b/iree/turbine/kernel/boo/ops/layout_customizable_conv.py
@@ -20,7 +20,10 @@ from ..conv_exports import (
 from ..driver.launch import get_launchable
 from ..runtime import LaunchableRuntimeCache
 
-from ....runtime.op_reg import CustomOp, KernelBuilder, KernelSelection
+from ....runtime.op_reg import CustomOp, KernelBuilder, KernelSelection, impl_helper
+from ....support.logging import aot_logger as logger
+from ....transforms.merger import Merger
+from ....support.ir_imports import Operation
 
 __all__ = [
     "boo_layout_customizable_convolution",
@@ -66,8 +69,8 @@ class layout_customizable_convolution(CustomOp):
             groups=groups.v,
         )
         output_shape = self.conv_sig.output_shape
-        o_meta = self.conv_sig.output_perms.inv()(
-            torch.empty(tuple(output_shape), dtype=self.conv_sig.dtype, device="meta")
+        o_meta = torch.empty(
+            tuple(output_shape), dtype=self.conv_sig.dtype, device="meta"
         )
         ksel.return_tensor(o_meta)
 
@@ -77,12 +80,23 @@ class layout_customizable_convolution(CustomOp):
             sample_args = sample_args + (ksel.arg_descs[2].t,)
         func_name = self.conv_sig.get_func_name()
         module_op = generate_custom_op_compatible_ir(
-            self.conv_sig.get_nn_module(use_custom=True),
+            self.conv_sig.get_nn_module(use_custom=False),
             args=sample_args,
             func_name=func_name,
         )
-        # TODO: Merge this with kb and yield returns.
-        raise NotImplementedError("convolution generate NYI.")
+        # A bit silly, but we need to load the module_op into the kb.context.
+        module_op = Operation.parse(str(module_op), context=kb.context)
+        merger = Merger(
+            module_op, kb.module_body.owner, target_symbol_table=kb.symbol_table
+        )
+        merger.merge()
+        func_op = kb.symbol_table[merger.translate_symbol(func_name)]
+        kb.yield_results(
+            *impl_helper.call_function(
+                func_op,
+                *[binding for binding in kb.arg_bindings if binding is not None]
+            )
+        )
 
     def eager_execute(self, *args):
         return _boo_layout_customizable_convolution_impl(*args)
--- a/iree/turbine/kernel/boo/ops/layout_customizable_conv.py
+++ b/iree/turbine/kernel/boo/ops/layout_customizable_conv.py
@@ -20,7 +20,7 @@ from ..conv_exports import (
 from ..driver.launch import get_launchable
 from ..runtime import LaunchableRuntimeCache
 
-from ....runtime.op_reg import CustomOp
+from ....runtime.op_reg import CustomOp, KernelBuilder, KernelSelection
 
 __all__ = [
     "boo_layout_customizable_convolution",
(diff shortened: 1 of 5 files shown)


iree/turbine/kernel/boo/ops/layout_customizable_conv.py
```diff
@@ -8,31 +8,100 @@
...
+            self.conv_sig.get_nn_module(use_custom=True),
+            args=sample_args,
+            func_name=func_name,
+        )
+        # A bit silly, but we need to load the module_op into the kb.context.
+        module_op = Operation.parse(str(module_op), context=kb.context)
+        merger = Merger(
+            module_op, kb.module_body.owner, target_symbol_table=kb.symbol_table
+        )
+        merger.merge()
```
AaronStGeorge, line 92: If this is repeating pattern it might be nice to pass a module + context into export and build the IR in place. It's probably a lot of plumbing work (`CompiledModule` is a bit intimidating with all the metaclasses)

iree/turbine/kernel/boo/ops/layout_customizable_conv.py
```diff
@@ -8,31 +8,103 @@
...
 # Forward Convolution Implementations #
 
-define_schema(
-    "layout_customizable_convolution",
-    "(Tensor x, Tensor w, Tensor? b, int[] stride, int[] padding, int[] dilation, int groups, str input_layout, str kernel_layout, str output_layout) -> Tensor",
-)
+
+@CustomOp.register(library=BOO_LIBRARY)
+class layout_customizable_convolution(CustomOp):
+    signature = "layout_customizable_convolution(Tensor x, Tensor w, Tensor? b, int[] stride, int[] padding, int[] dilation, int groups, str input_layout, str kernel_layout, str output_layout) -> Tensor"
```
rkayaith, line 37: can we use the full names everywhere
```suggestion
    signature = "layout_customizable_convolution(Tensor input, Tensor weight, Tensor? bias, int[] stride, int[] padding, int[] dilation, int groups, str input_layout, str kernel_layout, str output_layout) -> Tensor"
```

tests/kernel/boo/ops/boo_conv_aot_test.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import unittest
```
rkayaith, line 7: don't we use pytest? we should probably stop adding more `unittest` style tests

tests/kernel/boo/ops/boo_conv_aot_test.py
```diff
@@ -0,0 +1,96 @@
...
+                [1, 1],
+                1,
+            )
+            * 0.1
+        )
+
+
+class SimpleAOTTest(unittest.TestCase):
+    def testAOTLayoutCustomizable(self):
+        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
```
rkayaith, line 46: this makes it hard to to test the cpu version on a system with a gpu, we should probably just parametrize the test instead (or only run on CPU, if the device doesn't actually matter)

rkayaith (approved):
Some minor comments, LGTM otherwise

iree/turbine/kernel/boo/ops/layout_customizable_conv.py
```diff
@@ -8,31 +8,103 @@
...
+        )
+        kb.yield_results(
+            *impl_helper.call_function(
+                func_op,
+                *[binding for binding in kb.arg_bindings if binding is not None]
+            )
+        )
+
+    def eager_execute(self, *args):
+        return _boo_layout_customizable_convolution_impl(*args)
```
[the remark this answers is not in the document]
zjgarvey: True. It does make testing these ops in isolation a bit more difficult if we can't just apply them to tensors directly, and considering this op started off as only supporting eager execution, I figured we keep it since all of our conv op tests rely on it.

tests/kernel/boo/ops/boo_conv_aot_test.py
zjgarvey: You can always use `CUDA_VISIBLE_DEVICES=-1`, but yeah, that makes sense.

tests/kernel/boo/ops/boo_conv_aot_test.py
zjgarvey: Yeah, I'm frankly just mindlessly "going with the flow" with tests in this repo that are probably old.

iree/turbine/kernel/boo/ops/layout_customizable_conv.py
```diff
@@ -8,31 +8,100 @@
...
+    def generate(self, ksel: KernelSelection, kb: KernelBuilder):
+        sample_args = (ksel.arg_descs[0].t, ksel.arg_descs[1].t)
+        if ksel.variant == "biased":
+            sample_args = sample_args + (ksel.arg_descs[2].t,)
+        func_name = self.conv_sig.get_func_name()
+        module_op = generate_custom_op_compatible_ir(
+            self.conv_sig.get_nn_module(use_custom=True),
+            args=sample_args,
+            func_name=func_name,
+        )
```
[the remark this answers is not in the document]
zjgarvey: I think it's because the module itself needs to be imported to have access to it's args. Otherwise it tries to create proxies for these, but can't figure it out.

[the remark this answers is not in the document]
rkayaith: I was able to get something like this working earlier, but working around `ConvSignature` + `get_nn_module` not being traceable was a headache. I think we should stick with what's here for now since it works.

Eventually if we rely on `torch.compile` for inserting boo convs and everything goes through `replacement_convolution`, we could get rid of `ConvSignature`/simplify it to the point its traceable, and then this becomes a lot easier.

rkayaith (approved):

PR #1072, opened 2025-07-16
[Wave] minor fixes to option handling

ashay:
This PR contains three small patches.

> [Wave] include canonicalization in the cache key
> 
> Prior to this patch, since the cache manager didn't include the
> canonicalization flag when computing the cache key, Wave occasionally
> retrieved an incorrect kernel from the cache.  Specifically, if the
> cached kernel was not run through canonicalization (since the flag is
> off by default) and if the programmer requested canonicalization, the
> programmer would still get the non-canonicalized (but cached) IR.
> 
> This patch fixes the problem by including canonicalization among the
> details that are included when computing the cache key.

and

> [Wave] support printing of cached IR
> 
> When Wave retrieves a cached kernel, we didn't check whether the
> programmer had requested that the IR be printed to the console,
> resulting in silent output despite the programmer's request to print the
> IR.  This patch introduces the check so that `print_mlir` has the
> desired effect, whether or not Wave is working with a cached or uncached
> kernel.

and

> [Wave] compute cache key using all options that may affect compilation
> 
> Several other options besides canonicalization may impact the generated
> IR and these were not included in the computation of the cache key,
> which would have the effect of retrieving an incorrect IR from the cache
> when these options changed from one compilation to the next.  This patch
> adds these new options to the function that computes the cache key.

### DIFF
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -185,6 +185,7 @@ class WaveCacheManager(object):
             options.dynamic_symbols,
             options.schedule,
             options.use_scheduling_barriers,
+            options.canonicalize,
         ]
 
         # Add kernel/helper function specific hashes.
--- a/iree/turbine/kernel/wave/cache.py
+++ b/iree/turbine/kernel/wave/cache.py
@@ -181,11 +181,33 @@ class WaveCacheManager(object):
         key = [
             arg_dtypes,
             processed_constraints,
+            # We include only those fields from `WaveCompileOptions` that may
+            # alter the generated IR.  We deliberately do not use the string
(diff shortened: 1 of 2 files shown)


tgymnich:
Could we include all options?

ashay:
> Could we include all options?

Thanks for catching that!  I went over all of the options in the `WaveCompileOptions` type and added the ones that (I think, to the best of my ability) may impact compilation.  Let me know if I should add others.  I refrained from using something like `str(options)` since the kernel hash (i.e. the computed cache key) is itself a member field. of this type.

ashay:
Fixed formatting so that the pre-commit hook passes.

ashay:
Closing this PR because I've migrated it to https://github.com/iree-org/wave/pull/6.

PR #1076, opened 2025-07-16
[boo] Replace aten.convolution with boo convolution

rkayaith:
This adds the functionality for doing `1-1` op replacements after subgraph fusion. By default we replace `aten.convolution` with our custom convolution op.

### DIFF
--- a/iree/turbine/kernel/boo/fusion/replacement.py
+++ b/iree/turbine/kernel/boo/fusion/replacement.py
@@ -1,0 +1,68 @@
+from collections.abc import Callable
+
+import torch
+from torch import fx
+from torch.fx.node import Argument, Target
+
+
+from iree.turbine.kernel.boo import ops as boo_ops
+
+# 1 to 1 op replacements. The replacement function should return the new target
+# and arguments, or 'None' if it doesn't apply.
+ReplacementSchema = dict[
+    Target,
+    Callable[
+        [tuple[Argument, ...], dict[str, object]],  # (node.args, node.meta)
+        tuple[Callable, tuple[Argument, ...]] | None,  # (new_target, new_args) | None
+    ],
+]
+
+
+def apply_replacements(graph: fx.Graph, replacements: ReplacementSchema):
+    for node in graph.nodes:
+        if node.op == "call_function":
+            replacer_fn = replacements.get(node.target, lambda *_: None)
+            replacement = replacer_fn(node.args, node.meta)
+            if replacement is None:
+                continue
+            target, target_args = replacement
+            with graph.inserting_after(node):
+                call_boo = graph.call_function(target, target_args)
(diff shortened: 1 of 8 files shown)


iree/turbine/kernel/boo/fusion/replacement.py
```diff
@@ -0,0 +1,68 @@
...
+
+# 1 to 1 op replacements. The replacement function should return the new target
+# and arguments, or 'None' if it doesn't apply.
+ReplacementSchema = dict[
+    Target,
+    Callable[
+        [tuple[Argument, ...], dict[str, object]],  # (node.args, node.meta)
+        tuple[Callable, tuple[Argument, ...]] | None,  # (new_target, new_args) | None
+    ],
+]
```
rkayaith, line 18: this is here instead of `schema.py` to avoid a cyclic dependency (`schema.py` imports `replace_aten_convolution` from here). this is a mismatch with the fusion schema, which _does_ live in `schema.py`.

I wonder if we really need `schema.py` at all? We could just move all the fusion-related stuff into `apply.py` and the replacement stuff into here.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -66,7 +66,10 @@ def get_custom_graph_op(
     """Converts a graph module into a custom operator."""
     gm_string = str(gm.print_readable(print_output=False, include_stride=True))
     hash = sha1(gm_string.encode(), usedforsecurity=False).hexdigest()
-    op_name = f"fused_op_{hash}"
+    call_function_names = "_".join(
+        n.name for n in gm.graph.nodes if n.op == "call_function"
+    )
+    op_name = f"fused_op_{call_function_names}_{hash}"
```
rkayaith, line 72: kind of hacky but this is for debugging/testing, so we can tell what the fused op is implementing

tests/kernel/boo/fusion/integration_test.py
```diff
@@ -0,0 +1,94 @@
...
+from iree.turbine.dynamo.backends import boo
+
+
+def test_custom_boo_conv_used():
+    """Test that we're using our custom convolution op"""
+    recorder = EagerAndRecordGraphs()
+    compiled_conv = torch.compile(
+        torch.ops.aten.convolution,
+        backend=boo.backend(nested_backend=recorder),
+    )
```
zjgarvey, line 23: Although I'm pretty sure pytest handles cleaning up files which get generated during the runs, I usually set up a temporary directory to set the cache to. This way, you can also inspect the IR etc. during the test.

zjgarvey: I'd recommend doing this just in case pytest doesn't automatically remove the cached files.

rkayaith: thanks!

PR #1078, opened 2025-07-17
[boo] introduce op registry

ftynse:
Introduce a registry class handling metadata classes for BOO ops. This automatically traverses the ops available in `op_exports` and registers them with a singleton registry class that can be later queried using the op "name". This name is supposed to be a unique string that cannot appear in the CLI syntax of another op. This is suboptimal and is difficult to ensure statically without a formal syntax, but will work as long as the syntax mirrors that of MIOpen driver.


iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -196,20 +196,25 @@ def _get_arg_parser() -> argparse.ArgumentParser:
         "--seed", type=int, required=False, help="Random number generator seed."
     )
     parser.add_argument("--device", type=int, default=0, help="The device to run on.")
+    parser.add_argument(
+        "--use-custom",
+        type=bool,
+        default=False,
```
rkayaith, line 202: we should default this to `True` everywhere I think

iree/turbine/kernel/boo/driver/preload.py
```diff
@@ -98,7 +93,7 @@ def run(
         self._assemble_signatures()
         logger.debug(
             "Pre-populating signatures: %s",
-            str(list([s._signature for s in self.signatures])),
+            str(list[self.signatures]),
```
rkayaith, line 96:
```suggestion
            str(list(self.signatures)),
```

iree/turbine/kernel/boo/driver/registry.py
```diff
@@ -0,0 +1,109 @@
...
+class BooOpRegistry:
+    """Bag-of-ops registry mapping a string key of the operation to
+    implementation classes and other metadata."""
+
+    _BOO_OP_REGISTRY: dict[str, _BooOpEntry] = {}
+
+    @staticmethod
+    def get_signature(key: str) -> type[OpSignature] | None:
+        """Get the signature class of the op identified by the given key."""
+        entry = __class__._BOO_OP_REGISTRY.get(key)
```
rkayaith, line 33: I've never seen `__class__` used like this before, that's interesting. If you use `@classmethod` it's a bit less magic, but I don't know if either is more idiomatic.


```suggestion
    @classmethod
    def get_signature(cls, key: str) -> type[OpSignature] | None:
        """Get the signature class of the op identified by the given key."""
        entry = cls._BOO_OP_REGISTRY.get(key)
```

iree/turbine/kernel/boo/driver/registry.py
```diff
@@ -0,0 +1,109 @@
...
+    def _register(
+        key: str, signature_cls: type[OpSignature], parser_cls: type[OpCLIParser]
+    ):
+        """Register the op."""
+        __class__._BOO_OP_REGISTRY[key] = _BooOpEntry(signature_cls, parser_cls)
+
+
+# On module load, traverse all non-private Python files under ../op_exports/,
+# import them as modules, and look for the BOO_OP_EXPORT tag to get registration
+# info.
```
rkayaith, line 90: Could we just explicitly register the exports, rather than have file traversal magic? It would be easier to understand, and I don't think we'll have enough different ops for it to be unwieldly.

zjgarvey: Perhaps we can add an `__init__.py` file to `op_exports`, and move `registry.py` to that directory. I think the following would work:

In `__init__.py`:

```python
# Put any new registered op files here.
from .conv import *
from .layer_norm import *

from registry import *
```

Each `opname.py` should get imported first (these will import from the parallel `registry.py` and register the op). Scoping between files is a bit tricky and I'm not sure how it works with class variables. I.e., I'm not sure if importing `BooOpRegistry` in these different files and calling the class methods will actually update the original or not. If this ends up being problematic, you could do something similar to `boo.runtime.LaunchableRuntimeCache` which uses static methods to access and modify a global variable (which would live in `registry.py`) instead of using a class like a global variable.

```python
from iree.turbine.kernel.boo import op_exports

registry = op_exports.BooOpRegistry # or BooOpRegistry.get() if using an approach analagous to LaunchableRuntimeCache
```

iree/turbine/kernel/boo/driver/registry.py
zjgarvey: I do believe `@classmethod` is the natural choice here, since we are accessing class variables.

zjgarvey (commented):
Overall I think this is a welcome restructuring. I have some suggestions about the setup for registration. Let me know what you think.

iree/turbine/kernel/boo/driver/registry.py
ftynse: Done.

iree/turbine/kernel/boo/driver/preload.py
ftynse: Done.

iree/turbine/kernel/boo/driver/registry.py
ftynse: I just went for explicit registration + README, this is the shortest and least magical.

ftynse:
PTAL, I'll move gemm separately so we don't have more merge conflicts

iree/turbine/kernel/boo/README.md
```diff
@@ -59,21 +62,95 @@ input_shape = ...
...
+
+y = conv_fwd(x, w)
+
+# get a random dLdy to back-prop.
+dLdy, _ = wrw_sig.get_sample_args(device=torch_device, seed=2)
+dLdw = conv_wrw(dLdy, x)
 ```
 
-### Use a `BooConv2d` module:
+### Replace `Conv2d` in a model with `BooConv2d`
```
rkayaith, line 117: This section was removed on main, looks like it was accidentally resurrected.

PR #1080, opened 2025-07-17
[BOO] Enable filtering based on functions provided to the fusion schema

zjgarvey:
Adds another item to `OpFusionSpec` : `match_filters`, where we can specify a sequence of `Callable[[Node], bool]` to determine whether to give nodes to IREE.

In particular, it will not start making a subgraph if the root node fails to match the filters, and the filters can be used to additionally restrict the nodes we fuse into the subgraph.

### DIFF
--- a/tests/kernel/boo/fusion/integration_test.py
+++ b/tests/kernel/boo/fusion/integration_test.py
@@ -27,6 +27,7 @@ def test_custom_boo_conv_used():
         recorder = EagerAndRecordGraphs()
         compiled_conv = torch.compile(
             torch.ops.aten.convolution,
+            dynamic=False,
             backend=boo.backend(nested_backend=recorder),
         )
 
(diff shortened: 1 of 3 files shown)


tests/kernel/boo/fusion/integration_test.py
```diff
@@ -27,6 +27,7 @@ def test_custom_boo_conv_used():
         recorder = EagerAndRecordGraphs()
         compiled_conv = torch.compile(
             torch.ops.aten.convolution,
+            dynamic=False,
```
rkayaith, line 30: why's `dynamic` needed here?

tests/kernel/boo/fusion/integration_test.py
```diff
@@ -59,6 +60,44 @@ def test_custom_boo_conv_used():
         assert call_node_target_str.startswith("boo.fused_op_convolution_replacement_")
 
 
+def test_filter_transpose_conv():
+    """Test that we're using our custom convolution op"""
```
rkayaith, line 64:
```suggestion
    """Test that we don't use our custom convolution op for transposed convs"""
```

tests/kernel/boo/fusion/integration_test.py
zjgarvey: torch.compile will notice we already compiled conv in a previous test, and will recompile based on the new inputs. If `dynamic` isn't false, it adds symbolic ints to the graph, which we don't properly handle yet.

rkayaith: sounds like calling `torch.compiler.reset()` would avoid that, we should get pytest to call that automatically between tests. I can do that when I have some time.

PR #1085, opened 2025-07-22
[BOO] Update README with torch.compile information

zjgarvey:
This adds information about the boo `torch.compile` backends and removes information about the convolution-specific replacement utility.

### DIFF
--- a/iree/turbine/kernel/boo/README.md
+++ b/iree/turbine/kernel/boo/README.md
@@ -34,17 +34,14 @@ boo_model = torch.compile(model, dynamic=False, backend="iree_boo")
 boo_inductor_model = torch.compile(model, dynamic=False, backend="iree_boo_inductor")
 ```
 
-To customize the backend further, the boo backend can be called directly.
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/boo/README.md
```diff
@@ -1,21 +1,57 @@
...
+
+```python
+import torch
+
+import iree.turbine.dynamo.backends.boo as boo
+from iree.turbine.kernel.boo.fusion import FusionSchema, ReplacementSchema
+
+custom_fusion_schema : FusionSchema = ...
+custom_replacement_schema : ReplacementSchema = ...
+customized_backend = boo.backend(nested_backend="some_other_backend", fusion_schema=custom_fusion_schema, post_fusion_replacements=custom_replacement_schema)
```
rkayaith, line 47: I think we might want to avoid advertising the fusion/replacement schema for now, and leave it as more of an internal detail used for our own experimentation/testing. The specifics here could change, and it's probably more advanced than actual users would be interested in.

rkayaith (approved):

PR #1087, opened 2025-07-22
[boo][driver] Replace Tracy with torch profiler

rkayaith:
Under the hood `torch.profiler` is using ROCTracer to collect timing information. Removing the tracy dependency greatly simplifies IREE python bindings setup, and avoids various workarounds we've needed.

ROCTracer consistently reports execution times that are _shorter_ than tracy; on average by 5 - 10 microseconds.

Note that you'll likely see warnings like:
```
[W722 11:30:12.244194100 collection.cpp:1085] Warning: ROCTracer produced duplicate flow start: 5 (function operator())
```
which seems to come from some issue with the pytorch+roctracer integration: https://github.com/ROCm/roctracer/issues/111


zjgarvey (approved):
Nice! LGTM

ftynse (approved):
Nice, much simpler and likely consistent with what external folks would use anyway!

PR #1091, opened 2025-07-23
[BOO] Update examples and test BatchNorm2d

zjgarvey:
This PR:

1. Enables using "iree_boo" backend on resnet 18 training example.
2. Fixes an issue with multi-output producer fusions. Testing resnet 18 with conv + batch norm fusion, I noticed `getitem` ops weren't getting fused into the graph and we were failing to handle the multi-output result.
3. Fixes an issue with op names getting too long for the OS to handle.
4. Adds tests for the path names and for `BatchNorm2d` with backward.

### DIFF
--- a/iree/turbine/kernel/boo/examples/resnet_18_backward.py
+++ b/iree/turbine/kernel/boo/examples/resnet_18_backward.py
@@ -29,7 +29,11 @@ from iree.turbine.kernel.boo.modeling import replace_conv2d_with_boo_conv
 
 
 def get_model(
-    use_boo=True, compile=False, memory_format=torch.channels_last, **boo_conv_kwargs
+    use_direct_replacement=False,
+    compile=False,
+    compile_backend="iree_boo",
+    memory_format=torch.channels_last,
+    **boo_conv_kwargs,
 ):
     """Sets up a resnet 18 model on cuda device based on provided parameters."""
     # base model
@@ -37,13 +41,18 @@ def get_model(
     # boo replacement
     resnet_model = (
         replace_conv2d_with_boo_conv(resnet_model, **boo_conv_kwargs)
-        if use_boo
+        if use_direct_replacement
         else resnet_model
     )
     # move to gpu and apply memory format
     resnet_model = resnet_model.to(device="cuda", memory_format=memory_format)
     # compile model
-    resnet_model = torch.compile(resnet_model) if compile else resnet_model
+    resnet_model = (
+        torch.compile(resnet_model, dynamic=False, backend=compile_backend)
+        if compile
+        else resnet_model
+    )
+    resnet_model.train()
     return resnet_model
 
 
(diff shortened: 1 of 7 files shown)


iree/turbine/kernel/boo/examples/resnet_18_backward.py
```diff
@@ -29,21 +29,30 @@
...
     # move to gpu and apply memory format
     resnet_model = resnet_model.to(device="cuda", memory_format=memory_format)
     # compile model
-    resnet_model = torch.compile(resnet_model) if compile else resnet_model
+    resnet_model = (
+        torch.compile(resnet_model, dynamic=False, backend=compile_backend)
+        if compile
+        else resnet_model
+    )
+    resnet_model.train()
```
ftynse, line 55: Is this needed to enable batchnorm? If so, worth an explanatory comment.

iree/turbine/kernel/boo/examples/resnet_18_backward.py
```diff
@@ -29,21 +29,30 @@
 
 
 def get_model(
-    use_boo=True, compile=False, memory_format=torch.channels_last, **boo_conv_kwargs
+    use_direct_replacement=False,
+    compile=False,
+    compile_backend="iree_boo",
+    memory_format=torch.channels_last,
```
ftynse, line 35:
```suggestion
    use_direct_replacement: bool = False,
    compile: bool = False,
    compile_backend: str = "iree_boo",
    memory_format: torch.memory_format = torch.channels_last,
```

tests/kernel/boo/ops/graph_ops_test.py
```diff
@@ -28,11 +28,48 @@ def forward(self, x):
...
 class GraphOpsTest(unittest.TestCase):
     def setUp(self):
         LaunchableRuntimeCache.clear()
         LaunchableRuntimeCache.set_cache_limit(0)
 
+    def testLongPathName(self):
+        with tempfile.TemporaryDirectory() as td:
+            set_cache_dir(Path(td))
+
+            m = LongFusionSample(16, 32, 10)
```
ftynse, line 59: Do we also want to test that all trailing elementwise operations got fused? Otherwise why bother with a long module

ftynse (approved):

iree/turbine/kernel/boo/examples/resnet_18_backward.py
```diff
@@ -29,21 +29,30 @@
...
 ):
     """Sets up a resnet 18 model on cuda device based on provided parameters."""
     # base model
     resnet_model = torchvision.models.resnet18(pretrained=False)
     # boo replacement
     resnet_model = (
         replace_conv2d_with_boo_conv(resnet_model, **boo_conv_kwargs)
-        if use_boo
+        if use_direct_replacement
         else resnet_model
```
rkayaith, line 45: should we just remove `replace_conv2d_with_boo_conv` from the example, and always `torch.compile`

zjgarvey: I'm going to leave it in for now, since the behavior might be different and I'd like to be able to compare the two replacement approaches.

iree/turbine/kernel/boo/examples/resnet_18_backward.py
zjgarvey: Yeah, the `train()` is to be able to safely assume we are getting a certain BatchNorm op. I'll add a comment.

tests/kernel/boo/ops/graph_ops_test.py
zjgarvey: The graph module is being directly exported and into a single op (I'm not forcing a single dispatch, just making sure the func name is correct).

PR #1095, opened 2025-07-24
[boo] fold ConvSignatureStorage into ConvSignature

ftynse:
The separation was necessary primarily for serialization purposes and there are now other mechanisms to support that. This also improves property discovery by no longer proxying to the storage object via `__getattr__` that was confusing the language server.


zjgarvey (approved):
Thanks, this is a very welcome change.

How do you feel about removing mode as an attribute of the conv signature, and instead have `Signature.get_nn_module` and the op-specific `get_launchable` take a mode argument. This way we won't need `as_init_kwargs`?

PR #1101, opened 2025-07-29
[boo] Add a pytest fixture for enabling the file cache

rkayaith:
This adds a `boo_cache_dir` pytest fixture, which enables the file cache in a test-local directory for the duration of the test, and returns the path so that the test can inspect the contents. This avoids each test needing to set up its own temporary directory. A following change will turn off the file cache by default, so only tests that explicitly request `boo_cache_dir` (or turn on the cache internally) will have file caching enabled.

A change in behaviour here is that the cache is left intact after test execution finishes, which allows for easier debugging. By default pytest dumps test files under `/tmp/pytest-of-$USER/pytest-current/`

This also adds an `autouse` fixture which sets the BOO cache directory to a test-local directory, in order to isolate tests from each other.

### DIFF
--- a/tests/kernel/boo/conftest.py
+++ b/tests/kernel/boo/conftest.py
@@ -4,7 +4,7 @@ from iree.turbine.kernel.boo.runtime import set_cache_dir, use_cache_dir
 
 
 @pytest.fixture(autouse=True)
-def _isolate_tests(tmp_path):
+def _isolate_tests(tmp_path: Path):
     # Ensure each test has its own cache directory, in case the cache is enabled.
     set_cache_dir(tmp_path / ".cache" / "turbine_kernels" / "boo")
 
--- a/tests/kernel/boo/conftest.py
+++ b/tests/kernel/boo/conftest.py
@@ -1,0 +1,16 @@
+from pathlib import Path
+import pytest
+from iree.turbine.kernel.boo.runtime import set_cache_dir, use_cache_dir
+
+
+@pytest.fixture(autouse=True)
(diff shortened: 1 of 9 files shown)


tests/kernel/boo/conftest.py
```diff
@@ -0,0 +1,16 @@
+from pathlib import Path
+import pytest
+from iree.turbine.kernel.boo.runtime import set_cache_dir, use_cache_dir
+
+
+@pytest.fixture(autouse=True)
+def _isolate_tests(tmp_path):
```
zjgarvey, line 7: nit (unless the pytest `tmp_path` doesn't allow it):
```suggestion
def _isolate_tests(tmp_path: Path):
```

Side note, pretty wild that pytest knows that an arg named `tmp_path` means make a test-specific temporary path (directory?).

zjgarvey (approved):
Nice, I like that this resolves some annoying issues with `tempfile` re-using old temp dirs.

tests/kernel/boo/conftest.py
rkayaith: yea these fixtures are magic, they're useful but also scary😂

tests/kernel/boo/conftest.py
```diff
@@ -0,0 +1,16 @@
+from pathlib import Path
```
ftynse, line 1: License header missing

ftynse (approved):
Nice! I didn't look deep into parts that look mechanical, let me know if you want me to

PR #1104, opened 2025-07-30
[RFC][BOO] Modify input backward conv IR generation

zjgarvey:
For strides != 1, we need to spread out the `grad_output` tensor by the strides. This is taking a first stab at getting something to compile, but doesn't work well yet when the offsets are non-zero, since I'm hacking through a bit of a gross generic op in that case. 

It might be better to pre-pad or something instead for these cases, but need to test it out a bit.

This also demos using mlir bindings to define a custom op instead of the jinja templates. This might be worth using for the `generic_conv` `CustomOp` which is a bit hacky in it's template use.


zjgarvey:
Closing since we are likely going to try and resolve this in the compiler.

PR #1105, opened 2025-07-31
[boo] Automatically reset caches between tests

rkayaith:
This avoids interference between tests.

### DIFF
--- a/iree/turbine/kernel/boo/runtime/cache.py
+++ b/iree/turbine/kernel/boo/runtime/cache.py
@@ -118,11 +118,13 @@ class LaunchableRuntimeCache:
         return _launchable_cache
 
     @staticmethod
-    def clear():
-        global _launchable_cache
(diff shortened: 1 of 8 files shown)


iree/turbine/kernel/boo/runtime/cache.py
```diff
@@ -118,11 +118,13 @@ def get_launchable_cache():
         return _launchable_cache
 
     @staticmethod
-    def clear():
-        global _launchable_cache
-        if not "_launchable_cache" in globals():
-            return
-        _launchable_cache.session_cache.clear()
+    def reset():
```
rkayaith, line 121: the main difference here is that deleting the whole cache object will also reset any options that were set, i.e. the cache limit.

zjgarvey (commented):
Some minor comments, but I feel like it would be good to address/discuss them before approval.

zjgarvey (approved):

PR #1106, opened 2025-07-31
[boo] Disable file caching by default

rkayaith:
The file cache has limited usefulness outside of testing/debugging, and is error-prone for users/developers as we don't have a good system for automatically invalidating it.


zjgarvey (approved):
Yeah, I think this is a good idea. For the future, how do you feel about removing the environment variable `BOO_CACHE_ON` entirely, and only enable the cache if `BOO_CACHE_DIR` is set?

PR #1108, opened 2025-07-31
[boo] add a combined backward layer norm

ftynse:
Add an op export for combined computaiton of all gradients in layer norm. This may be more efficient than executing them one by one in some cases and requires separate testing.

### DIFF
--- a/iree/turbine/kernel/boo/op_exports/layer_norm.py
+++ b/iree/turbine/kernel/boo/op_exports/layer_norm.py
@@ -22,6 +22,7 @@ class Mode(ModeBase, IntEnum):
     INPUT_BACKWARD = 1
     WEIGHT_BACKWARD = 2
     BIAS_BACKWARD = 3
+    FULL_BACKWARD = 4
 
 
 class LayerNormSignature(OpSignature):
@@ -175,6 +176,7 @@ class LayerNormSignature(OpSignature):
         input = forward_args[0]
         # TODO: is this possible at this level?
         weight = forward_args[1] if len(forward_args) > 1 else None
+        bias = forward_args[2] if len(forward_args) > 2 else None
         _, mean, rstd = forward_results
         if self.mode == Mode.INPUT_BACKWARD:
             return (input, weight, mean, rstd)
@@ -182,6 +184,8 @@ class LayerNormSignature(OpSignature):
             return (input, mean, rstd)
         if self.mode == Mode.BIAS_BACKWARD:
             return ()
+        if self.mode == Mode.FULL_BACKWARD:
+            return (input, weight, bias, mean, rstd)
         assert False, "Unsupported mode."
 
     def as_init_kwargs(self) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


iree/turbine/kernel/boo/op_exports/layer_norm.py
```diff
@@ -419,6 +498,8 @@ def get_signature(args: argparse.Namespace) -> LayerNormSignature:
                 mode = Mode.WEIGHT_BACKWARD
             case 4:
                 mode = Mode.BIAS_BACKWARD
+            case 5:
+                mode = Mode.FULL_BACKWARD
```
rkayaith, line 502: do these match the MIOpen interface? for conv at least, `forw` is actually a bitset (`1 << 0` is forward, `1 << 1` is input backward, `1 << 2` is weight backward, so all backwards would be `0b110 == 6`)

iree/turbine/kernel/boo/op_exports/layer_norm.py
```diff
@@ -374,6 +390,69 @@ def forward(self, grad_output: torch.Tensor) -> torch.Tensor:
...
+        rstd: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        assert self.need_weight != (
+            weight is None
+        ), "Weight must be provided if its gradient is requested."
+        assert self.need_bias != (
+            bias is None
+        ), "Bias must be provided if its gradient is requested."
+        if self.use_aten:
+            return torch.ops.aten.native_layer_norm_backward(
```
rkayaith, line 427: In the `torch.compile` integration I assume `torch.ops.aten.native_layer_norm_backward` is what we'll actually see. Is there any reason to _not_ use it for all cases? I'm a bit concerned that anything else won't accurately reflect the actual model performance.

ftynse: `use_aten` defaults to `True`, so we will use this. The override is needed because apparently integrations may have a flag to use layernorm as rmsnorm, which does less computation and needs to be reflected in the impl.

ftynse: > right, can the aten op be used for the other backwards variations too?

good point, yes we could use that too

> I don't really follow how this would work, how do you see the use_aten option being used? Is it intended to be plumbed through to the driver at some point? 

So far I was just changing the default and rerunning the benchmark script to generate a second CSV table and then align columns to get aten and non-aten results side-by-side, but we can expose this through the driver indeed.

> As for the implementation here, what would we see when going through torch.compile? I assume it'll be some combination of aten ops, could we use that as the implementation instead?

This is part of core aten IR according to https://docs.pytorch.org/docs/stable/torch.compiler_ir.html, so we should see it as is. Somewhere in the pipeline torch-mlir would expand it to the ops similar from the manual implementation below.

> (let me know if you'd like to have a quick chat about this, there may be some context I'm missing)

Time zones are hard... The context is just me pushing out the bits I needed to get layernorm numbers for the sizes of reference and discovering fun edge cases. Your comments make sense to me though, I just haven't thought about all the things :)

rkayaith (approved):

PR #1109, opened 2025-08-01
[BOO] Add some driver report improvements

zjgarvey:
This PR:

- Prints pytorch kernel launches in addition to iree for reference.
- Computes theoretical total op gpu min time from the per-dispatch stats if multiple dispatches are found.

We will likely need to relax "make-single-dispatch" attribute, and this will allow us to record some value for those cases.

### DIFF
--- a/iree/turbine/kernel/boo/driver/driver.py
+++ b/iree/turbine/kernel/boo/driver/driver.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from contextlib import nullcontext
 import gc
 import argparse
-from typing import Callable, Sequence, Literal
+from typing import Callable, Sequence, NamedTuple
 import os
 import shlex
 import statistics
@@ -22,7 +22,18 @@ from iree.turbine.kernel.boo.driver.launch import get_launchable
 from iree.turbine.kernel.boo.op_exports.registry import BooOpRegistry
 
 ZoneData = dict[str, list[float]]
-ZoneStats = dict[Literal["min", "max", "mean", "stddev", "count"], float | int]
+
+ALL_STATS = ["min", "max", "mean", "stddev", "num_dispatches"]
+
+
+class ZoneStats(NamedTuple):
+    min: float | str = "N.A."
+    max: float | str = "N.A."
+    mean: float | str = "N.A."
+    stddev: float | str = "N.A."
+    num_dispatches: int | str = "N.A."
+
+
 ZoneStatsSummary = dict[str, ZoneStats]
 
 
@@ -46,8 +57,10 @@ command-line arguments are appended to the arguments from the file.
         "--reference-backend",
         type=str,
         choices=["torch", "torch-compile"],
-        default=None,
+        action="append",
+        default=[],
         required=False,
+        help="Choose reference backends to compare performance against.",
     )
     parser.add_argument(
         "--csv",
@@ -67,7 +80,7 @@ command-line arguments are appended to the arguments from the file.
     parser.add_argument(
         "--splat-input-value",
         default=None,
-        type=int,
+        type=float,
         help="Use a splat value for inputs (defaults to random values).",
     )
     return parser
@@ -98,19 +111,22 @@ def main():
         mio_args = [extra_cli_args]  # use CLI arguments
 
     # Check the reference backend
-    ref_backend = meta_args.reference_backend
+    ref_backends = meta_args.reference_backend
     # TODO: Add ability to benchmark against torch-compile (inductor).
-    if ref_backend == "torch-compile":
+    if "torch-compile" in ref_backends:
         raise NotImplementedError(
             "Comparing against torch-compiled reference not yet implemented."
         )
 
     # Setup a csv output file with headers.
+    csv_stats = ALL_STATS
+    backends = ["iree_boo"] + ref_backends
     csv_file = open(meta_args.csv if meta_args.csv is not None else os.devnull, "w")
-    csv_headers = ["arguments", "iree min time (us)"]
-    if ref_backend is not None:
-        csv_headers.append(f"{ref_backend} min time (us)")
-    csv_file.write(",".join(csv_headers))
+    csv_headers = ["arguments"]
+    for b in backends:
+        for stat in csv_stats:
+            csv_headers.extend([f"{b} {stat}"])
+    csv_file.write(",".join(csv_headers) + "\n")
 
     timing_parser = _get_timing_parser()
 
@@ -131,9 +147,6 @@ def main():
         sample_inputs = _get_sample_args(
             signature, meta_args.splat_input_value, devices
         )
-        launchable = get_launchable(signature)
-        for device in devices:
-            launchable.preload(device)
         output_num_bytes = signature.get_output_size()
 
         profile_context = (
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -62,50 +62,115 @@ def main():
...
-            if len(dispatch_zone_names) > 1:
-                s += f"\t({zone_name})"
-            print(s)
 
-        if len(dispatch_zone_names) == 0:
-            csv_file.write("no timing info\n")
-        elif len(dispatch_zone_names) > 1:
-            csv_file.write("multiple dispatches\n")
+        # get iree stats and print
+        iree_filter: Callable[[str], bool] = lambda name: name.startswith(func_name)
```
rkayaith, line 100: I noticed a lot of GPU entries were just from generating the random inputs. if we hoist that out of `run()`, do we still need this filtering?

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -62,50 +62,115 @@ def main():
...
+def get_theoretical_min_time(results: ZoneStatsSummary) -> float:
+    """Computes a theoretical min time by summing the min time from each zone with multiplicity."""
+    total = 0.0
+    count: None | int = None
+    for stats in results.values():
+        if count is None:
+            count = int(stats["count"])
         else:
-            csv_file.write(f"{min(zones[dispatch_zone_names[0]]):.2f}\n")
+            count = min(count, int(stats["count"]))
```
rkayaith, line 129: don't we expect all zones to have the same count?

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -62,50 +62,115 @@ def main():
...
 
     runner_parser = argparse.ArgumentParser()
     runner_parser.add_argument("--time", "-t", type=int, help="Enable timing")
+
     for file_args in mio_file_args:
         driver_args = file_args + extra_cli_args
         timing_args, runner_args = runner_parser.parse_known_args(driver_args)
-        func = lambda: run(runner_args, args.gpu_id)
+        iree_func = lambda: run(runner_args, args.gpu_id, use_iree=True)
+        ref_func = lambda: run(runner_args, args.gpu_id, use_iree=False)
```
rkayaith, line 88: I think it would be better to only run one these at a time, or have flags to specify which backends to run. We probably want to add other pytorch backends as reference (e.g. comparing to inductor is probably more fair to pytorch), and when collecting IREE reference numbers for a big batch we don't really want to waste time measuring pytorch.

iree/turbine/kernel/boo/driver/driver.py
rkayaith: could we plumb the iteration count into here and assert/warn if the count isn't a multiple? just want to be extra careful we're measuring what we intended

iree/turbine/kernel/boo/driver/driver.py
rkayaith: I think it'd be good to do. and if you have a mode that runs pytorch-only without IREE, the current approach won't work since you won't have `iree_zones` available to do the filtering

zjgarvey:
@rkayaith Sorry for the significant code change. While I was hoisting the example input generation and adding another command line arg, I ended up deciding to refactor a bit.

iree/turbine/kernel/boo/driver/driver.py
zjgarvey: Yeah I was filtering out the zones which are shared between iree and pytorch (i.e. input gen) but that's maybe a bit unreliable, so I hoisted.

iree/turbine/kernel/boo/driver/driver.py
zjgarvey: Yeah, that's a good point. I'll add something to make sure we are measuring theoretical total min time correctly.

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -177,22 +237,86 @@ def run(cli_args: Sequence[str], gpu_id: int):
...
+        for device in devices
+    ]
+    return per_device_data
+
+
+def _get_theoretical_min_time(results: ZoneStatsSummary, iter: int) -> float:
+    """Computes a theoretical min time by summing the min time from each zone with multiplicity."""
+    total = 0.0
+    for stats in results.values():
+        total += stats["min"] * (stats["count"] // iter)
```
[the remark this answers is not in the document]
zjgarvey: I'm also suspicious that perhaps the profiler has some maximum allowed number of events unless an option is passed. I'll take a look into this.

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -51,117 +64,163 @@ def main():
...
+        print(f">>> Per-launch GPU min time (IREE): {iree_theoretical_min:.2f}us")
+        csv_file.write(f"{iree_theoretical_min:.2f}")
+
+        if not ref_backend:
+            csv_file.write("\n")
+            continue
+
+        ref_func = signature.get_nn_module(use_custom=False)
+
+        ### Reference run ###
```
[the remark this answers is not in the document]
zjgarvey: Yeah, I'll see if I can clean it up a bit.

iree/turbine/kernel/boo/driver/driver.py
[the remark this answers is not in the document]
rkayaith: yeah I was playing around with alternatives earlier, I tried using `cuda.Event`s like here: https://github.com/pytorch/pytorch/blob/dcc36e38bb1273d034de2ae25b5900cccb65e8a4/torch/_inductor/runtime/benchmarking.py#L269

the overhead reported for small kernels was very significant, but for larger ones (500us) it added ~15us which seemed fine.

there's also `torch.profiler.record_function` which worked pretty well for pytorch functions (creates a separate zone that covers the full op) but didn't seem to work when profiling iree for some reason.

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -51,117 +64,164 @@ def main():
         " of available devices, please use the environment variables `CUDA_VISIBLE_DEVICES` or "
         " `ROCR_VISIBLE_DEVICES` along with '--gpu-id=-1'.",
     )
-    args, extra_cli_args = parser.parse_known_args()
+    parser.add_argument(
+        "--splat-input-value",
+        default=None,
+        type=int,
```
ftynse, line 70: Nit: do we want float instead?

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -36,6 +42,13 @@ def main():
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     parser.add_argument("--commands-file", type=str, help="read commands from file")
+    parser.add_argument(
+        "--reference-backend",
+        type=str,
+        choices=["torch", "torch-compile"],
```
ftynse, line 48: How about accepting multiple options here, i.e., both torch and torch compiler?

iree/turbine/kernel/boo/driver/driver.py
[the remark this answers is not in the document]
ftynse: Is the overhead from using events constant-ish? If so, maybe we can have two events in a row, take the time difference between them and report it for future corrections?

PR #1111, opened 2025-08-05
[device] Use SKU as HIP target when possible

rkayaith:
Supplying a specific SKU will allow IREE to optimize for specific target details.

This is implemented as a table mapping torch device names to IREE SKU strings. For now I've included the known Instinct devices listed here:
https://github.com/iree-org/iree/blob/66b2f337c8019abf4716487ad32adea032afbb81/compiler/src/iree/compiler/Codegen/Dialect/GPU/TargetUtils/KnownTargets.cpp#L425-L440

Unfortunately this requires keeping the table in sync with IREE as new targets are added.

### DIFF
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -50,6 +50,17 @@ __all__ = [
     "DeviceState",
 ]
 
+TORCH_DEVICE_NAME_TO_SKU = {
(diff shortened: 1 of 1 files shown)


jerryyin (approved):
LGTM! Thanks for the fast PR!

PR #1113, opened 2025-08-06
[BOO] Remove force_single_dispatch

IanWood1:


zjgarvey (commented):
How do we know if this is working correctly? Can we add some tests to verify this doesn't regress, etc?

ftynse:
Why the rush? Can't we just treat this as an optimization option?

MaheshRavishankar:
> Why the rush? Can't we just treat this as an optimization option?

We can keep it as an optimization option, but it would be better to keep it default off. The default on is breaking  a lot of cases. The issue that you pointed to is triaged and I think there is a fix for that already.

zjgarvey:
Yeah, the change I made should work as a temporary workaround until something like https://github.com/iree-org/iree-turbine/pull/1134 lands (where the fusion spec for an op already determines whether it uses the attr).

zjgarvey (approved):
LGTM

PR #1114, opened 2025-08-06
[BOO] Save compilation stats file to cache

IanWood1:


zjgarvey:
@IanWood1 sorry for getting back to this kinda late.

I needed to update some compiler flags anyway (and customize them based on target backend). 

With https://github.com/iree-org/iree-turbine/pull/1118, you should be able to define a testing compiler flags callback with this option enabled, but I don't think I would want this to always be the default (especially since we are just going to throw it away unless the cache is enabled).

PR #1115, opened 2025-08-06
add an option to use aten for layernorm

ftynse:
Add a switch to all implementations whether to use the aten native layer norm functions or an equivalent implementation from primitives. The latter is sometimes more efficient when processed by the compiler. Add a test to ensure the implementation from primitives matches the aten version.


rkayaith (commented):
Sorry for the late review, this fell off my radar. Just some minor comments/questions.

PR #1134, opened 2025-09-03
[boo] use torch compile backend in driver

ftynse:
Go through `torch.compile` in the boo driver instead of calling the kernel directly. This allows to collect more realistic execution time statistics that are close to what the final user will see after integration. The direct kernel invocation is kept as a reference option to evaluate the overhead.

Tweak the profiler configuration to ignore initialization and cleanup steps so as to avoid skewing aggregate statistics.


ftynse:
For the multi-device case though, it looks like inductor wants to recompile every time because device ID changes. If somebody has an idea on fixing that, please go ahead! I tried having per-device functions/modules, but it doesn't seem to have helped (or I did it wrong).

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -217,33 +206,61 @@ def get_aggregate_stats(
...
+        else nullcontext()
+    )
+
     results: tuple[torch.Tensor, ...] | torch.Tensor | None = None
-    for iter in range(iter_per_device + 1):
-        for device_idx, launch_args in enumerate(per_device_args):
-            if iter == iter_per_device and device_idx >= rem_iter:
-                break
+    with profile_context as prof:
+        for iter in range(timing_args.iter + num_devices):
```
rkayaith, line 250: could you add a comment about the number of iterations here? I think this is "number of requested iterations" + "one warmup per device"? do we also need to add extra iterations for the 'needs cleenup' steps that have profiling skipped?

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -252,12 +269,22 @@ def run(
         print(
             f">>>\tresult #{i} shape: {result.shape}; dtype: {result.dtype}; device type: {result.device.type}"
         )
-    return
+    return prof if timing_args.time else None
+
+
+def get_torch_compiled_module(signature: OpSignature, backend: str) -> Callable:
+    mod = signature.get_nn_module(use_custom=False)
+    return torch.compile(mod, dynamic=False, backend=backend)
```
rkayaith, line 277: To completely match what users will get, we need to update the op signatures to set the layout of the inputs (i.e. `get_sample_args()` should be calling `input.to(memory_format=torch.channels_last)` for NHWC) rather than doing transposes inside `get_nn_module`. I _think_ this might work out to be close enough for now though, and we can just adjust the signature implementations in a separate step.

ftynse: Memory format only gives channels/3d-channels last. Do we care about other layouts?

zjgarvey: I don't think we care about too many other layouts, although for things like matmul, I think some simple transpose layouts are useful. 

It may be useful to do something generic for this.

E.g., we could take the inputs which are generated in "NHWC" format, permute them back to "NCHW" and those permuted tensors would be in channels-last format. If we pass these inputs to the boo-compiled `signature.get_nn_module(use_custom=False)`, where we manually set the signature layouts to the default "NCHW", ideally the memory format handling will work out properly through the `iree_boo` backend.

zjgarvey:
Yeah, I'll take a look today

ftynse:
Argh, didn't actually merge it because github makes stacked PRs needlessly painful. #1140 needs to be merged instead

ftynse:
Rebased

rkayaith:
@ftynse I've tried this out locally, and pushed a few small changes. I think the driver changes overall look good to me, however I'm seeing regressions on convolutions.

I think it would be a good idea to change the default backend here to stay as `iree_boo_legacy`, so we can get this merged. I'll continue to debug the issue and we can flip the default once they're resolved.

PR #1135, opened 2025-09-04
[boo] add experimental torch.compile backend variant, make layernorm go through it

ftynse:
Add a second, experimental BOO backend for the purposes of performance comparisons. It is allowed to have more fusion specs, including unprofitable ones and is not going to be on by default.

Add a pseudo-fusion spec that intercepts layer norm operations, but doesn't actually do fusion since we don't know at this point what do we want to fuse with. Known-good operations, without permutations, are accepted by the main BOO backend. The remaining operations are only accepted by the experimental backend.

### DIFF
--- a/tests/kernel/boo/fusion/integration_test.py
+++ b/tests/kernel/boo/fusion/integration_test.py
@@ -176,8 +176,8 @@ def test_boo_layer_norm_used(permuted: bool, experimental: bool):
             node_target = node.target
             break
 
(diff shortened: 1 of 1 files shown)


zjgarvey (approved):
Yeah, we do well with layer norm, so it sounds good to add it into the schema.

ftynse:
Rebased in main. Feel free to add fixes if needed and land

PR #1139, opened 2025-09-05
Added verbose option to avoid displaying info on terminal by default

pdhirajkumarprasad:
Added verbose option to avoid displaying info on terminal by default

### DIFF
--- a/iree/turbine/kernel/boo/driver/driver.py
+++ b/iree/turbine/kernel/boo/driver/driver.py
@@ -83,6 +83,12 @@ command-line arguments are appended to the arguments from the file.
         type=float,
         help="Use a splat value for inputs (defaults to random values).",
     )
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        default=False,
+        help="Print command/output on STDOUT.",
+    )
     return parser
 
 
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -131,15 +137,20 @@ def main():
...
+        if meta_args.verbose:
+            print(f"\n>>> {command}\n")
+        else:
+            print("Running test :", testCount)
         timing_args, runner_args = timing_parser.parse_known_args(driver_args)
         csv_file.write(shlex.join(driver_args) + ",")
         signature = BooOpRegistry.parse_command(shlex.join(runner_args))
 
-        if signature is None:
+        if signature is None and meta_args.verbose:
```
zjgarvey, line 153: We need to "continue" here whether we are verbose or not.

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -171,23 +189,24 @@ def main():
 
             zones = _extract_zones(prof)
 
-            if len(zones.keys()) == 0:
+            if len(zones.keys()) == 0 and meta_args.verbose:
```
zjgarvey, line 192: Same as above comment.

zjgarvey (changes requested):
You need to sign off on commits to pass DCO.

Instead of the conditionals increasing the indentation, I'd also prefer using a logger to control verbosity, or define a function like:

```python
def _print(msg: str, verbose: bool):
    if not verbose:
        return
    print(msg)
```

PR #1142, opened 2025-09-05
Remove pre-pad from non-custom conv impl

zjgarvey:
This was causing an additional memcopy since the pre-pad wasn't getting handled properly. We introduced this for IR gen, but since we are no longer using the non-custom impl for IR gen, it will be simpler to remove the prepad.


ftynse (approved):

zjgarvey:
@ftynse I think I'm going to close this and re-open against main, since I don't think it makes sense to merge this into your other branch.

PR #1153, opened 2025-09-17
[boo] Default to verbose driver output when no csv file is specified

rkayaith:
The current default behaviour is quite poor for interactive use, the only terminal output is
```
Running test : 1
```

This changes the default output to be quiet when an output csv file is specified, and verbose otherwise. `--verbose/--no-verbose` can be used to override this.

### DIFF
--- a/iree/turbine/kernel/boo/driver/driver.py
+++ b/iree/turbine/kernel/boo/driver/driver.py
@@ -88,8 +88,14 @@ command-line arguments are appended to the arguments from the file.
     parser.add_argument(
         "--verbose",
         action="store_true",
-        default=False,
+        default=None,
         help="Print command/output on STDOUT.",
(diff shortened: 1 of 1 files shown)


zjgarvey (approved):
Yeah, this is a good idea. To double check, I'm pinging @pdhirajkumarprasad to make sure he isn't running any scripts in the CI without specifying `csv` arg.

pdhirajkumarprasad (approved):
Approved but before merging, let me know so that I can add '--no-verbose' else it will dump o/p in CI log

rkayaith:
@pdhirajkumarprasad just to clarify, since the CI outputs to a csv it shouldn't be necessary to add `--no-verbose` (but it's probably best to add it anyways, just to be safe)

PR #1154, opened 2025-09-17
[boo] Add an entrypoint for the boo driver

rkayaith:
The driver script is now accessible as `iree-boo-driver` after installing the `iree_turbine` package.


zjgarvey (approved):
Nice, this seems useful.

PR #1157, opened 2025-09-19
Add support for tensor transfers in eager to allow for multi-device execution

sogartar:
Currently we only issue device transfer ops when exporting. With this change a new export-device-affinity-to-torch-device configuration map is introduced that allows us to do actual transfers in eager.

```
t = torch.tensor([1, 2], device="cuda:2")
with IreeDeviceAffinityToTorchDevice({
    DeviceAffinity(0): torch.device("cuda:2"),
    DeviceAffinity(1): torch.device("cuda:3")
}):
    t2 = transfer_to_logical_device("1", t) # transfer to cuda:3
    t3 = transfer_to_logical_device("0", t2) # transfer back to cuda:2
```

### DIFF
--- a/iree/turbine/aot/tensor_traits.py
+++ b/iree/turbine/aot/tensor_traits.py
@@ -17,17 +17,12 @@ __all__ = [
 ]
 
 
+@dataclass(frozen=True)
 class DeviceAffinity:
     """This is used to provide device affinities to exported function arguments."""
 
-    def __init__(self, ordinal: int, queues: Optional[list] = None):
-        self.ordinal = ordinal
-        self.queues = queues
-
-    def __eq__(self, other) -> bool:
-        if not isinstance(other, DeviceAffinity):
-            return False
-        return self.ordinal == other.ordinal and self.queues == other.queues
+    ordinal: int
+    queues: Optional[list] = None
 
     def __repr__(self) -> str:
         if self.queues is None:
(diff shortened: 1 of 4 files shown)


iree/turbine/ops/iree.py
```diff
@@ -151,6 +167,53 @@ def generate(self, ksel: KernelSelection, kb: KernelBuilder):
         kb.yield_results(result)
 
 
+################################################################################
+# IREE device affinity to torch device map
+################################################################################
```
sogartar, line 172: I am not sure if I should move this section somewhere else. For example in [iree/turbine/runtime/device.py](https://github.com/iree-org/iree-turbine/blob/66830917b9be8253d83f80abe4dcccfbe8cc6352/iree/turbine/runtime/device.py).

iree/turbine/aot/tensor_traits.py
```diff
@@ -17,17 +17,12 @@
 ]
 
 
+@dataclass(frozen=True)
```
rsuderman, line 20: No, device affinity is not a dataclass. You should only use this annotation when the type is struct like.

rsuderman (changes requested):
This looks like a solution in search of a problem. Device affinity is predominately used for when tracing and will essentially be disconnected from device level tracing. E.g. when tracing for 8 devices this could be done a a single cpu torch instance as devices are not required at tracing time.

Requiring the device affinity and tensor placements are aligned / correct will likely just generate more upkeep for a feature that is not needed.

PR #1165, opened 2025-09-29
[BOO] Add flag to run transpose filter preprocessing pipeline

yzhang93:
This flag is applied together with https://github.com/iree-org/iree/pull/22100 to transpose the convolution filter layout for input backward convolutions.


yzhang93:
> Is this expected to have any effect on forward convs?

No, this won't affect forward or weight backward convs.

PR #1168, opened 2025-09-29
[BOO] Rework `numerics.py` script to use `iree_boo_experimental` backend.

zjgarvey:
This PR addresses #1167 by calling into the `iree_boo_experimental` `torch.compile` backend.

The main points:

- Calls backward implementations directly. 
- Removes the need for various `OpSignature` methods.
- Adds `boo_cpu` as a comparison point and compares all available result tensors.
- Modifies csv writing to make new rows for each result index.
- Attempts to make typing a bit more descriptive.

Besides a general re-alignment of the numerics validation tool, the goal of this PR is to simplify the process of adding new `op_export` files by reducing the one-off uses of certain abstract methods for `OpSignature`.

Some future goals for the numerics script:

- Add command-line options to specify which result tensors to generate. E.g. if only `pytorch_gpu` and `boo_gpu` should be compared.
- More elegantly handle the collation of result comparisons for multi-output computations.
- Find an acceptable numerics validation strategy and automate CI testing for configs we care about.

### DIFF
--- a/iree/turbine/kernel/boo/driver/numerics.py
+++ b/iree/turbine/kernel/boo/driver/numerics.py
@@ -152,6 +152,8 @@ def _run(
             for i in range(num_results):
                 summary[i][f"{cmp_key} vs. {ref_key}"] = comparison_results[i]
         results[c] = tuple(summary)
+        # Reset the cache to ensure we don't silently hit re-compile limits.
+        torch.compiler.reset()
 
     return results
 


iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -68,107 +94,66 @@ def _wrap_in_tuple(x: _T | tuple[_T, ...] | list[_T]) -> tuple[_T, ...]:
...
-            "gpu vs. pt_cpu": _compare(grad, args_cpu[derivative_arg_index].grad),
-            "pt_gpu vs. pt_cpu": _compare(
-                args[derivative_arg_index].grad, args_cpu[derivative_arg_index].grad
-            ),
-        }
+            assert len(comparison_results) == num_results
+            for i in range(num_results):
+                summary[i][f"{cmp_key} vs. {ref_key}"] = comparison_results[i]
+        results[c] = tuple(summary)
+        # Reset the cache to ensure we don't silently hit re-compile limits.
```
dan-garvey, line 155: Oof, worth filing a bug upstream?

PR #1170, opened 2025-09-30
[WIP][BOO] Rework graph ops to avoid re-exporting layout managed modules

zjgarvey:
Small issues to address:

1. Failing batch norm test (maybe we aren't decomposing with new import path).
2. Remove remaining dead code.
3. Add some small unit tests to verify graph utils.

### DIFF
--- a/iree/turbine/kernel/boo/fusion/replacement.py
+++ b/iree/turbine/kernel/boo/fusion/replacement.py
@@ -1,81 +1,119 @@
 from collections.abc import Callable
+from typing import Literal
 
+from operator import getitem
 import torch
 from torch import fx
-from torch.fx.node import Argument, Target
+from torch.fx.node import Target, Node
 
-
+from ..op_exports.conv import DEFAULT_LAYOUTS
 from iree.turbine.kernel.boo import ops as boo_ops
+from ..ops.utils import get_memory_format_permutation, MemoryFormatPermutation
+from ..ops.graph import call_permute, permute_metadata
 
-# 1 to 1 op replacements. The replacement function should return the new target
-# and arguments, or 'None' if it doesn't apply.
-ReplacementSchema = dict[
-    Target,
-    Callable[
-        [tuple[Argument, ...], dict[str, object]],  # (node.args, node.meta)
-        tuple[Callable, tuple[Argument, ...]] | None,  # (new_target, new_args) | None
-    ],
-]
+# 1 to 1 op replacements. The replacement function should modify input Node graph if applicable.
+ReplacementSchema = dict[Target, Callable[[Node], None]]
 
 
 def apply_replacements(graph: fx.Graph, replacements: ReplacementSchema):
     for node in graph.nodes:
         if node.op == "call_function":
-            replacer_fn = replacements.get(node.target, lambda *_: None)
-            replacement = replacer_fn(node.args, node.meta)
-            if replacement is None:
-                continue
-            target, target_args = replacement
-            with graph.inserting_after(node):
-                call_boo = graph.call_function(target, target_args)
-            node.replace_all_uses_with(call_boo, propagate_meta=True)
-            graph.erase_node(node)
+            replacer_fn = replacements.get(node.target, lambda n: None)
+            replacer_fn(node)
 
 
-def replace_aten_convolution(args: tuple[Argument, ...], meta: dict[str, object]):
+def _apply_perms(
+    src_node: Node,
+    perms: MemoryFormatPermutation | None,
+    perm_item: Literal["permutation", "inverse_permutation"],
+) -> Node:
+    if perms is None:
+        return src_node
+    new = call_permute(src_node, getattr(perms, perm_item))
+    new.meta = permute_metadata(src_node, getattr(perms, perm_item))
+    return new
+
+
+def replace_aten_convolution(node: Node):
     "Replace 'torch.ops.aten.convolution' with custom BOO implementation."
     (
-        input,
-        weight,
-        bias,
+        x,
+        w,
+        b,
         stride,
         padding,
         dilation,
         transposed,
         _output_padding,
         groups,
-    ) = args
+    ) = node.args
 
-    # BOO convolution doesn't support transpose. 'output_padding' is ignored
-    # in non-transpose cases.
     if transposed is not False:
-        return None
+        return
 
-    example_out = meta["val"]
-    assert isinstance(example_out, torch.Tensor)
-    output_is_channels_last = example_out.is_contiguous(
-        memory_format=torch.channels_last
+    graph = node.graph
+
+    assert isinstance(x, Node)
+    x_fake = x.meta.get("val")
+    assert isinstance(x_fake, torch.Tensor)
+    assert isinstance(w, Node)
+    w_fake = w.meta.get("val")
+    assert isinstance(w_fake, torch.Tensor)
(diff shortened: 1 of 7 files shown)


iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -59,18 +62,146 @@ def _get_schema(
...
+        stride=permuted_val.stride(),
+        memory_format=torch.contiguous_format,
+        is_quantized=og_meta.is_quantized,
+        qparams=og_meta.qparams,
+    )
+    new_meta = {
+        "val": permuted_val,
+        "tensor_meta": permuted_meta,
+    }
+    return new_meta
```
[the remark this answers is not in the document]
zjgarvey: No, we aren't using any other metadata during mlir import (i.e., in `FxImporter` from torch-mlir).

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -59,18 +62,146 @@ def _get_schema(
...
+    for ret in output_og_args:
+        assert isinstance(
+            ret, Node
+        ), f"Expected returns to be nodes. Got {ret} with type {type(ret)}."
+        new_ret, _perm = convert_output(ret)
+        permuted_output_args.append(new_ret)
+        output_perms.append(_perm)
+
+    # Create an output node.
+    g.create_node(op="output", target="output", args=(tuple(permuted_output_args),))
```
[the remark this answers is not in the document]
zjgarvey: No, it doesn't copy the output node (it wouldn't really make sense from a utility perspective). The code is short for this method, and even if we request `return_output_node` it returns the output node from the source graph of the copy (`g`), and never copies the output node itself to the target graph (`self`).

```python
    for node in g.nodes:
            if node.op == "output":
                rv = map_arg(node.args[0], lambda n: val_map[n])
                return rv if not return_output_node else (rv, node)
            val_map[node] = self.node_copy(node, lambda n: val_map[n])
    return None
```

iree/turbine/kernel/boo/fusion/replacement.py
```diff
@@ -84,18 +122,91 @@ def replace_aten_convolution_backward(
...
+        with graph.inserting_before(consumer):
+            new_output = graph.call_function(getitem, args=(replacement_conv, index))
+            new_output.meta = (
+                consumer.meta
+                if _perm is None
+                else permute_metadata(consumer, _perm.permutation)
+            )
+            replacement = _apply_perms(new_output, _perm, "inverse_permutation")
+            consumer.replace_all_uses_with(replace_with=replacement)
+            graph.erase_node(consumer)
```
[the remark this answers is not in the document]
zjgarvey: Yeah, that's a good idea.

iree/turbine/kernel/boo/fusion/apply.py
```diff
@@ -17,12 +19,21 @@
...
     "fusion_transform",
 ]
 
 
+def infer_example_inputs(
+    graph_module: fx.GraphModule,
+) -> tuple[torch.Tensor | None, ...]:
+    return tuple(
+        n.meta.get("val") for n in graph_module.graph.find_nodes(op="placeholder")
+    )
```
rkayaith: do we want to allow `None` results here? what cases would that be expected (and I'm not sure the caller is handling `None` properly)

iree/turbine/kernel/boo/fusion/replacement.py
```diff
@@ -1,81 +1,163 @@
...
+        to_layout(w_perms),
+        to_layout(output_perms),
     )
+    with graph.inserting_before(node):
+        replacement_conv = graph.call_function(
+            torch.ops.boo.layout_customizable_convolution.default, args=call_args
+        )
+        replacement_conv.meta = (
+            node.meta
+            if output_perms is None
```
rkayaith: when would `output_perms` be `None`?

iree/turbine/kernel/boo/runtime/launch.py
```diff
@@ -288,3 +292,77 @@ def get_launchable(
...
+            func_op = module_op.regions[0].blocks[0].operations[0]
+            try:
+                pipeline_attr = Attribute.parse(
+                    '#util.preprocessing_pipeline<"iree-preprocessing-make-single-dispatch">'
+                )
+                func_op.attributes["preprocessing_pipeline"] = pipeline_attr
+            except MLIRError:
+                warnings.warn(
+                    f"Failed to attach #util.preprocessing_pipeline attr to func op. Please try using a newer version of IREE."
+                )
```
rkayaith: we can probably drop the `try/except` at this point

tests/kernel/boo/fusion/integration_test.py
```diff
@@ -46,11 +46,9 @@ def test_custom_boo_conv_used():
     [compiled_module] = recorder.graphs
     assert isinstance(compiled_module, fx.GraphModule)
     [call_node] = [n for n in compiled_module.graph.nodes if n.op == "call_function"]
-    # Make sure we're using 'boo.ops.convolution_replacement'. We have to do a
-    # string check unfortunately, as the target is a fused custom op that we
-    # can't inspect.
+    # Make sure we're using layout customizable op.
     call_node_target_str = str(call_node.target)
-    assert call_node_target_str.startswith("boo.fused_op_convolution_replacement_")
+    assert call_node_target_str.startswith("boo.fused_op_layout_cus")
```
rkayaith: i think this is no longer testing what was intended? seems like it's only checking whether we're using BOO, but for this test we want to check whether we're using the custom conv replacement. same thing for `test_boo_convolution_backward_replacement`

rkayaith (commented):
Generally looks good to me. Though I'm a bit confused about what cases we expect metadata to be missing, so it's hard to follow whether those paths are correct.

iree/turbine/kernel/boo/fusion/replacement.py
```diff
@@ -1,81 +1,163 @@
...
+
+def _apply_perms(
+    src_node: Node,
+    perms: MemoryFormatPermutation | None,
+    perm_item: Literal["permutation", "inverse_permutation"],
+) -> Node:
+    if perms is None:
+        return src_node
+    new = call_permute(src_node, getattr(perms, perm_item))
+    new.meta = permute_metadata(src_node, getattr(perms, perm_item))
```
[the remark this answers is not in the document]
zjgarvey: Sounds good. I'll find a way to simplify this.

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -59,18 +62,170 @@ def _get_schema(
...
+def permute_metadata(
+    source_node: Node, perms: Sequence[int] | None | tuple[Sequence[int] | None, ...]
+) -> dict:
+    """Returns a node meta resulting from applying `perm` to `source_node.meta`.
+
+    If `source_node` returns multiple outputs, this will only return `val` metadata, and a tuple of permutations are expected.
+    This is only handling `tensor_meta` and `val` entries, since only these are being used by the mlir import path.
+    """
+
+    def _permute_fake_tensor(og_val, perm) -> torch.Tensor | None:
```
rkayaith: could you type hint the args of all these local functions too, mypy doesn't try to infer them

rkayaith (approved):
LGTM, thanks for all the digging for this!

PR #1175, opened 2025-10-10
remove split and split_with_sizes from decomp lists

raayandhar:
See https://github.com/llvm/torch-mlir/issues/4339 for more context.

We need to remove `split.tensor` and `split_with_sizes` from the decomp list so we don't further decompose to `as_strided`, which is problematic as described in the issue above.


zjgarvey (approved):
Thanks, LGTM

PR #1181, opened 2025-10-16
[boo driver] Add aten.mm/bmm parsers

rkayaith:
Adds support for `aten.mm` and `aten.bmm` through the BOO driver. The interface matches how arguments are specified in pytorch chrome traces. Examples:
```
iree-boo-driver aten::mm "((3840, 4352), (4352, 3840))" "('c10::BFloat16', 'c10::BFloat16')" "((1, 3840), (3840, 1))" "('', '')"
```
```
iree-boo-driver aten::bmm "((85, 3, 3), (85, 3, 2304))" "('c10::BFloat16', 'c10::BFloat16')" "((9, 3, 1), (6912, 2304, 1))" "('', '')"
```

The parser logic is generic and could be re-used for other ops as well, however I didn't implement handling of concrete args yet, which will be necessary for most other ops.

### DIFF
--- a/iree/turbine/kernel/boo/op_exports/aten.py
+++ b/iree/turbine/kernel/boo/op_exports/aten.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import ast
+import base64
 from collections.abc import Sequence
 from dataclasses import asdict, dataclass
 import torch
@@ -111,15 +112,18 @@ class AtenSignature(OpSignature):
     @property
     @override
     def func_name(self) -> str:
-        return ":".join(
-            [
-                self.name,
-                str(self.input_dims),
-                str(self.input_type),
-                str(self.input_strides),
-                str(self.concrete_inputs),
-            ]
-        )
+        # This name is used as a file system path, but the fields here may
+        # contain special characters. A URL-safe b64 encode ensures only valid
+        # characters are used.
+        return base64.urlsafe_b64encode(
+            (
+                self.name
+                + str(self.input_dims)
+                + str(self.input_type)
+                + str(self.input_strides)
+                + str(self.concrete_inputs)
+            ).encode()
+        ).decode()
 
     @override
     def as_init_kwargs(self) -> dict[str, Any]:


iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -258,6 +257,8 @@ def run(
         get_device_from_torch(device).hal_device.allocator.trim()
     # Reset torch.compile caches to avoid hitting re-compile limits.
     torch.compiler.reset()
+    # Reset BOO caches to isolate tests.
+    LaunchableRuntimeCache.reset()
```
zjgarvey: The func names should be separate. 

E.g., running with boo cache enabled, it won't matter if the runtime cache is reset, it's just going to load the cached files based on the func name.

Also, I'm not sure if `:` is allowed in file names/mlir func names either. Should definitely spend a bit to come up with a way to make the name safe/unique to the config.

zjgarvey (commented):
I think it would be good to uniqueify the func names.

Maybe add simple parser tests for other ops. I guess only conv, since the other ones that aren't customized can just go through this aten parser now.

rkayaith:
> Maybe add simple parser tests for other ops. I guess only conv, since the other ones that aren't customized can just go through this aten parser now.

It's on my todo list to add CI tests which invoke the driver, which should cover this, but I'll leave that for a follow-up if that's okay (for now I've manually verified the commands in `sample_commands`.

iree/turbine/kernel/boo/driver/driver.py
rkayaith: this is fixed now, and there's no issue with weird characters

iree/turbine/kernel/boo/op_exports/aten.py
```diff
@@ -0,0 +1,151 @@
...
+        raise NotImplementedError()
+
+    @override
+    def arrange_backward_launch_args(self, forward_args, forward_results):
+        raise NotImplementedError()
+
+    @property
+    @override
+    def func_name(self) -> str:
+        return ":".join(
```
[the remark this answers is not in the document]
rkayaith: mlir functions don't have any issues with special symbols, and linux file names only disallow `/`. windows _does_ have a bunch of restrictions though: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

I've changed this to do a b64 encoding which will ensure only valid characters are used (this is a bit safer than character replacement since that could cause unintended collisions in some cases, though I doubt we'd ever hit a case like that).

PR #1185, opened 2025-10-21
[boo driver] Properly format csv output

rkayaith:
Use the `csv` module for writing CSV output. This properly handles escaping values that have commas or spaces in them.


zjgarvey (approved):
Nice.

PR #1195, opened 2025-10-31
[BOO] Add split reduction and global optimization flags

yzhang93:
Adding two flags:
1. For split reduction on weight backward convolutions.
2. Global optimization to control edge reshape propagation https://github.com/iree-org/iree/pull/22438.


rkayaith (approved):
We'll have to bump the IREE pin before this can be merged

PR #1206, opened 2025-11-06
Make FxProgramsBuilder compatible with torch 2.9

rkayaith:
The previous approach no longer works with pytorch 2.9, hitting the issue described here due to the use of variadics: https://github.com/pytorch/pytorch/issues/162831

This switches to a new approach of dynamically defining a function using `exec` (via the `decorator` library) to work around the issue.


Groverkss (approved):
LGTM

PR #1220, opened 2025-11-17
[BOO] Enable in-place invocations for graph ops

zjgarvey:
This change modifies a graph op's `GraphModule` to introduce output init tensor placeholders and copy ops from the outputs to to these placeholders. Additionally, we construct an `ExportGraphSignature` to encode these input mutations, and directly assemble a minimal `ExportedProgram` to pass to the `FxImporter`.

Although we could modify the graph module more simply and re-export with torch.export, this became an untenable approach, since this export uses gpu memory, and has issues with the pre-existing fake mode from `torch.compile`. 

Another approach could be manually modifying the exported MLIR, but this would require translating quite a bit of information back and forth between MLIR and pytorch.

### DIFF
--- a/tests/kernel/boo/ops/boo_conv_aot_test.py
+++ b/tests/kernel/boo/ops/boo_conv_aot_test.py
@@ -135,8 +135,5 @@ def test_AOT_layout_conv_replacement(device: torch.device, boo_cache_dir: Path):
     cached_items = list(Path.glob(boo_cache_dir, "*/"))
     op_name = graph_op._qualified_op_name.split(":")[-1]
     # Re-running this test may not re-save to the cache if the torch-op still exists.
-    assert (
-        len(cached_items) <= 1
-    ), f"Expected at most one cached item. Got {len(cached_items)}."
-    if len(cached_items) == 1:
-        assert cached_items[0].name == op_name, "Expected op name to match cache entry."
+    assert len(cached_items) == 1, "Expected one cached item for this test."
+    assert cached_items[0].name == op_name, "Expected op name to match cache entry."


iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -281,14 +290,105 @@ def _handle_layouts(
...
+    output_mem_format_perms: Sequence[MemoryFormatPermutation | None],
+) -> tuple[ExportedProgram, list[MemoryFormatPermutation | None], list[torch.Tensor]]:
+    std_g = std_gm.graph
+    output_node = std_g.output_node()
+    outs = output_node.args[0]
+    outs = outs if isinstance(outs, (tuple, list)) else (outs,)
+    input_mutations = {}
+    seen_names = []
+    init_perms = []
+    init_fakes = []
```
rkayaith: can you add type hints to these

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -281,14 +290,105 @@ def _handle_layouts(
...
+        with std_g.inserting_after(o):
+            copy_node = std_g.call_function(
+                torch.ops.aten.copy.default, args=(init_plh, o, True)
+            )
+            copy_node.meta = {k: v for k, v in init_plh.meta.items()}
+            input_mutations[init_name] = copy_node
+
+    new_outputs = tuple(input_mutations.values()) + tuple(outs)
+    output_node.args = (new_outputs,) + tuple(output_node.args[1:])
+    input_specs = [
```
rkayaith: could you add a comment here about why we don't just use `torch.export`. this function could probably use a docstring too

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -281,14 +290,105 @@ def _handle_layouts(
...
     gm: GraphModule,
     og_gm: GraphModule,
     op_name: str,
-    input_mem_format_perms,
-    output_mem_format_perms,
+    input_mem_format_perms: Sequence[MemoryFormatPermutation | None],
+    output_mem_format_perms: Sequence[MemoryFormatPermutation | None],
     *,
     force_single_dispatch: bool = False,
+    inplace_convert: bool = True,
```
rkayaith: I would either remove this flag, or add a test that exercises the `False` path so we know it still works

iree/turbine/kernel/boo/ops/graph.py
```diff
@@ -281,34 +298,145 @@ def _handle_layouts(
...
+    Some implicit assumptions:
+
+    1. The graph modules we are encountering are stateless.
+    2. All inputs/outputs are tensors with valid metadata.
+    """
+    std_g = std_gm.graph
+    output_node = std_g.output_node()
+    outs = output_node.args[0]
+    outs = outs if isinstance(outs, (tuple, list)) else (outs,)
+    input_mutations = {}
```
rkayaith: can you type hint this as well (basically any empty list/dict/tuple variable like this needs a hint for type checking to work)

rkayaith (approved):

tests/kernel/boo/ops/boo_conv_aot_test.py
```diff
@@ -133,6 +133,10 @@ def test_AOT_layout_conv_replacement(device: torch.device, boo_cache_dir: Path):
...
-    assert len(cached_items) == 1, f"Expected one cached item, got {cached_items}."
-    p = cached_items[0]
-    assert p.name.endswith("_2x16x16x32xfloat32_2x1x1x32xfloat32")
+    op_name = graph_op._qualified_op_name.split(":")[-1]
+    # Re-running this test may not re-save to the cache if the torch-op still exists.
+    assert (
+        len(cached_items) <= 1
+    ), f"Expected at most one cached item. Got {len(cached_items)}."
+    if len(cached_items) == 1:
+        assert cached_items[0].name == op_name, "Expected op name to match cache entry."
```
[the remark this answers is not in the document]
zjgarvey: Ah yeah. I just wanted to change this test to reflect the full name.

The rest is an artifact of locally trying to hoist the launchable creation out of runtime. In that case, all of the caches indeed get cleared, but the launchable itself is part of the pytorch op definition, and the op definition doesn't get cleared and doesn't need caches anymore.

I'll restore for now.

PR #1230, opened 2025-11-21
[runtime][rocm] ensure we do not sync streams during dlpack handoff

zjgarvey:
Recently, the meaning of `stream=None` has been changed to mean that we are synchronizing with the default stream (see https://github.com/pytorch/pytorch/commit/1d526fe78fcb8e6185f25d4b4340599b35f6a6d9 ).

We could pass `t.__dlpack__(stream=-1)`, but recently added checks in this method will actually error out if the `torch.cuda.current_device` doesn't match that of `t` (this is likely to ensure the "default" stream is appropriate, but we aren't doing any stream synchronization here). Also, things like checking `t.layout == torch.strided_layout` are automatic from our contiguous check. Therefore, I'm opting to bypass `Tensor.__dlpack__` entirely, as all checks in this function are completely redundant for us.

This change results in a reduction of a whopping 25us of CPU overhead *per-pointer* on recent (nightly) pytorch versions.

### DIFF
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -548,9 +548,9 @@ def _device_import_torch_tensor_cuda_hip(
     # No explicit synchronization is made during the pointer handoff to IREE.
     # We launch kernels through IREE runtime on the same stream as pytorch. If using multiple
     # streams, the user is expected to properly manage stream synchronization.
-    # The lower level `_C._to_dlpack` is used to bypass redundant checks in `Tensor.__dlpack__`.
+    # The direct call to `torch.to_dlpack` is used to bypass redundant checks in `Tensor.__dlpack__`.
     # E.g., we know `t`` has a strided layout since we force it to be continuous.
-    capsule = torch._C._to_dlpack(t)
+    capsule = torch.to_dlpack(t)
     bv = device.hal_device.from_dlpack_capsule(capsule)
     return bv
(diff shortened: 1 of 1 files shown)


iree/turbine/runtime/device.py
```diff
@@ -545,10 +545,12 @@ def _device_import_torch_tensor_cuda_hip(
...
     if not t.is_contiguous():
         t = t.contiguous()
-    # The None passed to tensor.__dlpack__ indicates we are doing no stream synchronization here.
+    # No explicit synchronization is made during the pointer handoff to IREE.
     # We launch kernels through IREE runtime on the same stream as pytorch. If using multiple
     # streams, the user is expected to properly manage stream synchronization.
-    capsule = t.__dlpack__(stream=None)
+    # The lower level `_C._to_dlpack` is used to bypass redundant checks in `Tensor.__dlpack__`.
+    # E.g., we know `t`` has a strided layout since we force it to be continuous.
+    capsule = torch._C._to_dlpack(t)
```
rkayaith: I'm not sure calling internal methods like this is a good idea, e.g. on older versions `__dlpack__` is calling `torch.to_dlpack` instead (I'm not sure what the difference is, or if their implementation has changed across versions).
Can we pass the check by setting the current cuda stream, and use `stream=-1` to avoid synchronization?

PR #1234, opened 2025-11-25
Add a max allowed stddev to benchmarking

lialan:
Adding a new `max-stddev-allowed` option.

So if `--max-stddev-allowed=0.1` is turned on, it will retry until stddev falls below the threshold. After 5 failed tries, it will just exit the driver as it is likely that we are testing on a shared machine.

This is probably the easiest way to guard us off from invalid perf numbers, which we have collected in previous tries.

### DIFF
--- a/iree/turbine/kernel/boo/driver/driver.py
+++ b/iree/turbine/kernel/boo/driver/driver.py
@@ -96,7 +96,7 @@ list of arguments.
         "--max-stddev-allowed",
         default=None,
         type=float,
-        help="Maximum allowed standard deviation (in microseconds). If exceeded, the run will be retried up to 5 times.",
+        help="Maximum allowed standard deviation as a ratio of the mean (e.g., 0.1 for 10%%). If exceeded, the run will be retried up to 5 times.",
     )
     parser.add_argument(
         "--verbose",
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -92,6 +92,12 @@ def _get_main_driver_parser() -> argparse.ArgumentParser:
         type=float,
         help="Use a splat value for inputs (defaults to random values).",
     )
+    parser.add_argument(
+        "--max-stddev-allowed",
+        default=None,
+        type=float,
+        help="Maximum allowed standard deviation (in microseconds). If exceeded, the run will be retried up to 5 times.",
```
zjgarvey: I'm going to take an hour or two to consider this. 

For certain situations, e.g., kernels with a mean launch time of 6000us, setting the same absolute stddev threshold as 5us kernels seems flaky. I'm not sure if specifying a threshold for something like (stddev/mean) is better or not. Eg., maybe 2us is acceptable stddev for a 10us kernel, but 20% of 6000us seems unrealistic to allow.

lialan: @zjgarvey I think you are right. we should change it to percentage instead.

PR #1246, opened 2025-12-10
Add iree_tensor_ext.barrier ops to turbine

rkayaith:
This exposes the IREE barrier ops as turbine custom ops. These are intended to be used as part of the custom lowering of batch norm in BOO.

### DIFF
--- a/iree/turbine/ops/iree.py
+++ b/iree/turbine/ops/iree.py
@@ -16,6 +16,7 @@ from ..support.ir_imports import (
     StringAttr,
     Value,
     flow_d,
+    iree_tensor_ext_d,
     tensor_d,
 )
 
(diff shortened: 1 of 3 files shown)


iree/turbine/ops/iree.py
```diff
@@ -151,6 +152,50 @@ def generate(self, ksel: KernelSelection, kb: KernelBuilder):
...
+    signature = "compute_barrier_start(Tensor tensor) -> Tensor"
+
+    def select(self, ksel: KernelSelection):
+        ta = ksel.arg_tensor(0)
+        spec = [i for i, s in enumerate(ta.t.shape) if isinstance(s, int)]
+        ta.specialize_dims(*spec)
+        ksel.return_tensor(ta.t).specialize_dims(*spec)
+
+    def eager_execute(self, tensor: torch.Tensor):
+        return tensor.clone()
```
zjgarvey: Why clone?

tests/ops/iree_test.py
```diff
@@ -88,6 +89,39 @@ def forward(self, x):
...
 
 
+class TensorBarrierTest(unittest.TestCase):
+    def testEager(self):
+        t1 = torch.randn(3, 4)
+        t2 = ops.iree.compute_barrier_start(t1)
+        t3 = ops.iree.compute_barrier_end(t2)
+        assert torch.all(t1 == t2)
+        assert torch.all(t2 == t3)
+        assert t1 is not t2 is not t3
```
zjgarvey: Oh. Maybe I misunderstood when I made the comment about clone. Do we actually want these to be distinct pointers? I thought this op was merely to indicate the boundaries of a dispatch region.

zjgarvey (commented):
This looks good, but I have a quick question about the nature of the op before approving.

PR #1247, opened 2025-12-11
[runtime] Try to load hip libraries from `rocm_sdk` package if available

zjgarvey:
This patch uses the runtime flag `--hip_dylib_path` together with the `file:` option to explicitly set a path to a library for loading, allowing us to match the exact path pytorch uses whenever `rocm_sdk` is installed.

See https://github.com/pytorch/pytorch/blob/db17cbfb36094ca6a813bfdaf0f4fd20944fc35a/torch/cuda/_utils.py#L11-L24

I'm not sure if using runtime flags is the best way to handle the library loading. Being able to set this thread-locally from python is important to have turbine "just work" seamlessly with nightly pip packages from the Rock.

### DIFF
--- a/iree/turbine/runtime/device.py
+++ b/iree/turbine/runtime/device.py
@@ -741,7 +741,8 @@ def _align_hip_library_with_torch() -> None:
         _CURRENT_THREAD.hip_dylib_set = True
     hip_dylib_path = str(rocm_sdk.find_libraries("amdhip64")[0].absolute())  # type: ignore
     loaded_library = ctypes.CDLL(hip_dylib_path)
-    if loaded_library._handle != torch.cuda._utils._get_gpu_runtime_library()._handle:
+    torch_gpu_library = torch.cuda._utils._get_gpu_runtime_library()._handle  # type: ignore
+    if loaded_library._handle != torch_gpu_library:
         raise RuntimeError(
             """Runtime library inconsistency found between pytorch and IREE.\n
 


zjgarvey:
Let's also wait on a review from @ScottTodd, also. Thanks @rkayaith for adding him as a reviewer. 

He should have context on the python library packaging from [ROCm/TheRock](https://github.com/ROCm/TheRock).

tests/runtime/device_test.py
```diff
@@ -27,6 +29,49 @@
...
+        print("Not using python install of rocm_sdk.")
+        return
+    assert len(paths) != 0
+    return str(paths[0].absolute())
+
+
+def test_hip_dylib_path_iree(dylib_path: str | None):
+    if dylib_path is None or not torch.cuda.is_available():
+        # This skip isn't a mark, since we are skipping based on a fixture value.
+        pytest.skip(reason="Test requires pytorch with `rocm_sdk` python install.")
```
rkayaith: I think you can just skip inside the fixture

tests/runtime/device_test.py
```diff
@@ -27,6 +29,49 @@
 from iree.turbine.support.exceptions import *
 
 
+@pytest.fixture
+def dylib_path() -> str | None:
+    try:
+        import rocm_sdk
+
+        paths = rocm_sdk.find_libraries("amdhip64")
+    except (ImportError, ModuleNotFoundError, KeyError):
```
rkayaith: do we actually want to ignore `ImportError` and `KeyError` here? those seem like they might indicate a bug

tests/runtime/device_test.py
```diff
@@ -27,6 +29,49 @@
...
+    d = get_device_from_torch(torch.device("cuda:0"))
+    device_info = d.dump_device_info().splitlines()
+    found_path = ""
+    for line in device_info:
+        items = line.split(" ")
+        if len(items) < 3:
+            continue
+        if items[1] == "amdhip64_dylib_path:":
+            found_path = items[2]
+            break
```
rkayaith: could you have a comment with an example of what's being parsed here

ScottTodd (approved):

iree/turbine/runtime/device.py
```diff
@@ -704,9 +705,43 @@ def _create_cuda_device(torch_device: torch.device, props) -> Optional[Device]:
...
+
+    The benefits of doing this:
+
+    1. If the only rocm install is through python, IREE won't find it without help.
+    2. If there is an existing install of rocm, e.g. in `/opt/rocm`, there may be
+       version incompatibilty between this version and the one pytorch is using.
+       this incompatibility can cause issues with the torch profiler.
+
+    If `rocm_sdk` is available to python, pytorch uses it for runtime libraries.
+    Otherwise it looks in the same places as IREE, so we shouldn't need to set this explicitly.
```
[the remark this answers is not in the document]
zjgarvey: Thanks! I've updated the comment.

iree/turbine/runtime/device.py
```diff
@@ -709,32 +711,48 @@ def _align_hip_library_with_torch() -> None:
...
-    _CURRENT_THREAD.hip_dylib_set = True
+    except ModuleNotFoundError:
+        return
+    finally:
+        # Whether we were successful or not, don't try again.
+        _CURRENT_THREAD.hip_dylib_set = True
+    hip_dylib_path = str(rocm_sdk.find_libraries("amdhip64")[0].absolute())  # type: ignore
+    loaded_library = ctypes.CDLL(hip_dylib_path)
+    if loaded_library._handle != torch.cuda._utils._get_gpu_runtime_library()._handle:
+        raise RuntimeError(
```
zjgarvey: @AWoloszyn Do you think this is a good way to compare hip libraries without needing to do platform-dependent checks?

rkayaith (approved):
thanks for digging into this!

iree/turbine/runtime/device.py
zjgarvey: Maybe it isn't? E.g., if pytorch created the CDLL in a slightly different way, but it ended up being the same library, would the `_handle` attribute be different? I need to look into this a tiny bit more.

AWoloszyn: I would EXPECT them to be shared across multiple calls to dlopen (on the C++ side) but not sure what python would do differently if anything.

PR #1248, opened 2025-12-15
Add batch norm support to BOO

rkayaith:
This adds NHWC batch norm support to BOO in the experimental backend. An initial parser matching the MIOpen format is also added, though only support for certain flags is supported.

Batch norm has a custom replacement added, which forces the batch norm to be computed in CNHW layout, with transpose dispatches at the input/output surrounding it.

This is currently the preferred approach for IREE as batchnorm contains reductions that are parallel over the channels dimension, and so far codegen has been focused on supporting on outer-parallel reductions.
Having transpose dispatches is inefficient, so future work will aim to improve inner-parallel performance and remove this replacement.

### DIFF
--- a/iree/turbine/kernel/boo/op_exports/batch_norm.py
+++ b/iree/turbine/kernel/boo/op_exports/batch_norm.py
@@ -6,13 +6,9 @@
 
 import argparse
 from dataclasses import asdict, dataclass
-from enum import IntEnum
 from typing import Any, Sequence, Optional, final
 import torch
-import math
-from functools import cached_property
-from .utils import Permutation, permute_layout
-from ..exports.signature import OpSignature, ModeBase
+from ..exports.signature import OpSignature
 from ..exports.parser import OpCLIParser
 
 


iree/turbine/kernel/boo/fusion/replacement.py
```diff
@@ -235,3 +235,27 @@ def replace_aten_convolution_backward(node: Node):
...
+
+def replace_aten_batch_norm(node: Node) -> None:
+    """
+    Force batch norm to be computed in CNHW layout, with separate transpose
+    dispatches. This is currently the preferred approach for IREE as batch norm
+    contains reductions that are parallel over the channels dimension, and IREE
+    codegen currently performs best on outer-parallel reductions.
+    """
+
+    def replacement_fn(input_nchw: torch.Tensor, *args):
```
zjgarvey: Does this get expanded when running decompositions? That would be kind of convenient.

rkayaith: yep the call gets inlined and the batch norm op gets expanded.

PR #1253, opened 2025-12-17
[BOO] Add driver csv tests

zjgarvey:
This adds tests for the `--csv` flag in the `iree-boo-driver`. The two main benefits of these tests:

1. Ensure the csv output of the driver is what we expect.
2. Ensure the `num_dispatches` for forward conv doesn't regress. E.g., introducing memcopies as referenced in https://github.com/iree-org/iree-turbine/pull/1252

### DIFF
--- a/tests/kernel/boo/driver/csv_output_test.py
+++ b/tests/kernel/boo/driver/csv_output_test.py
@@ -1,0 +1,88 @@
+from pathlib import Path
+import shlex
+import pytest
+import torch
+import tempfile
+import csv
+
+from iree.turbine.kernel.boo.driver import driver
+
+
+def _read_csv_as_dicts(
+    path: Path, *, backends: list[str] = [driver.DEFAULT_BACKEND]
+) -> list[dict[str, str]]:
+    """Reads the csv file at `path`, verifies headers, and returns the data as a list of dicts."""
+    assert path.is_file()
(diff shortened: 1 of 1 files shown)


tests/kernel/boo/driver/csv_output_test.py
```diff
@@ -0,0 +1,88 @@
...
+        # Check we don't encounter an error.
+        assert driver.main(args) == 0
+        data = _read_csv_as_dicts(csv_file, backends=[backend])
+        # Check we have exactly one command being run in the driver.
+        assert len(data) == 1
+        row = data[0]
+        # Check the arguments column contains the individual command.
+        assert row["arguments"] == shlex.join(command_args)
+        # Check the number of dispatches is equal to the number of iterations.
+        assert row[f"{backend} num_dispatches"] == str(iters)
```
rkayaith: not for this PR, but maybe we should normalize `num_dispatches` so it's per-iteration instead of total. right now it's a bit annoying to interpret since you need the iteration count from the command.

rkayaith (approved):
nice idea, thanks!

tests/kernel/boo/driver/csv_output_test.py
zjgarvey: Yeah, thats a good idea.

PR #1259, opened 2026-01-08
Added support for post-replacement replace_getitem_users calls

keshavvinayak01:
After adding support for POST_DECOMPOSITION replacements (https://github.com/iree-org/iree-turbine/pull/1224): 
- Decomposition captures the dispatcher's actual selections
- Graph manipulation preserves our desired ops without re-execution

Changes in this patch:
- Changing metadata derivation for decomposed `getitem` nodes to support replacements after decomposition. 
- Skip permutations when the parent reported None. This keeps multi-output replacements stable after we moved decomposition ahead of replacements.

If a replacement creates ops that need decomposition, doing replacement first would let decomposition handle them. But:
- Current replacements don't create such ops
- If needed, you can add those ops to the decomposition table

For any future replacement policies written, this patch would come in handy.

### DIFF
--- a/iree/turbine/kernel/boo/fusion/replacement.py
+++ b/iree/turbine/kernel/boo/fusion/replacement.py
@@ -5,6 +5,7 @@ from operator import getitem
 import torch
 from torch import fx
 from torch.fx.node import Target, Node
+from torch.fx.passes.shape_prop import TensorMetadata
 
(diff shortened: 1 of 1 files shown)


rkayaith:
imo we should hold off on this until we have a concrete use case, but I'll defer to @zjgarvey if he thinks it's useful to merge now.

Please add a test if we decide to go forward with this though.

PR #1264, opened 2026-01-13
[CI] Remove torch from `requirements.txt` and add nightly pytorch ci

zjgarvey:
This PR removes pytorch from the base `requirements.txt` file, and adds a nightly cron job to test main turbine with nightly pytorch on CPU.

The reason we are removing `torch`, `torchvision`, and `torchaudio` from the `requirements.txt` file is primarily because it is never used. It is not used by CI scripts, it is not used in the build and release process, and it is not used when following any user-facing documentation for setup. Furthermore, having the base requirements include torch (with a pinned upper limit) has been inconvenient for external nightly testing (e.g., https://github.com/nod-ai/amd-shark-ai/pull/2758), and sometimes for our own jobs (e.g. https://github.com/iree-org/iree-turbine/pull/1218).

Although this PR does not include fixes for nightly rocm + pytorch from TheRock due to an existing issue (https://github.com/iree-org/iree-turbine/issues/1265), we should fix these issues and add a `pytorch-rocm-nightly-requirements.txt` to include in the nightly test matrix.

### DIFF
--- a/.github/workflows/test_pytorch_nightly.yaml
+++ b/.github/workflows/test_pytorch_nightly.yaml
@@ -47,14 +47,9 @@ jobs:
       - name: Install pip deps
         run: |
           python -m pip install --no-compile --upgrade pip
-          # Note: We install in three steps in order to satisfy requirements
-          # from non default locations first. Additionally, installing
-          # cpu/rocm-specific pytorch wheels first avoids installing
-          # unnecessary libraries included in default pytorch.
           pip install --no-compile -r ${{ matrix.pytorch_requirements }}
           pip install --no-cache-dir -r requirements-iree-pinned.txt
-          pip install -r requirements.txt
-          pip install -e .
+          pip install -r requirements.txt -e .
 
       - name: Run unit tests
         if: ${{ !cancelled() }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,10 +61,6 @@ jobs:
       - name: Install pip deps
         run: |
           python -m pip install --no-compile --upgrade pip
-          # Note: We install in three steps in order to satisfy requirements
-          # from non default locations first. Additionally, installing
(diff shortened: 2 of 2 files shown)


.github/workflows/test_pytorch_nightly.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Run unit tests
+        if: ${{ !cancelled() }}
+        run: |
+          echo "Running unit tests for torch nightly."
+          echo $(pip list | grep torch)
+          pytest -n 4 --timeout=300 --capture=tee-sys -vv .
+
+
+  # Depends on all other jobs to provide an aggregate job status.
+  ci_summary:
```
ScottTodd: Use a unique step name here. Don't share the `ci_summary` name with https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/ci.yaml

```suggestion
  # Depends on all other jobs to provide an aggregate job status.
  test_pytorch_nightly_summary:
```

We've been using these summary steps as "required status checks" with github branch protection and if there are multiple steps with the same name (even in different workflow files) the UI does not disambiguate them clearly: 

Image

.github/workflows/test_pytorch_nightly.yaml
```diff
@@ -0,0 +1,84 @@
+# Copyright 2025 The IREE Authors
```
ScottTodd: nit: you can start using 2026 in the copyright year for files authored this year

.github/workflows/test_pytorch_nightly.yaml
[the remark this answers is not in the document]
zjgarvey: I ran this job here: https://github.com/zjgarvey/iree-turbine/actions/runs/21006414636/job/60389806394

It had some failures, but I'd like to get this logged on turbine main and start filing some issues.

ScottTodd (approved):

PR #1287, opened 2026-01-29
[Turbine] Add single-writer write analysis and guarded store codegen

miguelcsx:
- Implement  to determine single-writer guarantees (unique, owner predicate, or needs guard).
- Add guarded store emission supporting owner predicate and atomic first-writer guard.
- Add unit tests for analysis and guarded-store dispatch.
- Expose  option in codegen base options.
- Address missing work on #713


zjgarvey (changes requested):
## Review findings

- **[P1] Guarded code generation is unreachable.** In `iree/turbine/kernel/compiler/vector_codegen.py:426`, every affected write calls `analyze_write(None, ref_shape)`. The analyzer immediately classifies `mapping is None` as `PROVEN_UNIQUE`, so production writes always take the unconditional-store path and neither guard can be emitted. Please derive and pass the actual write mapping/index relationship.

- **[P1] The atomic fallback cannot reliably select a writer.** In `iree/turbine/kernel/compiler/write_codegen.py:77-86`, `memref.alloca` produces an uninitialized flag, but no writer initializes it to zero or synchronizes before the atomic increment. Address space 3 is also workgroup-local, so it cannot arbitrate colliding workgroups, and a single scalar flag cannot choose one writer for each distinct output equivalence class.

- **[P1] The owner predicate is insufficient for multidimensional mappings.** In `iree/turbine/kernel/compiler/write_analysis.py:81-82`, every all-constant output mapping becomes only `grid[0] == 0`. For a mapping with multiple iterators, all values of the remaining axes still write the same destination. The predicate must constrain every collapsed axis or otherwise derive a unique owner from the mapping.

- **[P2] The identity proof ignores `ref_shape`.** In `iree/turbine/kernel/compiler/write_analysis.py:70-71`, `mapping.is_identity()` is accepted even when the mapping output shape does not cover the referenced memory shape. The existing `Write.has_identity_mapping()` also requires `mapping.output_shape == mem_shape`; this analysis needs an equivalent shape check before proving uniqueness.

The new tests do not exercise guarded IR emission: `NEEDS_GUARD` is invoked only while guards are disabled. The unit/type-check jobs pass, but pre-commit currently fails on formatting and trailing whitespace.

PR #1299, opened 2026-02-06
[BOO] Rework numerics testing with statistical validation

zjgarvey:
## Summary

Reworks `numerics.py` to provide both statistical validation and structured testing of kernel numerics, with a pass/fail API integrated into `driver.py` CSV and terminal output.

We check that our errors (BOO vs high-precision CPU reference) are centered around zero, and have a comparable stddev with PyTorch GPU vs. the same reference. The checks use an allclose-style tolerance: `|value| < atol + rtol * reference`, so small absolute errors always pass and the threshold scales with output magnitude or PyTorch noise level.

Additionally, we add structured numerics testing, which creates a small block of 1's inside splat-zero inputs. This catches egregious indexing errors that statistical tests might miss.

## Key changes

**`numerics.py` — full rewrite:**
- Three key comparisons: `boo_gpu_err`, `pytorch_gpu_err`, `boo_pytorch_diff` (all vs configurable-dtype CPU reference, default float64)
- Batched error collection: collects `min_samples` elements across multiple seeded batches before computing stats
- Statistical pass criteria: mean near zero (allclose-style with `mean_check_atol` + `mean_check_rtol * ref_abs_max`), stddev within tolerance (`stddev_check_atol` + `stddev_check_rtol * pytorch_stddev`), no BOO NaN mismatch
- Structured input testing: splat-0 tensors with an adaptive block of 1's to catch indexing bugs (block size scales with tensor size, random offset via seed)
- Clean verdict API: `NumericsVerdict` dataclass with `format_verdict_simple` (one-liner PASS/FAIL with brief reasons) and `format_verdict_verbose` (full stats)

**`driver.py` — numerics integration:**
- CSV always emits all 11 numerics columns when `--verify-numerics` is set
- Terminal always prints a verdict; `--numerics-verbose` controls whether it shows full stats or just PASS/FAIL
- Failures are caught gracefully without aborting the timing run
- New CLI args: `--verify-numerics`, `--numerics-verbose`, `--numerics-min-samples`, `--numerics-stddev-rtol`, `--numerics-stddev-atol`, `--numerics-mean-atol`, `--numerics-mean-rtol`, `--numerics-reference-dtype`, `--skip-structured-tests`

**Tests:**
- `numerics_smoke_test.py`: unit tests for statistics computation, criteria evaluation, structured pattern generation, output formatters, and GPU end-to-end verification

### DIFF
--- a/iree/turbine/kernel/boo/driver/driver.py
+++ b/iree/turbine/kernel/boo/driver/driver.py
@@ -354,7 +354,7 @@ def main(args: list[str] = sys.argv[1:]) -> int:
                 )
                 verdict = verdicts[0]
             except Exception as exc:
-                if meta_args.verbose:
+                if meta_args.numerics_verbose:
                     traceback.print_exception(exc)
                 csv_row += ["N.A."] * len(numerics_csv_cols)
                 test_error += 1


iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,597 @@
...
+    """
+    Compute element-wise absolute errors between two tensors.
+
+    Returns a 1D tensor of absolute errors.
+    """
+    # Cast both to float64 for accurate error computation
+    comparee_f64 = comparee.to(dtype=torch.float64, device="cpu")
+    reference_f64 = reference.to(dtype=torch.float64, device="cpu")
+    errors = (comparee_f64 - reference_f64).abs().flatten()
+    return errors
```
zjgarvey: Subtraction between two numbers isn't really that risky, and pytorch will automatically promote mixed floating point dtypes for subtraction. Let's not explicitly cast here. I'm also a bit unsure if we need to transfer to cpu for this.

Let's always transfer to the "comparee" device (which should always be GPU) if possible, let's do this with `non_blocking=True` so we can transfer to GPU asynchronously.

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,597 @@
...
+    if isinstance(pytorch_result, torch.Tensor):
+        pytorch_result = (pytorch_result,)
+    if isinstance(boo_result, torch.Tensor):
+        boo_result = (boo_result,)
 
-_T = TypeVar("_T")
+    # Compare main results
+    main_idx = sig.main_result_index
+    pytorch_main = pytorch_result[main_idx].to(device=cpu, dtype=torch.float64)
+    boo_main = boo_result[main_idx].to(device=cpu, dtype=torch.float64)
```
zjgarvey: Let's do this on the GPU.

tests/kernel/boo/driver/numerics_test.py
```diff
@@ -0,0 +1,468 @@
...
+        """Test that required fields must be provided."""
+        stats = ErrorStatistics(
+            mean=1.0, stddev=0.5, max_abs_err=2.0, num_samples=100
+        )
+
+        assert stats.mean == 1.0
+        assert stats.stddev == 0.5
+        assert stats.max_abs_err == 2.0
+        assert stats.num_samples == 100
+        assert stats.shapiro_pvalue is None
```
zjgarvey: This is pretty weak. I'd remove this test.

zjgarvey (commented):
Some initial comments.

tests/kernel/boo/driver/numerics_test.py
```diff
@@ -0,0 +1,434 @@
...
+        errors = torch.randn(10000)
+        stats = compute_error_statistics(errors)
+        assert stats.num_samples == 10000
+
+        try:
+            from scipy.stats import normaltest  # noqa: F401
+
+            assert stats.normality_pvalue is not None
+        except ImportError:
+            assert stats.normality_pvalue is None
```
zjgarvey: We can remove this now that we aren't restricting the upper limit for the sample size.

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,611 @@
...
+        # Get main result tensors
+        f64_ref_main = f64_ref[main_result_idx]
+        pytorch_gpu_main = pytorch_gpu_result[main_result_idx]
+        boo_gpu_main = boo_gpu_result[main_result_idx]
+
+        # Compute errors
+        boo_gpu_errors_list.append(_compute_element_errors(boo_gpu_main, f64_ref_main))
+        pytorch_gpu_errors_list.append(
+            _compute_element_errors(pytorch_gpu_main, f64_ref_main)
+        )
```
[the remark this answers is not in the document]
zjgarvey: I think without somehow inferring the reduction size in a hacky way, there isn't a great way to do this.

E.g., looking at input/output shapes to try and infer reduction sizes, or adding another function to the `OpSignature` class? Both seem bad.

stddev should scale with machine eps and sqrt of reduction size. I'm not sure of a good way to get a threshold without knowing the latter quantity.

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,611 @@
...
+    return ErrorStatistics(
+        mean=mean,
+        stddev=stddev,
+        max_abs_err=max_abs_err,
+        num_samples=num_samples,
+        normality_pvalue=normality_pvalue,
+    )
+
+
+def evaluate_statistical_criteria(
```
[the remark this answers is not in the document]
zjgarvey: I think I've resolved this. Let me know if the changes make sense to you.

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,641 @@
...
+        return (
+            False,
+            f"Structured test failed: max_diff={max_diff:.2e} > atol={atol:.2e}",
+        )
+
+    return True, None
+
+
+def verify_numerics(
+    commands: Sequence[str],
```
amd-eochoalo: I'm not yet too familiar with MIOpen commands (nor boo for that matter), but as I understand it, boo also allows for users to input torch modules as input programs. E.g., from the [README](https://github.com/iree-org/iree-turbine/tree/6bcf2db91a3a203bab11dac6ff361e46d8fbfca7/iree/turbine/kernel/boo#enabling-supported-fusions-in-a-model)

```python
import torch

model : torch.nn.Module = ...

# To just use boo fusions:
boo_model = torch.compile(model, dynamic=False, backend="iree_boo")
# To use boo fusions and inductor:
boo_inductor_model = torch.compile(model, dynamic=False, backend="iree_boo_inductor")
```

Do you know if:

* the conditions you've determined necessary for passing (e.g., normality) can be generalized for any torch modules?
* is there a way we can pass torch modules to this testing framework?
* or is this out of scope?

Thanks!

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,641 @@
...
+        num_samples=errors.numel(),
+    )
 
-_T = TypeVar("_T")
 
+def is_approximately_negligible(
+    value: float, reference: float, atol: float, rtol: float
+):
+    """Combined absolute + relative check for negligibility, similar in style to allclose."""
+    return abs(value) < atol + rtol * abs(reference)
```
bjacob: Formally positive conditions (i.e. simple calls to this function without a negation) should define closed sets, not open sets, so I would prefer this inequality to be a `<=` rather than a `<`.

Checking: https://numpy.org/doc/2.3/reference/generated/numpy.allclose.html  does use `<=` not `<`.  So you can also just view that as a matter of consistency with numpy if you don't feel swayed by my argument about aligning topology and logics.

Finally, you could resolve that by looking at the degenerate case where one or both tolerances is exactly 0.  Then you still want a numerical error of exactly 0 to pass.

iree/turbine/kernel/boo/driver/driver.py
```diff
@@ -245,6 +320,69 @@ def main(args: list[str] = sys.argv[1:]) -> int:
...
+            for stats in [
+                verdict.boo_gpu_err,
+                verdict.pytorch_gpu_err,
+                verdict.boo_pytorch_diff,
+            ]:
+                if stats is not None:
+                    csv_row += [
+                        f"{stats.mean:.2e}",
+                        f"{stats.stddev:.2e}",
+                        f"{stats.max_abs_err:.2e}",
```
bjacob: See my other comment about precision in error diagnostics.

iree/turbine/kernel/boo/driver/numerics.py
zjgarvey: Unfortunately, the normality tests ended up being too flaky with the high sample sizes, so I've opted to remove it in favor of simply looking at the mean and standard deviation of elementwise signed error.

The specific math I have been looking at have been for reduction ops like mm and conv. 

In general, I don't know if these kinds of basic stat tests can carry over to full QDQ models, for example.

zjgarvey: We could certainly apply the same checks to a general nn module.

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,641 @@
...
+    except Exception as e:
+        return ErrorSampleResult(error_message=f"BOO compilation failed: {e}")
 
-    return tuple(_compare(a, e)._asdict() for a, e in zip(comparee, reference))
+    main_result_idx = sig.main_result_index
+    num_batches: Optional[int] = None
+    batch_idx = 0
 
+    while num_batches is None or batch_idx < num_batches:
+        seed = batch_idx * 12345 + 42  # Deterministic but varied seeds
```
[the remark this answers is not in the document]
zjgarvey: Yeah, that makes sense. The seeds are being varied either way, so it would be wasteful to be arbitrarily scaling and translating them by constants.

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -4,239 +4,641 @@
...
+        `stddev_check_atol`: Low-absolute bar. If boo_stats.stddev < atol, stddev check passes.
+        `stddev_check_rtol`: Relative to `pytorch_stats.stddev`. These should be comparable ~ 1.0.
+
+    Returns:
+        (mean_near_zero, stddev_ok, failure_reasons)
+    """
+    failure_reasons: list[str] = []
+
+    mean_near_zero = is_approximately_negligible(
+        boo_stats.mean, ref_abs_max, atol=mean_check_atol, rtol=mean_check_rtol
```
bjacob: Using ref_abs_max here is another instance of the L-infinity vs L-2 pitfall. How about passing stdev instead here?   The condition would then read " is the mean negligible compared to the stdev" which is as fair a definition of "being unbiased" as I can imagine.

bjacob: The adverse case is where the vector is filled with a constant nonzero value:
[1e-7, 1e-7, ..... , 1e-7]
Then mean = 1e-7 but stdev ~= 0, and maybe we want to tolerate that because mean is considered close enough to 0 in this instance (because maybe we know that the input values are close to 1.0 in magnitude and the factor 1e-7 is epsilon) but there's no universal reason why this is OK (as we discussed about scale-invariance and floats treating all scales on equal footing down to 1e-38), so this needs to be conveyed in some additional parameter, just passing that problem down to the user as the framework can't meaningfully decide on this. That is what `atol` is good for.

zjgarvey:
I'm going to land this as-is and iterate in smaller patches.

PR #1337, opened 2026-04-07
Bump PyTorch version pin to 2.11 and ROCm to 7.2

rkayaith:
Bump the PyTorch version upper bound from 2.10 to 2.11 and update the ROCm
index from 7.1 to 7.2.

Torch 2.11 enables `_AnalyzeCustomOpInputOutputMode` by default
([pytorch/pytorch@b777c6b](https://github.com/pytorch/pytorch/commit/b777c6bfd615eaa712515c899c1b5cc0092594d4)),
which excludes the `PreDispatch` dispatch key from TLS during custom op
dispatch. This breaks BOO ops that call `torch.export.export` during op
expansion; export traces via `PreDispatch`, so excluding it causes export
to constant-fold instead of tracing, crashing fx_importer.

The fix here is to temporarily un-exclude `PreDispatch` around the `torch.export.export`
calls in BOO op expansion.

Fixes #1338

### DIFF
--- a/iree/turbine/kernel/boo/ops/utils.py
+++ b/iree/turbine/kernel/boo/ops/utils.py
@@ -4,6 +4,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+import contextlib
 import os
 import math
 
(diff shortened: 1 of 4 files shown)


iree/turbine/kernel/boo/ops/utils.py
```diff
@@ -16,6 +17,20 @@
 from ....support.ir_imports import Operation, PassManager, Context
 from ....transforms.general.custom_op_expansion import ExpandCustomOpsPass
 
+
+@contextlib.contextmanager
+def enable_predispatch_key():
```
zjgarvey, line 22: Does this work with older versions of pytorch? Does it make sense to case based on the torch version if not?

rkayaith: I tested 2.5 and 2.10 and it works on both. I've added a comment saying this was introduced specifically for 2.11 though, in case this causes issues down the road.

zjgarvey (approved):

PR #1342, opened 2026-04-13
Bump IREE requirement pins and update BOO compile flag

yzhang93:
Update iree preprocessing flag according to https://github.com/iree-org/iree/pull/24055.


zjgarvey (approved):
LGTM

PR #1348, opened 2026-04-27
[BOO] add arg_compare (argmax/argmin) op

bangtianliu:
This PR adds `aten::argmax` and `aten::argmin` support to the BOO driver.

Then I can use BOO to evaluate the arg_compare implementation from the IREE end. 

Assisted-by: [Claude Code](https://claude.ai/code)


tests/kernel/boo/fusion/integration_test.py
```diff
@@ -372,3 +374,46 @@ def test_boo_sdpa_replacement(device: str, experimental: bool, dtype: torch.dtyp
                 assert "aten._scaled_dot_product_flash_attention" in call_strings
         else:
             assert "aten._scaled_dot_product_flash_attention_for_cpu" in call_strings
+
+
+@pytest.mark.xfail(
```
rkayaith: can you file an issue and reference it here

amd-eochoalo (approved):
This looks ok to me, but I am not too familiar with iree-turbine.

iree/turbine/kernel/boo/driver/numerics.py
```diff
@@ -139,6 +139,8 @@ def _compute_element_errors(
     # Exclude positions where both are NaN
     valid = ~(comparee_nan & ref_nan)
     errors = (comparee_on_ref[valid] - reference[valid]).flatten()
+    if not errors.is_floating_point():
+        errors = errors.to(torch.float64)
```
rkayaith: what's the issue with leaving the error as integers?

iree/turbine/kernel/boo/fusion/schema.py
```diff
@@ -120,6 +120,14 @@ def _layernorm_permute_filter(node: Node) -> bool:
     torch.ops.aten._scaled_dot_product_flash_attention.default: OpFusionSpec(),
     torch.ops.aten._scaled_dot_product_flash_attention_for_cpu.default: OpFusionSpec(),
     torch.ops.aten._scaled_dot_product_efficient_attention.default: OpFusionSpec(),
+    torch.ops.aten.argmax.default: OpFusionSpec(
+        recursive=False,
+        make_single_dispatch=True,
```
rkayaith: you should use the defaults here unless you specifically need these settings (and if so, I'd be curious why)

rkayaith (changes requested):
minor comments

tests/kernel/boo/fusion/integration_test.py
bangtianliu: Oh, forgot to change it. It should work on the CPU too, but let me double-check it.

iree/turbine/kernel/boo/driver/numerics.py
bangtianliu: Ops like argmax/argmin whose outputs are int64.

errors.mean() and errors.std() raise RuntimeError: mean(): could not infer output dtype.

rkayaith: minor nit but could you do `errors.mean(dtype=torch.float)` and `errors.float().std()` instead, so the conversion is done where its actually needed

rkayaith (approved):
LGTM, just one nit comment

PR #1350, opened 2026-04-29
Add RDNA3 devices to the torch mapping table

RattataKing:


MaheshRavishankar (approved):

iree/turbine/runtime/device.py
```diff
@@ -68,6 +68,17 @@
...
     "AMD Radeon AI PRO R9700": "r9700",
+    "AMD Radeon RX 7900 XTX": "rx7900xtx",
+    "AMD Radeon RX 7900 XT": "rx7900xt",
+    "AMD Radeon RX 7800 XT": "rx7800xt",
+    "AMD Radeon RX 7700 XT": "rx7700xt",
+    "AMD Radeon PRO V710": "v710",
+    "AMD Radeon PRO W7900": "w7900",
+    "AMD Radeon PRO W7800": "w7800",
+    "AMD Radeon PRO W7700": "w7700",
+    "AMD Radeon Graphics": "phoenix",
```
rkayaith: is this one actually correct, "AMD Radeon Graphics" sounds too generic

rkayaith: looks like `AMD Radeon Graphics` is the fallback for when the rocm install is missing the mapping for the device, we definitely *don't* want to have an entry for that

rkayaith (changes requested):

rkayaith (approved):

PR #1353, opened 2026-05-04
[BOO] Auto pass split-reduction-low-parallelism for RDNA targets

yzhang93:
The IREE split-reduction heuristic gates a low-parallelism parameter set (tuned for RDNA-class GPUs) behind the boolean flag `--iree-dispatch-creation-split-reduction-low-parallelism`.

Auto-enable it when the HIP target's gfx/SKU prefix is RDNA-class (gfx10/11/12 or rx/r9/w7/w6/w5). CDNA targets keep the default.

Validation on RX 9070 XT: +1.3% to +4.3% geomean speedup with this flag.

### DIFF
--- a/iree/turbine/kernel/boo/runtime/launch.py
+++ b/iree/turbine/kernel/boo/runtime/launch.py
@@ -239,10 +239,26 @@ def default_compiler_flags_callback(device: Device, cache_dir: Path) -> list[str
         flags.append(
             "--iree-dispatch-creation-enable-fuse-padding-into-linalg-consumer-ops"
         )
-        flags.append("--iree-dispatch-creation-enable-split-reduction")
         flags.append(
             "--iree-preprocessing-pass-pipeline=builtin.module(iree-preprocessing-backward-data-conv-pipeline)"
         )
+        flags.append("--iree-dispatch-creation-enable-split-reduction")
+        # Auto-enable the low-parallelism split-reduction parameter set for
+        # RDNA-class targets (RX 9070 XT etc., or gfx10/11/12). CDNA-class
+        # targets use the default.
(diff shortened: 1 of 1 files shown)


iree/turbine/kernel/boo/runtime/launch.py
```diff
@@ -239,10 +239,26 @@ def default_compiler_flags_callback(device: Device, cache_dir: Path) -> list[str
         flags.append(
             "--iree-dispatch-creation-enable-fuse-padding-into-linalg-consumer-ops"
         )
-        flags.append("--iree-dispatch-creation-enable-split-reduction")
         flags.append(
             "--iree-preprocessing-pass-pipeline=builtin.module(iree-preprocessing-backward-data-conv-pipeline)"
         )
+        flags.append("--iree-dispatch-creation-enable-split-reduction")
+        # Auto-enable the low-parallelism split-reduction parameter set for
```
rkayaith, line 246: This is ok for now, but are there plans to move this logic into the compiler?

yzhang93: Yes, this flag can be removed after the split reduction is refactored as backend dependent. The details were shared by Hanhan previously,

rkayaith: for my own curiosity, do you have a link to this?

rkayaith (approved):

iree/turbine/kernel/boo/runtime/launch.py
yzhang93: @rkayaith https://github.com/iree-org/iree/discussions/24022

PR #1354, opened 2026-05-05
Preserve eager tensor alias lifetimes

### LINKED ISSUE #24324: local-task: queue_alloca FAILED_PRECONDITION on chained eager invokes
A short chain of eager iree-turbine compute ops on `local-task` fails with `FAILED_PRECONDITION` from `iree_hal_local_transient_buffer_*` during a `hal.device.queue.alloca` call. This is a new failure starting in the `3.12.0rc20260429` nightly.

This is minimized from an `iree-turbine` test: [`tests/dynamo/tensor_test.py::TensorTest::test_nn_linear`](https://github.com/iree-org/iree-turbine/blob/e38379a0b06ca910673ed4d3c3b29a1f3a398929/tests/dynamo/tensor_test.py#L96-L105)

### Repro

```bash
python -m venv .venv
source .venv/bin/activate
pip install --pre --find-links https://iree.dev/pip-release-links.html \
  iree-base-compiler==3.12.0rc20260429 \
  iree-base-runtime==3.12.0rc20260429 \
  iree-turbine==3.10.0rc20260430
pip install 'torch<=2.10.0,>=2.5' --index-url https://download.pytorch.org/whl/cpu
python repro.py
```

`repro.py`:

```python
import torch
from iree.turbine.dynamo import TurbineMode
from iree.turbine.runtime import Device

with TurbineMode():
    Device("local-task").set()
    try:
        a = torch.randn(2, 3).to("turbine")
        a = a.detach()
        a = a + 1
        out = a + 1
        print("PASS:", out)
    finally:
        Device.current().clear()
```

### Error output

```
Traceback (most recent call last):
  File "/.../repro.py", line 15, in <module>
    out = a + 1
          ~~^~~
  File "/.../iree/turbine/dynamo/tensor.py", line 87, in __torch_function__
    return super_fn(*args, **kwargs or {})
  File "/.../iree/turbine/dynamo/tensor.py", line 79, in super_fn
    return func(*args, **kwargs)
  File "/.../iree/turbine/dynamo/tensor.py", line 230, in __torch_dispatch__
    return compute_method(func, *args, **kwargs)
  File "/.../iree/turbine/dynamo/tensor.py", line 496, in compute_method
    ret_val = DeviceTensor.from_torch(torch.tensor(py_arg))
  File "/.../iree/turbine/dynamo/tensor.py", line 258, in from_torch
    dev_tensor = DeviceTensor._async_create_empty(
  File "/.../iree/turbine/dynamo/tensor.py", line 307, in _async_create_empty
    buffer = hal_device.queue_alloca(alloc_size, wait_semaphores, signal_semaphores)
RuntimeError: allocating memory on queue: c/runtime/src/iree/hal/local/transient_buffer.c:249: FAILED_PRECONDITION; transient buffer has not been committed; ensure the alloca signal semaphores are satisfied before accessing it
```

### Environment

- `iree-base-compiler==3.12.0rc20260429`
- `iree-base-runtime==3.12.0rc20260429`
- `iree-turbine==3.10.0rc20260430`
- `torch==2.10.0+cpu`

### Bisect

- Last passing nightly: `iree-3.12.0rc20260428`
- First failing nightly: `iree-3.12.0rc20260429`
- Diff: https://github.com/iree-org/iree/compare/iree-3.12.0rc20260428...iree-3.12.0rc20260429
- Suspect commit: add699f971

benvanik:
## Summary

- Keep eager `aten.detach` on an alias path that shares the original `DeviceTensor` storage.
- Reuse input storage for compiled eager results that return an input HAL buffer.
- Extend input storage lifetimes to the async invocation signal after a successful VM invocation, covering temporary scalar inputs and other short-lived inputs.
- Add a regression that drops the pre-detach tensor, forces Python GC, then runs two scalar adds through the detached alias before synchronizing the result.

## Diagnosis

`DeviceTensor` storage owns the HAL transient-buffer deallocation. The eager path was wrapping returned buffers in fresh `Storage` objects even when the returned HAL buffer aliased an input buffer. For `aten.detach`, this meant `a = a.detach()` could destroy the original tensor and queue a transient `queue_dealloca` while the detached alias still referenced the same buffer.

There was a second lifetime gap for temporary inputs created during eager execution, such as Python scalar constants converted to `DeviceTensor` inputs. Their storage readiness only tracked the producer fence. After the VM invocation was queued, those temporary inputs could be collected and queue a deallocation before the dispatch that reads them had completed. That can make a later queue operation observe the failed local-task timeline instead of the real operation order.

This PR gives eager aliases a single storage owner and advances every input storage used by a successfully queued invocation to the invocation signal fence. Deallocations queued by Python object destruction now wait for all already-scheduled consumers of that storage.

Fixes iree-org/iree#24324

## Testing

- `python tests/dynamo/tensor_test.py TensorTest.test_detach_alias_keeps_storage_alive`
- Reported issue reproducer from iree-org/iree#24324, run against this checkout.
- `python tests/dynamo/tensor_test.py`
- `git diff --check`

### DIFF
--- a/iree/turbine/dynamo/tensor.py
+++ b/iree/turbine/dynamo/tensor.py
@@ -309,6 +309,14 @@ class DeviceTensor(torch.Tensor):
         storage.ready_fence.insert(*alloca_complete_semaphore)
         return DeviceTensor(size, dtype, raw_data=storage)
 
+    def _alias(self, signal: Optional[HalFence] = None) -> "DeviceTensor":
+        """Creates a tensor alias sharing this tensor's storage owner."""
+        if signal is not None:
+            self._storage.ready_fence = signal
+        return DeviceTensor(
+            self.size(), self.dtype, raw_data=self._storage, requires_grad=False
+        )
+
     @staticmethod
     def _from_buffer(
         buffer: HalBuffer,
@@ -474,6 +482,33 @@ def _get_device_state() -> DeviceState:
     return DeviceState(driver="local-task")
 
 
+def _is_aten_op(src_op, op_name: str) -> bool:
+    return isinstance(src_op, torch._ops.OpOverload) and src_op.name() == op_name
+
+
+def _find_alias_storage(buffer: HalBuffer, input_tensors: Sequence[DeviceTensor]):
+    for input_tensor in input_tensors:
+        if (
+            isinstance(input_tensor, DeviceTensor)
+            and input_tensor._storage.buffer == buffer
+        ):
+            return input_tensor._storage
+    return None
+
+
+def _wrap_exec_result(
+    exec_res, input_tensors: Sequence[DeviceTensor], device: Device
+) -> DeviceTensor:
+    alias_storage = _find_alias_storage(exec_res.buffer, input_tensors)
+    if alias_storage is not None:
+        if exec_res.signal is not None:
+            alias_storage.ready_fence = exec_res.signal
+        return DeviceTensor(exec_res.size, exec_res.dtype, raw_data=alias_storage)
+    return DeviceTensor._from_buffer(
+        exec_res.buffer, exec_res.size, exec_res.dtype, device, exec_res.signal
+    )
+
+
 # Inspiration from https://github.com/nod-ai/SHARK-ModelDev/blob/8293de5414889c72ff5cd10bf33c43fb0a3ea3ee/python/iree/turbine/aot/builtins/jittable.py#L212-L237
 # and https://github.com/nod-ai/SHARK-ModelDev/blob/main/python/iree/turbine/dynamo/backends/cpu.py
 # TODO: Try to generalize for other devices.
@@ -483,6 +518,13 @@ def compute_method(super_fn, *args, **kwargs):
     # is often wrapped by DisableTorchFunction.
     init_py_args = args[:-1]
     src_op = args[-1]
+
+    if (
+        _is_aten_op(src_op, "aten::detach")
+        and len(init_py_args) == 1
+        and isinstance(init_py_args[0], DeviceTensor)
+    ):
+        return init_py_args[0]._alias()
 
     any_turbine_tensor = False
     devices_set = set()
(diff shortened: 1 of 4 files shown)


iree/turbine/dynamo/tensor.py
```diff
@@ -484,6 +519,13 @@ def compute_method(super_fn, *args, **kwargs):
     init_py_args = args[:-1]
     src_op = args[-1]
 
+    if (
+        _is_aten_op(src_op, "aten::detach")
+        and len(init_py_args) == 1
+        and isinstance(init_py_args[0], DeviceTensor)
+    ):
+        return init_py_args[0]._alias()
+
```
rkayaith: if I understood correctly, `_wrap_exec_result` *should* handle the detach case as well, but this is just a fast path optimization? if that's the case, can we just simplify and remove the `detach` special case, as far as I know there aren't any performance sensitive users for this, so I'd err on the side of keeping the code simple.

rkayaith (commented):
Thanks for the quick fix! Just one comment

iree/turbine/dynamo/tensor.py
benvanik: Yep, agreed. I pushed 66d1c91 to remove the `detach` fast path and the now-dead helpers; `aten.detach` now uses the same `_wrap_exec_result` alias-storage path as other compiled aliases. The tradeoff is that `detach` is no longer a zero-cost eager alias: first use for a shape/dtype can pay the normal Dynamo/export/IREE compile path, and cached uses still enqueue a no-op VM invocation/fence signal. That shows up as an extra tx timeline step in the issue reproducer. Given there are no perf-sensitive users here, keeping one aliasing policy seems better; if this ever matters, the right fix is probably a principled view/metadata-op fast path instead of a `detach`-only special case. Re-verified the focused regression, the issue reproducer, full `tests/dynamo/tensor_test.py`, and `git diff --check`.

rkayaith (approved):

